PR #3, opened 2014-07-08
Set modules as not installable and unported

dreispt:


pedrobaeza:
Hi, Daniel,

It appears that you have deleted modules instead moving them. Can you check it, please?

Regards.

bwrsandman:
Would seem he delete the modules in ca52f4d, I am not sure why. There is no description to the PR and the the commit message only says "Purge unported v6 modules"

pedrobaeza:
Well, then it could be logical if these modules haven't been ported since 6.0 to be removed on 8.0 branch. This is something we have to deal when a module is not maintained across two versions. Do we delete it? Keep it on **unported** folder for ever? What about if the module doesn't apply anymore but nobody says?

Regards.

guewen:
+1 to delete after 2 versions unmaintained. Still available in the history anyway.

dreispt:
I tried for the commit history to be self explaining. If you look individually at each one it should be clear what is being done.

Maybe the "Purge unported v6 modules" can cause confusion: it would be too messy to have v6 and v7 modules mixed in the `__unported__` directory.
Since they are available in the history and in the 7.0 branch, I figure the most sensible thing is to consider them abandoned and delete them from v8 branch.

bwrsandman:
> Since they are available in the history and in the 7.0 branch, I figure the most sensible thing is to consider them abandoned and delete them from v8 branch.

:+1: for this reasoning, we can just forward port it from the 7.0 or 6.1 branch if we ever need it
__unported__ should be for modules which are planned to be ported

I would say for v_n -> v_n+1
1. delete __unported__ contents (assuming this list was followed for v_n-1 -> v_n)
2. for each v_n module which is incompatible with v_n+1
1. move incompatible modules to __unported__
2. set installable to false
3. create a github issue to port it

pedrobaeza:
:+1:

PR #6, opened 2014-07-22
Port project baseuser modules

dreispt:


project_baseuser/security/project_security.xml
```diff
@@ -9,7 +9,7 @@
         <!-- Project Managers (modified): no longer see all projects: will have same visibility rules as Project Users -->
         <record model="ir.rule" id="project.project_project_manager_rule">
             <field name="name">Project: project manager: does not see all (modified)</field>
-            <field name="domain_force">[]</field>
+            <field name="domain_force">[(1, '=', 0)]</field>
```
pedrobaeza, line 12: Why this?

dreispt:
I have this change locally for some time; I believe that at some point, the empty `[]` brought errors or problems...

pedrobaeza:
But this way, the rule is never applied. Is that your intention?

dreispt:
Yes. Standard behaviour is Project Managers see all Tasks & Issues.
This "disables" that rule, and they get the same rule as other Project Users, inherited from that role.

pedrobaeza:
OK, got it

:+1:

PR #7, opened 2014-07-22
Port project categ to v8

dreispt:


project_categ/i18n/pt.po
```diff
@@ -0,0 +1,73 @@
...
+
+#. module: project_categ
+#: view:project.task:0
+msgid "onchange_project(project_id)"
+msgstr "onchange_project(project_id)"
+
+#. module: project_categ
+#: view:project.task:0
+msgid "{'required':[('task_categ_id','!=',False)]}"
+msgstr "{'required':[('task_categ_id','!=',False)]}"
```
bwrsandman, line 73: For these cases, isn't better to leave the translation blank?

project_categ/project_categ_model.py
```diff
@@ -53,11 +53,6 @@ def _name_get_fnc(self, cr, uid, ids, prop, unknow_none, context=None):
         'code': fields.char('Code', size=10),
     }
     _order = 'parent_id,name'
-    _constraints = [
-        (orm.Model._check_recursion,
-         'Error! Cannot create recursive cycle.',
-         ['parent_id'])
-    ]
```
bwrsandman, line 60: Was the removal of this all that was needed to port the module to 8.0?

bwrsandman:
In porting to 8.0. We should be using the new api.

project_categ/i18n/pt.po
dreispt, line 73: Not sure; probably this takes them out of unstranslated lists.

project_categ/project_categ_model.py
dreispt, line 60: Yes, no other changes needed.

dreispt:
I agree on using the new API, but the priority is to get modules installable on v8, so we can work on database migrations. So, I would prefer to work on that (and on test coverage) at a later stage.

pedrobaeza:
I also agree with Daniel. First: get v8 compatibility with minimal changes. Second: refactor with new API.

Regards.

eLBati:
On 07/22/2014 04:15 PM, Pedro M. Baeza wrote:

> I also agree with Daniel. First: get v8 compatibility with minimal 
> changes. Second: refactor with new API.

:+1:

PR #8, opened 2014-07-22
Port project task materials

dreispt:


project_task_materials/__init__.py
```diff
@@ -0,0 +1,2 @@
+# -*- coding: utf-8 -*-
+import project
```
pedrobaeza: Please put from . import project

dreispt:
OK, done.

pedrobaeza:
:+1:

PR #17, opened 2014-08-26
[FIX] POT file and access rules file

pedrobaeza:
[FIX] POT file
[FIX] Access rules file


dreispt:
Not familiar with the module, so this can be a dumb question: are you sure Project Users should be denied write access? They are expected to perform invoice related actions.

pedrobaeza:
Well, the module extends contracts view to allow to perform recurring invoices. It is supposed to be for fixed-price services, and it is setup one time at the contract creation, so it's logical to be done only by managers, not by everyone. What do you think?

dreispt:
The permission in on Invoice Lines. Out of context this suggests that Project Users won't be able to create recurring invoices. Can you do a test (even if manual) using the Demo user to confirm if it works correctly?

pedrobaeza:
Hi, @dreispt, the permission is in the model _account.analytic.invoice.line_, not _account.invoice.line_, which is a new model created in this module to add lines to the recurring invoice. I have already made manual tests.

Regards.

dreispt:
OK :+1:

gurneyalex:
:+1:

PR #20, opened 2014-09-03
[FIX] account_analytic_analysis_recurring: Currency for created invoices.

pedrobaeza:


account_analytic_analysis_recurring/account_analytic_analysis_recurring.py
```diff
@@ -141,8 +141,8 @@ def onchange_recurring_invoices(
         return value
 
     def _prepare_invoice(self, cr, uid, contract, context=None):
-        context = context or {}
-
+        if context is None:
+            context = {}
```
dreispt, line 145: Is there any benefit for this change. Personally I prefer the more compact formulation.

dreispt:
Code review LGTM :+1:

account_analytic_analysis_recurring/account_analytic_analysis_recurring.py
pedrobaeza, line 145: Yes, this is for avoiding a possible side effect, where context is assigned, but evaluated as False ({}, for example). Reassigning it, you kill the possibility to propagate upstream any changes made in the current method, because the variable has been reassigned.

dreispt, line 145: I see, I wasn't aware of that.
However, in this case the `context` is not modified, so that's not an issue.
In fact, it's not used at all, so the empty context check is unnecessary and could be removed.

pedrobaeza, line 145: Yeah, you're right. This is something I do systematic in each code that falls in my hands ;) I will remove it, although it's not the main purpose of this PR.

pedrobaeza, line 145: I see that I cannot remove it because is used on l.191

dreispt, line 145: Sorry - I was looking at the wrong method. All OK then.

mohamedhagag:
I just want to thank Mr. @pedrobaeza for his fast interaction and good commitment.

Now I installed the account_analytic_analysis_recurring to replace the old one which have the currency bug.

I went to the installed modules and selected this module and clicked on the upgrade button - and while trying to access the sales menu - I get this error message:

Client Traceback (most recent call last):
  File "/opt/oe7/openerp/addons/web/http.py", line 204, in dispatch
    response["result"] = method(self, **self.params)
  File "/opt/oe7/openerp/addons/web/controllers/main.py", line 1459, in load
    base_action = Actions.read([action_id], ['type'], req.context)
  File "/opt/oe7/openerp/addons/web/session.py", line 42, in proxy
    result = self.proxy.execute_kw(self.session._db, self.session._uid, self.session._password, self.model, method, args, kw)
  File "/opt/oe7/openerp/addons/web/session.py", line 30, in proxy_method
    result = self.session.send(self.service_name, method, *args)
  File "/opt/oe7/openerp/addons/web/session.py", line 103, in send
    raise xmlrpclib.Fault(openerp.tools.ustr(e), formatted_info)

Server Traceback (most recent call last):
  File "/opt/oe7/openerp/addons/web/session.py", line 89, in send
    return openerp.netsvc.dispatch_rpc(service_name, method, args)
  File "/opt/oe7/openerp/netsvc.py", line 296, in dispatch_rpc
    result = ExportService.getService(service_name).dispatch(method, params)
  File "/opt/oe7/openerp/service/web_services.py", line 622, in dispatch
    security.check(db,uid,passwd)
  File "/opt/oe7/openerp/service/security.py", line 40, in check
    pool = pooler.get_pool(db)
  File "/opt/oe7/openerp/pooler.py", line 49, in get_pool
    return get_db_and_pool(db_name, force_demo, status, update_module)[1]
  File "/opt/oe7/openerp/pooler.py", line 33, in get_db_and_pool
    registry = RegistryManager.get(db_name, force_demo, status, update_module)
  File "/opt/oe7/openerp/modules/registry.py", line 203, in get
    update_module)
  File "/opt/oe7/openerp/modules/registry.py", line 233, in new
    openerp.modules.load_modules(registry.db, force_demo, status, update_module)
  File "/opt/oe7/openerp/modules/loading.py", line 350, in load_modules
    force, status, report, loaded_modules, update_module)
  File "/opt/oe7/openerp/modules/loading.py", line 256, in load_marked_modules
    loaded, processed = load_module_graph(cr, graph, progressdict, report=report, skip_modules=loaded_modules, perform_checks=perform_checks)
  File "/opt/oe7/openerp/modules/loading.py", line 188, in load_module_graph
    load_data(module_name, idref, mode)
  File "/opt/oe7/openerp/modules/loading.py", line 76, in <lambda>
    load_data = lambda *args: _load_data(cr, *args, kind='data')
  File "/opt/oe7/openerp/modules/loading.py", line 118, in _load_data
    tools.convert_csv_import(cr, module_name, pathname, fp.read(), idref, mode, noupdate)
  File "/opt/oe7/openerp/tools/convert.py", line 935, in convert_csv_import
    raise Exception(_('Module loading %s failed: file %s could not be processed:\n %s') % (module, fname, warning_msg))
Exception: Module loading account_analytic_analysis_recurring failed: file account_analytic_analysis_recurring/security/ir.model.access.csv could not be processed:
 Line 1 : No matching record found for external id 'project.group_project_manager' in field 'Group'

mohamedhagag:
Is it required the security/ folder in the recurring module ? 

Also the security rule for project.group_project_manager makes this module depends on the project module - I think - which is not a requirement for most service businesses i think .

Regards,

mohamedhagag:
Nice work, commenting the security csv line in **openerp**.py solved the issue and now the recurring invoices are generated in the correct currency.

lots of thanks for Mr. @pedrobaeza you saved me :) :+1:  .

Regards,

dreispt:
The security issue seems to be another bug: this module does not depend on "Project", so security shouldn't be using Project groups.

mohamedhagag:
OK, I found that the right security for this model is like below:

"account_analytic_invoice_line_manager","Recurring user","model_account_analytic_invoice_line","base.group_sale_salesman",1,1,1,1

This is to allow sales and sales managers to create recurring invoices, Also this is the default security for this model in v8 .

it is Ok as this module already depends on sale module.

rvalyi:
:+1:  LGTM

PR #18, opened 2014-09-02
Fix view errors introduced by upstream changes

dreispt:

PR #21, opened 2014-09-05
8.0 project_gtd

sebalix:
The `project_gtd` module has been removed from Odoo 8.0, and this PR aims to integrate this module to OCA.
The last version of the source code of the module has been retrieved from the 8.0 branch, then fixed/cleaned/improved a bit.
Translations are not updated yet (I don't know if there is a tool to manage that since the Github migration).

It seems that the build fails against the `service_desk_issue` module.


dreispt:
That was fixed by #18 - I just merged it, you should rebase your PR.
Also, there are a couple of PEP8 issues to be fixed.

sebalix:
That's done. I was using Pylint which does not show the same errors that flake8. PR updated, all seems fine.

dreispt:
No point in reviewing the source in detail, since in comes from the official modules...
With TravisCI green, LGTM :+1:

pedrobaeza:
I agree with Daniel. :+1:

sebalix:
Great, thank you.

PR #22, opened 2014-09-15
[FIX] account_analytic_analysis_recurring: #START# and #END# set correctly

pedrobaeza:
[IMP] Some conventions
[IMP] Copy method


dreispt:
Since you're returning at this, maybe you could also fix the security ACL issue: permissions are given to Project groups that are not required by the module's dependencies.

dreispt:
Despite my comment above, I see no reason not to merge this :+1:

pedrobaeza:
I have fixed permissions also, @dreispt :)

dreispt:
Great Pedro! :+1:

PR #25, opened 2014-09-29
New module to restore State to Project Stages

dreispt:
The State can be useful for action rules or record rules that operate depending on some "canonical" states.
See the README.md.

### DIFF
--- a/project_stage_state/__init__.py
+++ b/project_stage_state/__init__.py
@@ -1,0 +1,2 @@
+# -*- coding: utf-8 -*-
+from . import project
--- a/project_stage_state/README.md
+++ b/project_stage_state/README.md
@@ -1,0 +1,8 @@
+This module restores the `state` fields to Project Stages, removed in Odoo 8.0.
+
+For some use cases it‘s necessary to be able to map the multiple Stages into 
+a few broad groups.
+
+For example, this can allow to define automated actions and business logic for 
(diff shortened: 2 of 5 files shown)


project_stage_state/__init__.py
```diff
@@ -0,0 +1,2 @@
+# -*- coding: utf-8 -*-
+import project
```
pedrobaeza: `from . import project`

pedrobaeza:
Anyway, I think it's better to link business logic to 1..n stages instead of making this double link. Let's see what people think.

dreispt:
Fixed.

dreispt:
I intend to use this for teh `project_baseuser` module.

It adds record rules so that Employee users can create Tasks, and can edit them while in Draft or Cancelled states. Once a Task is started, only Project users are allowed to write on them. More interestingly perhaps, `project_issue_baseuser` extends that to Project Issues.

In practice, and depending on Project configuration, several Stages can map to the "Draft" state.
Keeping the record rules attached to a generic State allows Project Managers to organize their Project Stages at will, without needing to touch record rules configurations.

pedrobaeza:
Another one that has been merged automatically... Well, your explanation convinced me, so for my part :+1: Let's see if anyone has any problem with this.

dreispt:
What do you mean with "merged automatically"?

dreispt:
OK I got it. I warned about it on the PRs description :smile:

PR #28, opened 2014-10-06
Service type unported

vrenaville:
This module is uninstallable so I move it to **unported** directory


dreispt:
As per [my comment](https://code.launchpad.net/~camptocamp/project-service/7.0-port-project_service_type/+merge/162135/comments/391784) on Launchpad, this module comes from v6, where Contract Management didn't exist in OpenERP.

Since v7 we have several contract types, that IMO make the "service type" deprecated. See: https://doc.odoo.com/book/4/4_12_Services/4_12_Services_contract/

Do you have a use case where the standard Contract management features are not enough?

vrenaville:
Yes, I see your comment, so perhaps it's better to delete it from 7.0 branch and 8.0 branch

dreispt:
Maybe I confused myself.
If the module is not installable yes it should be moved into `__unported__`.
But you should leave the `installable` flag as False.
Other than this, the PR is OK for me.

dreispt:
One more review needed here!

__unported__/project_service_type/__init__.py
```diff
@@ -26,4 +26,5 @@
 # Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #
 ##############################################################################
-import project
\ No newline at end of file
+
+import project
```
pedrobaeza, line 30: `from . import project`

__unported__/project_service_type/__openerp__.py
```diff
@@ -26,23 +26,20 @@
...
-        "project_view.xml"
-     ],
-     "active": False,
-     "installable": False
+''',
+    'author' : 'Camptocamp',
+    'website': 'http://camptocamp.com',
+    'depends' : ['project'],
+    'data' : ['project_view.xml'],
+    'demo_xml' : [],
```
pedrobaeza, line 40: This should be `demo`, but remove it anyway because it's empty.

__unported__/project_service_type/project_view.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+	<data>
+
+			<record model="ir.ui.view" id="view_service_type_form">
```
pedrobaeza, line 5: This is a strange and very broad indentation. Maybe you are using tabs? Please fix it.

dreispt:
@pedrobaeza this is moving a non installable module into the `__unported__` directory, no no code check required.
The `installable` flag should be set to False, but maybe it could be done before merge.

pedrobaeza:
Sorry, I saw it the other way.

:+1: then.

tafaRU:
I quote @dreispt :

> But you should leave the installable flag as False.
> Other than this, the PR is OK for me.

See https://github.com/vrenaville/project-service/pull/1

dreispt:
Made the fix locally in https://github.com/OCA/project-service/commit/7f5f2acd33a67a4c71829af428032a8162983951.

PR #31, opened 2014-10-20
Fix resource.calendar ACL for Projects

dreispt:
Adds missing ACLs that raises errors when calculating "working hours" statistics.

It's related to https://github.com/odoo/odoo/pull/3201: the fix there for Project Users needs to be replicated here for Employee users.


pedrobaeza:
Looks reasonable. Why don't you link to 'Employee' group?

dreispt:
That what's being done: Employee is the `base.group_user` group.

pedrobaeza:
Sorry, fast reading and misunderstanding...

:+1:

PR #35, opened 2014-10-24
Correct set accounting accounts for auto generated invoices on multicompany environment

fedexin40:
On multicompany environment the create recurring invoices function will create all the invoices with accounting accounts belong to company currently selected by admin user.

This merge request solve that situation.

### DIFF
--- a/account_analytic_analysis_recurring/account_analytic_analysis_recurring.py
+++ b/account_analytic_analysis_recurring/account_analytic_analysis_recurring.py
@@ -93,7 +93,7 @@ class account_analytic_invoice_line(osv.osv):
              'uom_id': uom_id or res.uom_id.id or False,
              'price_unit': res.list_price or 0.0})
         if res.description:
-            result['name'] += '\n'+res.description
+            result['name'] += '\n' + res.description
 
         res_final = {'value': result}
         if result['uom_id'] != res.uom_id.id:
@@ -164,7 +164,7 @@ class account_analytic_account(osv.osv):
             raise osv.except_osv(
                 _('Error!'),
                 _('Please define a sale journal for the company "%s".') %
-                (contract.company_id.name or '', ))
+                (contract.company_id.name or '',))
         partner_payment_term = contract.partner_id.property_payment_term.id
         inv_data = {
             'reference': contract.code or False,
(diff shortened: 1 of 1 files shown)


account_analytic_analysis_recurring/account_analytic_analysis_recurring.py
```diff
@@ -241,8 +241,30 @@ def recurring_create_invoice(self, cr, uid, automatic=False, context=None):
...
             context['old_date'] = old_date
             context['next_date'] = new_date
-            self._prepare_invoice(
-                cr, uid, contract, context=context)
+            # Force company for correct evaluate domain access rules
+            context['force_company'] = contract.company_id.id
+
+            # Create new cursor for handle multi company environments
+            from openerp import pooler
+            db, pool = pooler.get_db_and_pool(cr.dbname)
```
pedrobaeza: You don't need a new cursor for this purpose.

agb80:
Hello @pedrobaeza , I'm @fedexin40  co worker and in fact the one to propose the solution. Please advice us on the best way to work on this scenario:

The customer is an ISP provider with 3 companies for sell same services to different customers, we are using contracts and this module to auto generate invoices every month but the problem is as follow:
- Companies A, B & C generate contracts on the system

Case 1: Contract belong to  company A & admin user currently logged in on Company A
- Invoice Company: Company A (Correct)
- Invoice Account -> belongs to company A (correct)
- Product account -> belongs to company A (correct)

Case 2: Contract belong to  company B & admin user currently logged in on Company A
- Invoice Company: Company B (Correct)
- Invoice Account -> belongs to company A (wrong)
- Product account -> belongs to company A (wrong)

Case 3: Contract belong to  company C & admin user currently logged in on Company A
- Invoice Company: Company C (Correct)
- Invoice Account -> belongs to company A (wrong)
- Product account -> belongs to company A (wrong)

Make several test and observed that Invoice & Product account where selected per the company where admin user where currently logged in, that's why we tried the fix including only the  `context['force_company'] = contract.company_id.id`` before calling the _prepare_invoice function.

Then we face another problem, when calling the _prepare_invoice function first time, ORM optimizations where populating all the partner objects at same time resulting in a similar behavior than before:

Case 1: Contract belong to  company A & first invoice belongs to Company A
- Invoice Company: Company A (Correct)
- Invoice Account -> belongs to company A (correct)
- Product account -> belongs to company A (correct)

Case 2: Contract belong to  company B & first invoice belongs to Company A
- Invoice Company: Company B (Correct)
- Invoice Account -> belongs to company A (wrong)
- Product account -> belongs to company A (wrong)

Case 3: Contract belong to  company C & first invoice belongs to Company A
- Invoice Company: Company C (Correct)
- Invoice Account -> belongs to company A (wrong)
- Product account -> belongs to company A (wrong)

That's why we decide to create a new cursor that forces ORM populate the information for the current contract only, is working right in production at this moment but would like to learn if there is a better way to implement this.

pedrobaeza:
Hi, Agustín,

Thanks for the huge explanation, and I understand your motivations, but I don't know if you have tried before to make a browse for each contract after setting context with force_company key. As I see in your code, you are using same browse, and that's why you receive wrong data. I think this code can work:

```
for contract in self.browse(cr, uid, contract_ids, context=context):
    # Force company to evaluate correctly domain access rules
    context['force_company'] = contract.company_id.id
    # Re-read contract with correct company
    contract = self.browse(cr, uid, contract.id, context=context)
    ...
```

agb80:
Hello again Pedro, no we didn't have try that approach, let us check and in case needed correct back the merge request.

agb80:
Just notice that @fedexin40 forget notify that the commit with changes asked for @pedrobaeza are live now. Please review again, thanks a lot!

pedrobaeza:
Thanks for the changes.

:+1:

dreispt:
It would be nice to have a test, but :+1:

PR #36, opened 2014-10-29
Fix error on Project Manager rule

dreispt:
Condition (1,'=',0) errors. Workaround is to make the former "access to all" rule only
to the Technical Features group.

### DIFF
--- a/project_baseuser/security/project_security.xml
+++ b/project_baseuser/security/project_security.xml
@@ -9,16 +9,12 @@
         <!-- Project Managers (modified): no longer see all projects: will have same visibility rules as Project Users -->
         <record model="ir.rule" id="project.project_project_manager_rule">
             <field name="name">Project: project manager: does not see all (modified)</field>
-            <field name="domain_force">[(1, '=', 0)]</field>
+            <field name="domain_force">[('id', '=', 0)]</field>
             <!-- Original data:
             <field name="model_id" ref="model_project_project"/>
             <field name="domain_force">[(1, '=', 1)]</field>
-            <field name="groups" eval="[(4,ref('project.group_project_manager'))]"/>
+            <field name="groups" eval="[(6,0,[ref('base.group_no_one')])]"/>
             -->
-            <field name="perm_read" eval="True"/>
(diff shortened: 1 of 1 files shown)


dreispt:
Simple fix here...

pedrobaeza:
Well, with this patch, does not users with technical features see all projects?

dreispt:
Yes. Do you think it's inappropriate? Any suggestion?

dreispt:
The effect I want is the same as deleting the rule. But I dislike using  `<delete>`, so I'm trying to make it not affect "regular" users.

pedrobaeza:
There are a lot of regular users that have technical features activated, so I think it's not a good choice. But why don't make it non effective as previously? Does it have any side effect?

dreispt:
`(1,'=',0)` now errors. It looks like in v8 the left hand can't be a value. Except for `(1,'=',1)`, which is valid :astonished: 
Maybe use some sure to fail filter, such as `('id','=',0)`?

pedrobaeza:
Yeah, I think we can try that.

project_baseuser/security/project_security.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <openerp>
-    <data noupdate="1">
+    <data noupdate="0">
```
pedrobaeza: Should we change this for entire file?

dreispt:
It seems that the [upstream policy](https://github.com/odoo/odoo/blob/8.0/addons/project/security/project_security.xml#L3) now is to use `noupdate="0"`, so I'm adjusting to that.

pedrobaeza:
No, you can see in https://github.com/odoo/odoo/blob/8.0/addons/project/security/project_security.xml#L40 that this is not true for access rules.

dreispt:
I guess you're right. I'll change that back then.

yvaucher:
appart from noupdate things to be fixed :+1:

pedrobaeza:
:+1:

PR #38, opened 2014-11-04
Fix `frozen dict` error in project SLAs

dreispt:
Writing changes to documents could result in an error in methods
writing on the current context. In fact, in v8 context is a frozen dict
and is not writable.
As a workaround, the methods now use a `ctx` dict copy of `context`.


yvaucher:
:+1:

pedrobaeza:
:+1:

max3903:
:+1:

PR #39, opened 2014-11-04
Add code health indicator

dreispt:
Experimental support for landscape.io continuous code health monitoring.

### DIFF
--- a/.landscape.yaml
+++ b/.landscape.yaml
@@ -1,0 +1,2 @@
+ignore-paths:
+    - __unported__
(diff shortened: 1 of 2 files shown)


yvaucher:
Interesting

Some requirements might be missing:
https://landscape.io/github/OCA/project-service/1/messages/error

dreispt:
The why the "experimental" bit is there. :wink: 
In fact, we need the `prospector` checker to follow the same rules used in TravisCi for pep8 and pylint checks.
I have [asked about this](https://github.com/landscapeio/landscape-issues/issues/83) and it's possible by adding a `.prospector.yaml` file. So it's something to fine tune.

yvaucher:
Ok thanks for the details.

dreispt:
I would love to see this merged, so that we can follow how it behaves.

pedrobaeza:
I would like to know how health is measured, but of course we should have a test repo, and it can be this, that you're the main responsible, so I fast track the PR. We can revert it later if any trouble.

dreispt:
Code health improved +21% after this merge :grin: 
https://landscape.io/github/OCA/project-service/3

PR #41, opened 2014-11-07
Fix "TypeError: 'NoneType' object is not iterable" on SLA scheduler

dreispt:


project_sla/project_sla_control.py
```diff
@@ -92,7 +92,7 @@ def write(self, cr, uid, ids, vals, context=None):
         new_state = vals.get('sla_state')
         if new_state:
             # just update sla_state without recomputing the whole thing
-            ctx = dict(context) or {}
+            ctx = dict(context) if context else {}
```
pedrobaeza, line 95: I better prefer to put:

```
if context is None:
    context = {}
```

to avoid dereference any upstream variable that is empty ({})

dreispt, line 95: I want to dereference, because an upstream `{}` won't be writable.

pedrobaeza, line 95: Ah, OK

pedrobaeza:
:+1:

yvaucher:
:+1:

PR #42, opened 2014-11-11
Restore the State field to Project Task

dreispt:
For the cases where some 7.0 processes still rely on the `state` field.


yvaucher:
@dreispt as it was removed, should it be deprecated?

dreispt:
@yvaucher What do you mean?

yvaucher:
@dreispt Sorry I thought it wasn't on an official model.

yvaucher:
:+1:

pedrobaeza:
:+1:

dreispt:
Thanks for the reviews.
The build is failing because of "print" statements unrelated to this module.
That would be better fixed in a separate PR.

PR #49, opened 2014-12-16
Separated Lint tests

yvaucher:


.travis.yml
```diff
@@ -6,9 +6,10 @@ virtualenv:
   system_site_packages: true
 
 env:
-  - VERSION="8.0" ODOO_REPO="odoo/odoo"
-  - VERSION="8.0" ODOO_REPO="OCA/OCB"
-  - VERSION="8.0" UNIT_TEST="1"
+  - VERSION="8.0" LINT_CHECK="1"
+  - VERSION="8.0" ODOO_REPO="odoo/odoo" LINT_CHECK="0"
+  - VERSION="8.0" ODOO_REPO="OCA/OCB" LINT_CHECK="0"
+  - VERSION="8.0" UNIT_TEST="1" LINT_CHECK="0"
```
dreispt, line 12: Can't `LINT_CHECK="0"` be removed from these lines?
Simple is better.

yvaucher, line 12: For now we must specify LINT_CHECK I think it makes sense for retro compatibility.

yvaucher:
Few pylint errors

.travis.yml
pedrobaeza, line 12: Yeah, I put that it has to be explicitely marked to not make the pylint checks, to keep compatibility with branches not converted to this new style, as the current policy states.

pedrobaeza:
:+1:

.travis.yml
dreispt, line 12: Yes, I confirm that.
:+1: then

dreispt:
Thanks for the cleanup.
I'm merging.

PR #52, opened 2015-01-14
german translations for service management modules

tv-openbig:


project_categ/i18n/de.po
```diff
@@ -0,0 +1,72 @@
...
+
+#. module: project_categ
+#: view:project.task:0
+msgid "onchange_project(project_id)"
+msgstr ""
+
+#. module: project_categ
+#: view:project.task:0
+msgid "{'required':[('task_categ_id','!=',False)]}"
+msgstr ""
```
pedrobaeza, line 72: This is going to change the behaviour. Put the same string or remove it at all.

service_desk/i18n/de.po
```diff
@@ -0,0 +1,78 @@
...
+
+#. module: service_desk
+#: model:ir.model,name:service_desk.model_project_project
+msgid "Project"
+msgstr "Projekt"
+
+#. module: service_desk
+#: model:ir.model,name:service_desk.model_account_analytic_location
+msgid "account.analytic.location"
+msgstr ""
```
pedrobaeza, line 27: Translate it to the same string. Leaving empty doesn't mean the English translation is used.

service_desk_issue/i18n/de.po
```diff
@@ -0,0 +1,91 @@
...
+msgid "Project Issue"
+msgstr "Fall"
+
+#. module: service_desk_issue
+#: view:project.issue:0
+msgid ""
+"{'default_project_id': project_id, 'default_name': name, "
+"'default_analytic_account_id': analytic_account_id, 'default_location_id': "
+"location_id}"
+msgstr ""
```
pedrobaeza, line 45: The same.

PR #48, opened 2014-12-11
Project SLA: refactoring and improvement, to make it more generic

katyukha:
What is done:
1. Added 'start_field_id' field to 'project.sla' model same as
   'control_field_id' field. This will make simpler integration with
   objects that have no 'create_date' field, or such field have other
   name. Also this will allow to control not only time from
   confirmation, but any cobination of date fields of object (model)
2. Added sql constraints to check if 'limit_qty' and 'warn_qty' entered
   correctly.
3. Partly fixed bug when control field is of type 'date' not 'datetime'
   Partly, because of, may be, incorect calculation of time when not
   exact time specified. but now it will not throw error.
   May be start using UoMs here, to be able calculate time diferences
   in hours or days, what can be choosed by user?
4. Bugfix with user_id. Now there no errors will be thrown on attempt
   to create SLA for model, which has no 'user_id' field
5. Changed code, that checks for `doc.state != 'cancelled'` to
   `doc.state not in ('cancelled', 'cancel')` because of
   'cancelled' state is not standard in other apps, in most apps
   'cancel' state used instead
6. Views refactoring. Especialy tree view added for 'project.sla'
   model, which shows model name and control field names in tree.

PR #58, opened 2015-01-16
[7.0] Try to fix building errors

katyukha:
Some times on version 7.0 it resoves same errors (model is not
initialized correctly, or at right time)

Just a try...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,7 @@ virtualenv:
   system_site_packages: true
 
 env:
-  - VERSION="7.0" ODOO_REPO="odoo/odoo"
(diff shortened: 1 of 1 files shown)


katyukha:
This approach have no sense here(

pedrobaeza:
That's not needed, so there must be another problem. Do you have a traceback?

katyukha:
[Travis build](https://travis-ci.org/OCA/project-service/jobs/47304301)
Seems that odoo loads views in incorect order, model was not loaded.

katyukha:
seems to be builded correctly.

katyukha:
changes from here moved to #48, so there no need to keep separate pull request.

PR #62, opened 2015-02-06
Add license key in __openerp__.py

yvaucher:


dreispt:
:+1:

rvalyi:
:+1:

__unported__/project_service_type/__openerp__.py
```diff
@@ -35,6 +35,7 @@
 This will add a type of service on project allowing you to categorize them.
 """,
      "website": "http://camptocamp.com",
+     "license": "GPL-3 or any later version",
```
StefanRijnhart: Header of this file actually says GPL-2 or any later version.

dreispt: So we need to ask @jgrandguillaume if he agrees to relicense it as AGPL-3.

max3903:
:+1:

__unported__/project_service_type/__openerp__.py
yvaucher: I changed it to comply with header. This can be re-licensed when porting it.

PR #65, opened 2015-02-12
Make current user the Issues default Contact person

dreispt:
It is expected for users to open Issues, so the Contact information
should be the current user. The Responsible person is not an option,
since it will be changed to the assigned person in the support team.

This was already the behaviour in version 7.0, and was removed in
version 8.0.

### DIFF
--- a/service_desk_issue/project.py
+++ b/service_desk_issue/project.py
@@ -41,6 +41,12 @@ class ProjectIssue(orm.Model):
             'project_id', 'code', type='char', string="Project Code"),
         }
 
+    _defaults = {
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

dreispt:
Merge?

pedrobaeza:
We need one more approval. Anyone?

joaoalf:
:+1:  LGTM

joaoalf:
@dreispt There's a pep8 error but it's not in your code. It would be nice to fix it.

PR #67, opened 2015-02-18
The owner of the invoice generated was the administrator instead of the user

fedexin40:
The owner of the invoice generated was the administrator instead of the user
If we were working in a multi-company environment in the invoice were added the taxes of the all companies

### DIFF
--- a/account_analytic_analysis_recurring/account_analytic_analysis_recurring.py
+++ b/account_analytic_analysis_recurring/account_analytic_analysis_recurring.py
@@ -183,6 +183,7 @@ class AccountAnalyticAccount(orm.Model):
             'fiscal_position': fpos and fpos.id,
             'payment_term': partner_payment_term,
             'company_id': contract.company_id.id or False,
+            'user_id': contract.manager_id.id or False, 
         }
         invoice_id = inv_obj.create(cr, uid, inv_data, context=context)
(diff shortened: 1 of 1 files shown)


account_analytic_analysis_recurring/account_analytic_analysis_recurring.py
```diff
@@ -191,8 +192,12 @@ def _prepare_invoice(self, cr, uid, contract, context=None):
             if not account_id:
                 account_id = res.categ_id.property_account_income_categ.id
             account_id = fpos_obj.map_account(cr, uid, fpos, account_id)
-            taxes = res.taxes_id or False
-            tax_id = fpos_obj.map_tax(cr, uid, fpos, taxes)
+            taxes = res.taxes_id
```
pedrobaeza: This is more elegant this way:

```
taxes = [x for x in res.taxes if x.company_id.id == context['force_company']
```

pedrobaeza:
Hi, @fedexin40, v8 standard makes the same way as this, putting administrator, and I think this is good, because you can know which invoices have been autogenerated. I see little sense to put account manager as invoice user.

About the other point, I have put you a comment.

fedexin40:
Hi @pedrobaeza  how are you?, i changed the admin to a user because a customer needed to have a control about the generated invoices.. many thanks for yours comments

PR #69, opened 2015-03-27
Slovenian translations added

sysadminmatmoz:
Since I recently completely finished with the main odoo 8 branch translations, I'll try to do the same with the OCA repos.


pedrobaeza:
As this is only translations, I'll merge directly.

PR #71, opened 2015-04-10
Port SLA report from v7 to v8

dreispt:
The commits for v7 were reused and seem to work perfectly on v8.


dreispt:
To be accurate, these commits were cherry picked from v7 and already have been reviewed.
I feel that a new full review could be skipped in this case...

jgrandguillaume:
:+1:

PR #75, opened 2015-04-28
Fix state write error

dreispt:
See https://github.com/odoo/odoo/issues/4349 for the related bug report.
Also adds the State field to the "Extra information" tab.

### DIFF
--- a/project_stage_state/project.py
+++ b/project_stage_state/project.py
@@ -36,4 +36,5 @@ class ProjectTaskType(models.Model):
 
 class ProjectTask(models.Model):
     _inherit = 'project.task'
-    state = fields.Selection(related='stage_id.state', store=True)
+    state = fields.Selection(
+        related='stage_id.state', store=True, readonly=True)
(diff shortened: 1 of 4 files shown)


guewen:
:+1:

yvaucher:
:+1:

project_stage_state/project.py
```diff
@@ -36,4 +36,5 @@ class ProjectTaskType(models.Model):
 
 class ProjectTask(models.Model):
     _inherit = 'project.task'
-    state = fields.Selection(related='stage_id.state', store=True)
+    state = fields.Selection(
+        related='stage_id.state', store=True, readonly=True)
```
pedrobaeza, line 40: Make here a comment about the Odoo bug

dreispt, line 40: It's in the [commit](https://github.com/dreispt/project-service/commit/3bb72abb62fbe9757b528e5d556cb2f7d5dc2fcb) message

PR #76, opened 2015-04-28
project_sla: code refactoring

katyukha:
1. ProjectSLA.reapply_sla refactored to reapply SLA only to objects
   that belong to specified SLA, instead of reapplying SLA to all objects
   controlled by any SLA related to contracts related to specifed SLA.
   As example we could imagine situation where one analytic contract
   have 5 diferent SLA, each for specific model, and when we click on
   SLA reapply button, program will reapply sla on all 5 models, but
   we expect, that SLA will be applied only to one model.
2. Analytic contract method of reapplying SLA refactored to improve
   performance and to be able to recalculate only specified SLA's.
3. project.sla.control methods '_get_computed_slas' and
   'store_sla_control' refactored.
   - internal functions factored out to the begining of the module.
   - if document have no filled 'start_date' field, then just skip
     this SLA for tested document, no error.

### DIFF
--- a/project_sla/analytic_account.py
+++ b/project_sla/analytic_account.py
@@ -38,31 +38,45 @@ class AnalyticAccount(orm.Model):
         The ``recalc_closed`` flag allows to also recompute closed documents.
         """
         ctrl_obj = self.pool.get('project.sla.control')
+        sla_obj = self.pool.get('project.sla')
+
+        context = {} if context is None else context.copy()
 
         exclude_states = ['cancelled', 'cancel']
         if not recalc_closed:
             exclude_states += ['done']
 
-        for contract in self.browse(cr, uid, ids, context=context):
+        # If this is called from SLA model, than we could reduce number of
+        # documents that need recalculation
+        if context.get('sla_ids', False):
+            sla_ids = context['sla_ids']
+        else:
+            sla_ids = sla_obj.search(cr, uid, [('analytic_ids', 'in', ids)],
+                                     context=context)
+
+        sla_recs = sla_obj.browse(cr, uid, sla_ids, context=context)
+        for m_name in set([sla.control_model for sla in sla_recs]):
             # for each contract, and for each model under SLA control ...
-            for m_name in set([sla.control_model for sla in contract.sla_ids]):
-                model = self.pool.get(m_name)
-                doc_ids = []
-                if 'analytic_account_id' in model._columns:
-                    doc_ids += model.search(
-                        cr, uid,
-                        [('analytic_account_id', '=', contract.id),
-                         ('state', 'not in', exclude_states)],
-                        context=context)
-                if 'project_id' in model._columns:
-                    doc_ids += model.search(
-                        cr, uid,
-                        [('project_id.analytic_account_id', '=', contract.id),
-                         ('state', 'not in', exclude_states)],
-                        context=context)
-                if doc_ids:
-                    docs = model.browse(cr, uid, doc_ids, context=context)
-                    ctrl_obj.store_sla_control(cr, uid, docs, context=context)
+            model = self.pool.get(m_name)
+            domain = [('state', 'not in', exclude_states)]
+
+            # if model have both 'project_id' and 'analytic_account_id' fields,
+            # then we should use `OR` condition to search for both of these
+            # fields. Otherwise, search only by one of fields
+            if ('analytic_account_id' in model._columns and
+                    'project_id' in model._columns):
+                domain += ['|', ('analytic_account_id', 'in', ids),
+                                ('project_id.analytic_account_id', 'in', ids)]
+            elif 'analytic_account_id' in model._columns:
+                domain += [('analytic_account_id', 'in', ids)]
+            elif 'project_id' in model._columns:
+                domain += [('project_id.analytic_account_id', 'in', ids)]
+
+            doc_ids = model.search(cr, uid, domain, context=context)
+
+            if doc_ids:
+                docs = model.browse(cr, uid, doc_ids, context=context)
+                ctrl_obj.store_sla_control(cr, uid, docs, context=context)
         return True
 
     def reapply_sla(self, cr, uid, ids, context=None):
--- a/project_sla/project_sla_control.py
+++ b/project_sla/project_sla_control.py
@@ -67,16 +67,18 @@ class SLAControl(orm.Model):
     _description = 'SLA Control Registry'
 
     _columns = {
-        'doc_id': fields.integer('Document ID', readonly=True),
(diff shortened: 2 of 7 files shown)


dreispt:
Thanks @katyukha - from my first look at it it seems to be mostly performance optimizations.

But I have a functional issue to report: I noticed that the Time dimension (Year, Quarter, Month...) uses the SLA Date Start. If shouldn't: to have a reliable period SLA reporting, it should use the SLA "close" date instead (`sla_limit_date`).

If not, the statistics for a certain month will keep changing when issues are gradually getting closed as months go by.

katyukha:
I think it depends on what field is used to determined period of issue/task/other document. if we will use close date, then we may miss documents, that are not closed yet, so we will have incorect statistics.

Maybe it makes sens to add same date-related fields based on "close date"?

dreispt:
If you have to measure SLAs using Start date, you must make sure that non closed documents are also counted. In this case the response times should be short, so that at the reporting date the non-close documents are surely in a "Will Fail" state. 
With this, you will be sure that all opened incident will be included on the SLA report.

Using the close date instead is also a valid criteria, and can make sense for longer response times.

katyukha:
So, to be sure that non-closed documents are also counted, there could be added field 'unclosed_count' or 'unclosed_percent' in report.

But as a better way, there could be added ability to base report on start_date and close_date. As I see, It may be implemented by adding another group of fields like (close_year, close_quarter, close_week, close_date) and adding ability to group report lines on them (it seems to be possible to show only those fields, that used in group_by in a view, thus allowing to reduce visible amount of fields to readable quantity). Or, maybe, just add separate report.

dreispt:
@katyukha TravisCI is unhappy...

dreispt:
@katyukha I found the cause for the failing test while working in another feature.
Please help review #83 to have that fixed.
Afterwards, rebasing this branch will probably see a green light on TravisCI.

katyukha:
Hello! #83 is for branch _8.0_. But this pull request is for branch _7.0_, so it seems that #83 is not related to this one

project_sla/analytic_account.py
```diff
@@ -38,31 +38,36 @@ def _reapply_sla(self, cr, uid, ids, recalc_closed=False, context=None):
...
-                        context=context)
-                if doc_ids:
-                    docs = model.browse(cr, uid, doc_ids, context=context)
-                    ctrl_obj.store_sla_control(cr, uid, docs, context=context)
+            model = self.pool.get(m_name)
+            domain = [('state', 'not in', exclude_states)]
+            if 'analytic_account_id' in model._columns:
+                domain += [('analytic_account_id', 'in', ids)]
+            if 'project_id' in model._columns:
+                domain += [('project_id.analytic_account_id', 'in', ids)]
```
katyukha: @dreispt What do You think, is there should be 'or' or 'and' logic in domain, if there are both fields ('analytic_account_id' and 'project_id') present in model?

pedrobaeza:
Please rebase to start from an updated branch

rafaelbn:
Hi @katyukha could you rebase to start from an updated branch (https://github.com/OCA/project/pull/76#issuecomment-147546964). Thanks

katyukha:
Rebased

project_sla/analytic_account.py
dreispt: @katyukha Yes, it would be best for it to be an "OR".

project_sla/project_sla_control.py
```diff
@@ -56,6 +60,27 @@ def safe_getattr(obj, dotattr, default=False):
     return obj
 
 
+def datetime2str(dt_value, fmt):  # tolerant datetime to string
```
dreispt: Use a one line docstring instead of a comment. (OK, I've see I was the one who wrote that line 😄 )

project_sla/project_sla_control.py
```diff
@@ -320,9 +325,21 @@ class SLAControlled(orm.AbstractModel):
...
+            try:
+                cr.execute("SAVEPOINT %s_sla_save" % self._table)
+                cr.execute("""
+                    SELECT * FROM %s WHERE id IN %%s
+                    FOR UPDATE of %s NOWAIT
+                """ % (self._table, self._table), (tuple(ids),))
+            except:
+                cr.execute("ROLLBACK TO %s_sla_save" % self._table)
+                _logger.error("Cannot lock table %s", self._table)
+                raise
```
dreispt: Do we really need this?

dreispt:
I'm :+1: for merging this.
It is v7 and we shouldn't be as picky as we are for v8 or v9 . code standards were lower by then.

project_sla/project_sla_control.py
katyukha: What about a case, when there are 5000+ documents to be updated by scheduler, which takes at least 10-60 seconds and at same time, some user tryes to update document that is been updated by scheduler?

katyukha:
Strange ci bechavior... I have changed only function comment to docstring and condition of `AnalyticAccount._reapply_sla`, to search documents by `project_id` _OR_ `analytic_account_id`

dreispt:
It was already failing before, because of an unrelated module failing to install. As I said, v7 branches are not as clean as the later ones.

project_sla/project_sla_control.py
dreispt: No problem if it gets updated again. Each individual Document is independent, so no real need enforce transaction integrity.

katyukha: It was more that a year ago, when I was playing with this module, but as I remember, it had a very slow performance on mass updates, when write some field on all documents of model, trigger recalculation of all bound SLA lines, which takes too much time.

And there was situation when half of SLA lines was recomputed, but transaction breaks because one of documents that have not been processed yet, was updated or removed.

It was too long time ago, so I do not remember correct reasons for this code, but now it seems working, and at least this code is running and do not takes my attenting now.

To simplify this PR, i can remove this piece of code...

dreispt: I also don't use this version anymore. Yes, please do that.

katyukha: Removed)

katyukha:
Is anyone still interested in merging this pull request?
Or, may be, it is time to close it, due to no activity for a long time on this pull request?

dreispt:
I just reviewed the code and I'm OK with merging it.
[Module tests](https://travis-ci.org/OCA/project/jobs/158846919#L2048) are passing. 
But can you fix the Lint errors before we merge?

Are you planning to port this to v8?

katyukha:
Lint errors seems to be fixed for `project_sla` addon.

We have no plans to port this refactoring to v8 yet.

dreispt (approved):

PR #78, opened 2015-04-28
Add reassign feature to Project Issues

dreispt:


project_issue_reassign/README.rst
```diff
@@ -0,0 +1,38 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+The Responsible (``user_id``) field is made read only, and can instead be
```
pedrobaeza, line 4: Why not a double field with groups and one readonly and another not.

dreispt, line 4: Last time I looked forms didn't behave well when the same field was added twice.
But the wizard has additional advantages: you can make several changes on a single write operstion: now it's user_id and project_id, but I would like to extend that. For example, a message (think of GitHub's "Comment and Close").

pedrobaeza, line 4: OK then. I don't remember problem with a doubled field with groups, because it's filtered of _fields_view_get_.

dreispt, line 4: Ah! OK, that particular case could work.
Another relevant advantage for the wizard is that you can use it from the list view to mass update several Tasks. It's useful for a first-line "dispatcher" use case.

pedrobaeza:
:+1:

yvaucher:
:+1:

project_issue_reassign/README.rst
```diff
@@ -0,0 +1,38 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
```
yvaucher, line 3: Please add title

project_issue_reassign/README.rst
```diff
@@ -0,0 +1,38 @@
...
+Regular users (Employee group) will be able to see who is handling the request,
+but won't be able to change it.
+
+Project Users can click on the "=> Reassign" link, in front of the current
+responsible, to open a dialog where they can select the new responsible and/or
+new Project/Service Team it should be assigned to.
+
+Mass reassignments can also be made, through the context menu action on the
+Issues list view.
+
```
yvaucher, line 16: Please add bug tracking link

project_issue_reassign/__openerp__.py
```diff
@@ -0,0 +1,35 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Reassign Project Issues',
+    'summary': 'Reassign Issues to other Responsibles and Projects',
```
yvaucher, line 22: missing license key

rafaelbn:
Hi @dreispt do you want to continue with this PR? I could review it. Not sure as @yvaucher comments were not attended.

dreispt:
Yeah, it's hard to keep up with the changes in the README format...
Expect for the boilerplate I have two +1 reviews with no pending issues.

yvaucher:
For the README part I think it can be merged without it.

But would be better to put at least the missing license in the manifest file.

Thus I will merge it and add the license

dreispt:
@yvaucher I think can spend a few minutes doing the requested fixes...

yvaucher:
@dreispt already doing it ;)

dreispt:
OK, thanks then!

dreispt:
For the record, I think this feature could go even further:
We could also add a message text to post in chatter and offer the choice to also change Stage, inlcuding Close/Cancel it.

PR #81, opened 2015-04-28
Restore v7 Task state buttons, and their underlying API

dreispt:


dreispt:
I will implement this feature in a different way, through an enhancement to `project_stage_state`.

PR #82, opened 2015-04-30
Move "Contracts on Project Docs" features into properly named modules

dreispt:
The "Service Desk" modules are actually a use case for the feature
adding related Contract information to Project Tasks and Issues.

Moving those features to properly named modules will make it easier
for their adoption and reuse.

For backward compatibility the Service Desk modules are kept, although
with little or no additional features at the moment.


dreispt:
Depends on #83.

dreispt:
Thinking again, it is best to do this on the next version migration.

PR #83, opened 2015-05-05
Fix bug on the Reapply of Contract SLAs

dreispt:


dreispt:
@katyukha Can you help reviewing this?

project_sla/analytic_account.py
```diff
@@ -44,15 +44,16 @@ def _reapply_sla(self, cr, uid, ids, recalc_closed=False, context=None):
             ctrl_models = set([sla.control_model for sla in contract.sla_ids])
             for model_name in ctrl_models:
                 model = self.pool[model_name]
-                domain = [] if recalc_closed else [('stage_id.fold', '=', 1)]
+                base = [] if recalc_closed else [('stage_id.fold', '=', 0)]
+                doc_ids = []
                 if 'analytic_account_id' in model._columns:
```
katyukha, line 49: Before here may be added something like:

``` python
if 'analytic_account_id' in model._columns and 'project_id' in model._columns:
    domain.append('|')
```

This will add 'or' logic if there are both columns in model.

katyukha:
As i see on version 8.0 project states replaced with project stages (separate model). And 'closed condition' is now based on stages. For project related docs (issues and tasks) it is possibly normal.

But, as of applying SLA for other documents (not only project related, for example sale orders or stock picking) it may bring errors or incorect behavior.

Also, what means fold on stages? I think it is not always means closed or opend task/issue.

May be it would be better to allow user to define closed condition on each SLA or SLA Rule configuration? This will allow to controll time for diferent stages/states (for example how mach time issue or task in draft state spent, or how much time issue or task was unassigned).

aitorbouzas:
Just tried it now, it works well.

dreispt:
I'm merging.

Issue #33, opened 2014-10-22
[7.0] analytic_hours_block : missing dependency


laurentmeuwly:
The addon analytic_hours_block is well installing. But you are not able to use it with standard account_invoice.
In fact, the addon is refering an invoice.department_id, wich does not exist:

```
'department_id': fields.related(
            'invoice_id', 'department_id',
             type='many2one',
             relation='hr.department',
             string='Department',
            store={
                'account.hours.block': (lambda self, cr, uid, ids, c=None: ids,
                                        ['invoice_id'], 10),
                'account.invoice': (_get_invoice, ['department_id'], 10),
            },
             readonly=True),
```

Saving a newly created hours block results in this error:
KeyError: "Field 'department_id' does not exist in object 'browse_record(account.invoice, ...)'"

dreispt:
Yes, that dependency should be removed from here and added by a separate module in https://github.com/OCA/department/,
PRs welcome.

LoisRForgeFlow:
This is solved, let's close it!

PR #89, opened 2015-06-02
[FIX][analytic_hours_block][7.0] fix issue #33

Olivier-LAURENT:
As suggested in #33, the PR moves `department_id` field definition to another module in https://github.com/OCA/department repository  (see https://github.com/OCA/department/pull/46 that proposes this module).

### DIFF
--- a/analytic_hours_block/__openerp__.py
+++ b/analytic_hours_block/__openerp__.py
@@ -32,9 +32,9 @@ to follow for example the user support contracts.
 This means, you sell a product of type "hours block"
 then you input the spent hours on the hours block and
 you can track and follow how much has been used.
-
- """,
+""",
     "author": "Camptocamp,Odoo Community Association (OCA)",
+    "maintainer": "Camptocamp - Acsone SA/NV",
     "license": 'AGPL-3',
     "website": "http://www.camptocamp.com",
     "depends": [
@@ -53,6 +53,9 @@ you can track and follow how much has been used.
         "report.xml",
         "security/hours_block_security.xml",
         "security/ir.model.access.csv",
+    ],
+    "demo": [
+        "hours_block_demo.xml",
     ],
     "active": False,
     "installable": True
--- a/analytic_hours_block/hours_block_menu.xml
+++ b/analytic_hours_block/hours_block_menu.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" ?>
+<?xml version="1.0" encoding="utf-8"?>
 <openerp>
     <data>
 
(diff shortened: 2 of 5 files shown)


Olivier-LAURENT:
In addition, the PR adds also some demo data and an `onchange` method on `invoice_id` field to avoid inconsistency between the invoice and the analytic account of an hours block.

dreispt:
I disagree. In fact, I believe that OCA/department should be dismantled.
I invite for a discussion on this in https://github.com/OCA/department/issues/40

Olivier-LAURENT:
Hi @dreispt,

Thanks for the quick review.

What do you disagree exactly ? Here to solve the issue #33 I simply proposed a fix following your own recommandations. 

Maybe the dismantling of https://github.com/OCA/department is a good idea but can it be the reason of the reject of a PR fixing a simple technical bug ?

dreispt:
@Olivier-LAURENT I just disagree with the department_id removal. But that can be only my opinion. I didn't actually review the fix itself...

analytic_hours_block/__openerp__.py
```diff
@@ -32,9 +32,9 @@
 This means, you sell a product of type "hours block"
 then you input the spent hours on the hours block and
 you can track and follow how much has been used.
-
- """,
+""",
     "author": "Camptocamp,Odoo Community Association (OCA)",
+    "maintainer": "Camptocamp - Acsone SA/NV",
```
dreispt: It's best to use a comma to separate the author list

analytic_hours_block/hours_block_menu.xml
```diff
@@ -6,12 +6,13 @@
...
         <record model="ir.actions.act_window" id="action_all_block_hour">
-			<field name="context">{'search_default_running': 1, 'search_default_group_department_id': 1}</field>
+            <field name="context">{'search_default_running': 1}</field>
             <field name="name">Hours Blocks</field>
             <field name="res_model">account.hours.block</field>
             <field name="view_type">form</field>
             <field name="view_mode">tree,form</field>
-            <field eval="False" name="view_id"/>
+            <field name="view_id" ref="invoice_tree_hour_block"/>
+            <field name="search_view_id" ref="view_account_invoice_filter"/>
```
dreispt, line 15: Are these two lines really needed. If not, removing them would make it easier to extend the module.

dreispt:
One small remark, otherwise LGTM.
No point in blocking this fix because of what might happen to OCA/department.
Can you rebase so that we can confirm that all is good for TravisCI?

dreispt:
Sorry, just noticed that this targets 7.0, and Travis build are not clean there.
Do you plan making a forward port to 8.0?

analytic_hours_block/hours_block_menu.xml
Olivier-LAURENT, line 15: @dreispt  I usually specify these 2 references as soon as they are known.
Otherwise in debug mode the direct access to their definition first renders empty views ! We need to play with view mode to recover the expected behavior... not very convenient... I agree just a developper consideration.
In what way is it easier to extend the module without specify them ? Thx

dreispt, line 15: It allows you to create a new view with a lower priority, that will then be used instead of the "standard" on without touching it nor the act_window.
But AFAIK there is no common agreement that it should be a guideline, so it's just my opinion.

dreispt:
Reviews needed.

dreispt:
@Olivier-LAURENT Travis build failed, can you check please?

NL66278:
:+1:  LGTM. I think not solving this issue is keeping builds in this repository in the red.

dreispt:
@NL66278 It can't get worse than it it, can it?
I'm merging this in hope things improve.

PR #91, opened 2015-06-09
[MIG] Migrate analytic_hours_blocks and project_classification

charbeljc:


dreispt:
The commit history seems to be polluted with other unrelated changes. You should rewrite this PR.

PR #92, opened 2015-06-09
[MIG] Migrate project_closing

charbeljc:

PR #95, opened 2015-06-15
[WIP] [MIG] project_closing, analytic_hours_block, project_classification

charbeljc:
obsoletes #91, #92


dreispt:
@charbeljc I don't want to sound annoying, but closing your other PRs and openeing a new one concentrating them is not a good idea: it's much better to have a few small PRs than having a single big PR. And moving to new PRs loses the reviewers comment history, which is not nice.

You can redo your PRs from scratch withou changing branch or PR, with something like this:

```
git checkout 8.0
git pull upstream 8.0
git checkout my-branch
git reset 8.0
```

This will end with out branch history synced with the latest 8.0 branch, but preserving your local changes - you can now use diff to find them and make a commit for your PR. Then:

```
git push -f origin my-branch
```

And your PR will be rewritten.

charbeljc:
@dreispt thanks for your remarks. I'm currently brute forcing an instance under continuous integration for migration to 8.0. From your comments, I understand that merging my two PR was a bad move. I'll redo this with three distinct PR, one per module as soon as I have progressed on my port (eg. green status on travis for my instance).

dreispt:
Thank you, it's best that way. If you have dependencies between them you should mention that on the PR description, so that we know what should be reviewed and merged first. We can then force the TravisCI build to run again and confirm if it changes to green.

dreispt:
Replaced by #96, #97 and #98

PR #97, opened 2015-06-18
[MIG] port project_classification to 8.0

charbeljc:
Redoing #95 in three separate PRs, one per ported module.


project_classification/__openerp__.py
```diff
@@ -19,16 +19,19 @@
...
     'version': '1.0',
     'category': 'Generic Modules/Projects & Services',
     'description': """
 
-This Module allow you to setup different project classification to ease the data entry of
-new project. The parent project will be set as readonly to forbid users to change it.
-The parent is still available through the analytic account object. This is useful because
-this way, project manager will setup correctly the analytical account just by choosing the
-corresponing classification.
+This Module allow you to setup different project classification to ease the
```
pedrobaeza: Extract this to a README.rst file with OCA's template: https://github.com/OCA/maintainer-tools

project_classification/__openerp__.py
```diff
@@ -52,7 +55,7 @@
     ],
     'demo_xml': [],
     'test': [],
-    'installable': False,
+    'installable': True,
     'auto_install': False,
     'application': True,
```
pedrobaeza: This is not an application

project_classification/__openerp__.py
```diff
@@ -52,7 +55,7 @@
     ],
     'demo_xml': [],
```
pedrobaeza: Remove this deprecated entry

project_classification/project_classification.py
```diff
@@ -20,22 +20,31 @@
 ##############################################################################
 from openerp.osv import orm, fields
 
+
 class ProjectClassification(orm.Model):
     _name = "project.classification"
     _description = "Project classification"
 
-    _columns ={
+    _columns = {
```
pedrobaeza: What about passing this to new API? It's very short

project_classification/project_classification.py
```diff
@@ -52,24 +61,33 @@ def _child_project_compute(self, cr, uid, ids, name, arg, context=None):
...
+                        "invoice factor,..)",
+                        required=True),
+        'child_project_complete_ids':
+        fields.function(_child_project_compute,
+                        relation='project.project',
+                        method=True,
+                        string="Project Hierarchy", type='many2many'),
         }
 
 # vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
pedrobaeza: Remove this editor-dependent line

project_classification/project_classification_view.xml
```diff
@@ -46,10 +46,10 @@
       <field name="model">project.project</field>
       <field name="inherit_id" ref="project.edit_project" />
       <field name="arch" type="xml">
-        <xpath expr="/form/sheet/notebook/page[@string='Other Info']/group/group[3]/field[@name='parent_id']" position="replace">
+        <xpath expr="/form//page[@string='Other Info']//field[@name='parent_id']" position="replace">
```
pedrobaeza: Don't make `position="replace"`. Use `position="attributes"`

charbeljc: Something like that ?

``` xml
<field name="classification_id" position="attributes">
      <attribute name="on_change">onchange_classification_id(classification_id)</attribute>
</field>
```

This looks much nicer, thanks you @pedrobaeza

charbeljc:
Travis will fail because this PR depends on OCA/project-reporting/pull/4

damdam-s:
depends on https://github.com/OCA/project-reporting/pull/5

dreispt:
Why is the new dependency? Should reporting features sit on top of functional features, instead of the other way around?

mdietrichc2c:
@dreispt from what I get, the module was already dependent on `project_billing_utils` (which adds the field `project_ids` on the analytic account, and an action using this field), but it was not defined in the `__openerp__.py`.

I do agree though that a "functional" module should not depend on a reporting module, so I would rather move the field/action definition in a separate module (which will be in both `project_classification`and `project_billing_utils` dependencies, to avoid losing functionalities in the second one).

@charbeljc , @damdam-s : any opinion on the subject?

damdam-s:
@mdietrichc2c :+1:  for a new module on which 'project_classification' and 'project_billing_utils' depends

charbeljc:
@mdietrichc2c, @damdam-s restoring dependency on project_billing_utils for now for correctness, but I completely agree with you :+1:

charbeljc:
Just rebased and updated module's dependency to `project_analytic_line_view`, thanks @mdietrichc2c !

mdietrichc2c:
:+1: with the new dependency

project_classification/README.rst
```diff
@@ -0,0 +1,66 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Project classification (easy hierarchy and setup for project managers)
+======================================================================
+
+
+This Module allow you to setup different project classification to ease the
```
pedrobaeza: s/allow/allows

project_classification/README.rst
```diff
@@ -0,0 +1,66 @@
...
+data entry of new project. The parent project will be set as readonly to
+forbid users to change it.
+
+The parent is still available through the analytic account object.
+This is useful because this way, project manager will setup correctly
+the analytical account just by choosing the corresponing classification.
```
pedrobaeza: s/corresponing/corresponding

project_classification/README.rst
```diff
@@ -0,0 +1,66 @@
...
+
+A project classification is composed by :
+
+ * A name
+ * An Analytic Account which represent the parent project to set
```
pedrobaeza: s/represent/represents

pedrobaeza:
An easier name would be simply _project_type_

dreispt:
@pedrobaeza This is a ported module, and unless there is a strong reason otherwise, it's best to keep the original name.

mdietrichc2c:
Seems like the runbot failed because existing projects don't get a default value. Is it possible to add a default one?
edit: seems a bit complicated to set a default classification... Would it be better to make the classification non-mandatory in the field, but mandatory in the view (in order to force setting it on new products) ?

mdietrichc2c:
All okay, all green, :+1:

project_classification/README.rst
```diff
@@ -0,0 +1,66 @@
...
+
+This Module allows you to setup different project classification to ease the
+data entry of new project. The parent project will be set as readonly to
+forbid users to change it.
+
+The parent is still available through the analytic account object.
+This is useful because this way, project manager will setup correctly
+the analytical account just by choosing the corresponding classification.
+
+A project classification is composed by :
```
yvaucher: No space before `:` in english

project_classification/README.rst
```diff
@@ -0,0 +1,66 @@
...
+ * An optional Pricelist
+ * An optional Currency
+
+Those values will be set on a project when selecting a classification.
+
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
```
yvaucher: You can remove this mention

project_classification/README.rst
```diff
@@ -0,0 +1,66 @@
...
+`here <https://github.com/OCA/project-service/issues/new?body=module:%20project_classification%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Yannick Vaucher <yannick.vaucher@camptocamp.com>
```
yvaucher: Please add Joël Grand-Guillaume first he is the main author

project_classification/README.rst
```diff
@@ -0,0 +1,66 @@
...
+* Daniel Reis <dgreis@sapo.pt>
```
yvaucher: You might not want to add @dreispt here, his 2 commits are moving the module to `__unported__` and deactivate it.

project_classification/__openerp__.py
```diff
@@ -19,42 +19,26 @@
 #
 ##############################################################################
 {
-    'name': 'Project classification (easy hierarchy and setup for project managers)',
+    'name': "Project classification (easy hierarchy and setup "
+            "for project managers)",
     'version': '1.0',
```
yvaucher: Please use full version number `8.0.1.0`

project_classification/project_classification.py
```diff
@@ -0,0 +1,77 @@
...
+#
+##############################################################################
+from openerp import api, models, fields
+
+
+class ProjectClassification(models.Model):
+    _name = "project.classification"
+    _description = "Project classification"
+
+    name = fields.Char('Classification Name', required=True, size=64)
```
yvaucher: You can remove `size=64`

project_classification/project_classification.py
```diff
@@ -0,0 +1,77 @@
...
+                                 "a project.",
+                                 required=True)
+    to_invoice = fields.Many2one('hr_timesheet_invoice.factor',
+                                 'Reinvoice Costs',
+                                 help="Fill this field if you plan to "
+                                 "automatically generate invoices based "
+                                 "on the costs in this classification")
+    currency_id = fields.Many2one('res.currency', 'Currency')
+    user_id = fields.Many2one('res.users', 'Account Manager')
+    pricelist_id = fields.Many2one('product.pricelist',
```
yvaucher, line 42: Shouldn't this have a domain on type of pricelist = sale?

project_classification/project_classification.py
```diff
@@ -0,0 +1,77 @@
...
+                                        "of project (like pricelist, "
+                                        "invoice factor,..)",
+                                        required=False)
+
+    child_project_complete_ids = fields.Many2many(
+        'project.project',
+        compute='_child_project_compute',
+        string="Project Hierarchy")
+
+    @api.depends('child_complete_ids.project_ids')
```
yvaucher, line 65: as you loop on self, you could use @api.one

project_classification/project_classification.py
```diff
@@ -0,0 +1,77 @@
...
+    def _child_project_compute(self):
+        for project in self:
+            child_projects = project.mapped('child_complete_ids.project_ids')
+            project.child_project_complete_ids = child_projects
+
+    def onchange_classification_id(self, cr, uid, ids, classification_id):
```
yvaucher: Any reason to keep it in old API ? Would be better with new API as it would make it simple to inherit

project_classification/project_classification_view.xml
```diff
@@ -46,10 +48,10 @@
       <field name="model">project.project</field>
       <field name="inherit_id" ref="project.edit_project" />
       <field name="arch" type="xml">
-        <xpath expr="/form/sheet/notebook/page[@string='Other Info']/group/group[3]/field[@name='parent_id']" position="replace">
-          <field name="classification_id" widget="selection" on_change="onchange_classification_id(classification_id)"/>
+        <xpath expr="/form/sheet/notebook/page[@string='Other Info']//field[@name='parent_id']" position="replace">
+          <field name="classification_id" required="1" widget="selection" on_change="onchange_classification_id(classification_id)"/>
```
yvaucher: Why is it required in view but not on model?

yvaucher:
@charbeljc Thanks for the porting. Please fix the few things I mentioned.

rafaelbn:
@charbeljc I have tried to test module, I feel readme is incomplete. **Please** add Installation, configuration and Usage as described here: https://github.com/OCA/maintainer-tools/tree/master/template/module
**Usage** is a very important part for this kind of module which is not an easy one.
**Thanks**

project_classification/project_classification_view.xml
damdam-s: https://github.com/OCA/project-service/pull/97#issuecomment-130589888
if anyone has a better solution ...

mdietrichc2c: @yvaucher : if there was a way to define a default value, it should be on the model; but otherwise, like in this case, installing the module with existing projects will raise an error when the NOT NULL constraint will be added.

project_classification/README.rst
```diff
@@ -0,0 +1,64 @@
...
+ * An Analytic Account which represents the parent project to set
+ * An optional Invoice factor
+ * An optional Account Manager
+ * An optional Pricelist
+ * An optional Currency
+
+Those values will be set on a project when selecting a classification.
+
+With this module, you can easily define invoice factor, manager, ... which will be automatically setted when creating a project with this classification.
+No need to check which parent project, which invoicing factor, which mamanger.
```
gurneyalex: s/mamanger/manager/

project_classification/README.rst
```diff
@@ -0,0 +1,64 @@
...
+ * A name
+ * An Analytic Account which represents the parent project to set
+ * An optional Invoice factor
+ * An optional Account Manager
+ * An optional Pricelist
+ * An optional Currency
+
+Those values will be set on a project when selecting a classification.
+
+With this module, you can easily define invoice factor, manager, ... which will be automatically setted when creating a project with this classification.
```
gurneyalex: s/setted/set/

damdam-s:
fixed

project_classification/project_classification_view.xml
yvaucher: @mdietrichc2c I see an other option, ensuring that existing project gets a classification_id in an init script like we did for some other modules.

Here is an exemple using a pre_init_hook to fill a required field with a default value for old records:

https://github.com/OCA/stock-logistics-workflow/blob/8.0/stock_ownership_availability_rules/__openerp__.py#L32
https://github.com/OCA/stock-logistics-workflow/blob/8.0/stock_ownership_availability_rules/__init__.py#L4-L9

charbeljc:
Thanks @damdam-s ! :+1:

lepistone:
Ping me for a merge once the comment from @yvaucher is addressed. Thanks!

project_classification/project_classification_view.xml
mdietrichc2c: @yvaucher yes, but then we are back on the original issue: there is no default classification present to be set on existing projects.

It's probably best to use the pre_init_hook to create a "old project" classification if some exists, and set in on every existing project; I think in that case, the runbot will not be red (as a value was set on existing projects), and the "default" value will not be used for new projects.

yvaucher: The best way to go would be to create a classification name something like: "Old project to classify" that is created only if projects are already present in database.

It might be needed to create the column already in pre_init_hook in order to populate it before adding the constraint.

guewen: Agree with @yvaucher.
An idea discussed here: an init/pre-migration script that:
- returns if there is no record or all records have a classification_id
- create the column classification_id as nullable
- create a classification named "Unclassified"
- update classification_id with the new ID
- set classification_id NOT NULL

damdam-s: @guewen @yvaucher @lepistone done

lepistone:
@damdam-s thanks, but still red :sob:

project_classification/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+            try:
+                query_classification_id = (
+                    'SELECT count(id) from project_project '
+                    'where classification_id is null;'
+                    )
+                cr.execute(query_classification_id)
+                res2 = cr.fetchall()
+                if res[0][0] == res2[0][0]:
+                    # all project have a classification_id set
+                    return
```
guewen: I brought the confusion, but I think it is safe to replace the condition here by this one:

``` python
cr.execute("select exists(select * from information_schema.tables where table_name = 'project_classification')")
if not cr.fetchone():
   # all the stuff in the `except` block
```

Because if the table does not exist, the classification_id field neither (and further, as we install this module, the table should never happen to exist prior to the installation but we can keep the condition as a safeguard).

project_classification/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+##############################################################################
+from . import project_classification
+import psycopg2
+
+
+def fill_classification_id(cr):
+        query_nb_records = 'SELECT count(id) from project_project;'
+        cr.execute(query_nb_records)
+        res = cr.fetchall()
+        if res[0][0] != 0:
```
guewen: Could be simplified to:

``` python
cr.execute("SELECT id FROM project_project LIMIT 1")
if not cr.fetchone():
    return  # we won't violate the not-null constraint on classification_id
```

(also returning eagerly is better than returning in the `else` block because you avoid one indentation level in the whole body of the method)

project_classification/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+                    'where classification_id is null;'
+                    )
+                cr.execute(query_classification_id)
+                res2 = cr.fetchall()
+                if res[0][0] == res2[0][0]:
+                    # all project have a classification_id set
+                    return
+
+            except psycopg2.ProgrammingError:
+                cr.rollback()
```
guewen: What do you expect to rollback? It's gonna blow your database if you rollback what has been done before the init_hook is executed (anyway the `except` could be removed as proposed in my other comment).

project_classification/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+                );"""
+                cr.execute(query)
+                query = (
+                    "INSERT INTO project_classification (name, project_id) "
+                    "VALUES ('Unclassified', "
+                    "(select id from project_project limit 1));"
+                    )
+                cr.execute(query)
+                # 3. assign this classification to all projects
+                query = 'select id from project_classification limit 1;'
```
guewen: You can get the last inserted ID with:

``` python
cr.execute(
    "INSERT INTO project_classification (name, project_id) "
    "VALUES ('Unclassified', "
    "(select id from project_project limit 1)) RETURNING id;"
)
classification_id = cursor.fetchone()[0]
```

damdam-s:
Finally, classification_id in project not required because it's only an helper to fill other fields.

guewen:
Thanks :+1:

mdietrichc2c:
:+1: , not requiring the field at all is probably the best solution

yvaucher:
:+1: Thanks (travis error unrelated)

PR #100, opened 2015-07-23
[MIG] Migrated <sale_project_base>

oihane:


dreispt:
LGTM :+1:

pedrobaeza:
I prefer to have a project field that it's computed instead of this extra layer with an onchange that it's not triggered when creating programatically sales orders. Also the name doesn't seem to be the best for the functionality included. I prefer sale_order_project.

oihane:
And what about the button?

@dreispt what do you think about @pedrobaeza approach...

oihane:
Finally, I made the changes as required by @pedrobaeza 
- Rename module
- related_project_id computed field

sale_order_project/models/sale.py
```diff
@@ -22,48 +22,45 @@
...
-            project_id = project_obj.create(cr, uid, vals, context=context)
-            project = project_obj.browse(cr, uid, project_id, context=context)
+    @api.multi
+    def action_create_project(self):
+        project_obj = self.env['project.project']
+        for order in self:
+            vals = order._prepare_project_vals()
+            project = project_obj.create(vals)
             order.write({
-                'related_project_id': project_id,
```
qtheuret: 'project_id' variable doesn't exist. I think you would write 'project.id'.

qtheuret:
LGTM :+1:

sale_order_project/__openerp__.py
```diff
@@ -21,23 +21,19 @@
...
 {
-    'name': 'Sale Project Base',
+    'name': 'Sale Order Project',
     'version': '7.0',
     'category': 'Generic Modules/Others',
     'license': 'AGPL-3',
-    'description': """This module is a base module that give the possibility to
-    create a project from a quotation""",
-    'author': "Akretion,Odoo Community Association (OCA)",
+    'author': 'Akretion, Odoo Community Association (OCA)',
```
pedrobaeza: Expand authors

sale_order_project/models/sale.py
```diff
@@ -22,48 +22,45 @@
...
-            'Project',
-            readonly=True,
-            states={'draft': [('readonly', False)]}
-        ),
-    }
+    @api.one
+    @api.depends('project_id')
+    def _compute_related_project_id(self):
+        if self.project_id.use_tasks:
+            related_project_id = self.env['project.project'].search(
```
pedrobaeza: Make directly:

```
self.related_project_id = self.env['project.project'].search([('analytic_account_id', '=', self.project_id.id)], limit=1)[:1]
```

without making related_project_id = False, because this is set by default in ORM

sale_order_project/models/sale.py
```diff
@@ -22,48 +22,45 @@
...
-        if related_project_id:
-            project = project_obj.browse(cr, uid, related_project_id, context=context)
-            return {'value': {'project_id': project.analytic_account_id.id}}
-        return {}
+    related_project_id = fields.Many2one(
+        comodel_name='project.project', string='Project',
+        compute='_compute_related_project_id')
 
-    def _prepare_project_vals(self, cr, uid, order, context=None):
+    def _prepare_project_vals(self):
```
pedrobaeza: Decorate it with @api.model and don't lose `order` argument

pedrobaeza: Is this method really needed?

pedrobaeza:
Please exercise the functionality with some tests: one to check the computed project and another (if still valid) to check the project name.

sale_order_project/__openerp__.py
```diff
@@ -21,23 +21,19 @@
 #
 ###############################################################################
 
-
 {
-    'name': 'Sale Project Base',
+    'name': 'Sale Order Project',
     'version': '7.0',
```
dreispt: Hmm, edit this please. I suggest: `8.0.1.0`

sale_order_project/models/sale.py
```diff
@@ -22,48 +22,45 @@
...
 
-    def onchange_related_project_id(self, cr, uid, ids, related_project_id, context=None):
-        project_obj = self.pool['project.project']
-        if related_project_id:
-            project = project_obj.browse(cr, uid, related_project_id, context=context)
-            return {'value': {'project_id': project.analytic_account_id.id}}
-        return {}
+    related_project_id = fields.Many2one(
+        comodel_name='project.project', string='Project',
+        compute='_compute_related_project_id')
```
dreispt, line 43: Add the `readonly` and `states` attributes?

sale_order_project/models/sale.py
dreispt: @pedrobaeza It makes it easier to override the calculation of those default values

sale_order_project/models/sale.py
oihane, line 43: As it is computed, in the view is readonly, that's why I deleted the readonly and states parameters

dreispt, line 43: I see. That is a significant design change.
You should keep it a non computed field, as before.

oihane, line 43: I made that change because @pedrobaeza suggested

The button continues there, so you are able to create a project from the sale.order
If you add an analytic account with use_tasks checked it will add the related project.

pedrobaeza, line 43: I don't think is convenient to have a project and an unrelated analytic account in the same sales order. Do you have any use case?

oihane, line 43: Before this change, you weren't able to select an analytic account. The field was made invisible

pedrobaeza, line 43: So you were making impossible to select an analytic account that it's not link to a project, so I see this as a drawback

dreispt, line 43: So you pick an analytic account and then the (possible) corresponding project_id is calculated.
A calculated field makes sense then.

oihane:
Added some tests, as suggested by @pedrobaeza

sale_order_project/tests/test_sale_order_project.py
```diff
@@ -0,0 +1,33 @@
...
+class TestSaleOrderProject(common.TransactionCase):
+
+    def setUp(self):
+        super(TestSaleOrderProject, self).setUp()
+        self.sale_order_model = self.env['sale.order']
+        self.analytic_account_model = self.env['account.analytic.account']
+        self.sale_order = self.env.ref('sale.sale_order_2')
+        self.project = self.env.ref('project.project_project_1')
+
+    def testCreateProjectFromOrder(self):
```
pedrobaeza: Don't use CamelCase for this method names (there's only the special case of setUp)

oihane:
@pedrobaeza replaced camel case

pedrobaeza:
Thanks :+1:

PR #103, opened 2015-08-05
Added italian translation

sysadminmatmoz:
I closed the previous pull to commit a cleaner version.

### DIFF
--- a/project_gtd/i18n/it.po
+++ b/project_gtd/i18n/it.po
@@ -1,20 +1,21 @@
 # Translation of OpenERP Server.
 # This file contains the translation of the following modules:
-#	* project_gtd
+# 	* project_gtd
 #
 msgid ""
 msgstr ""
 "Project-Id-Version: OpenERP Server 6.0dev\n"
-"Report-Msgid-Bugs-To: support@openerp.com\n"
-"POT-Creation-Date: 2012-12-21 17:06+0000\n"
-"PO-Revision-Date: 2011-01-13 06:41+0000\n"
-"Last-Translator: Nicola Riolini - Micronaet <Unknown>\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2015-08-02 08:27+0200\n"
+"PO-Revision-Date: 2015-08-02 08:35+0200\n"
+"Last-Translator: Matjaz Mozetic <m.mozetic@matmoz.si>\n"
 "Language-Team: \n"
+"Language: it\n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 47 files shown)


project_gtd/i18n/it.po
```diff
@@ -407,7 +410,6 @@ msgstr "o"
 #~ msgid "This Month"
 #~ msgstr "Questo mese"
 
-#, python-format
```
pedrobaeza, line 410: All of these strings are commented, so they're useless. Maybe a good idea would be to regenerate italian file, syncing previously all the terms.

sysadminmatmoz:
Ok. I recreated the template (the existing one was kept since the version 6) and synced all the languages according to the new one. Now it should be clean.

pedrobaeza:
Wow, what a huge work

:+1:

sysadminmatmoz:
The question is if it has sense to keep such a number of translations since at least half of them is actually empty (translated less than 20%)

pedrobaeza:
Well, less is nothing. We are moving to Transifex, so this work will soon be easy.

sysadminmatmoz:
I know, I already translated in slovene nearly all OCA (v8) modules that are already there. :smile: and did the same with the Odoo ones (it's a lot to do if you are the "community" in your country).
But something should be done to ease the translation process by merging OCA and Odoo modules on transifex, otherwise the translated terms are not shared and you get cases, where the same term has a different translation in an Odoo module than it has in an OCA module since Transifex translation memory is not shared through different repositories as it was on launchpad - that's why I'm working on a custom poedit based platform in local - to have all the terms shared independently of the repo.

pedrobaeza:
That's difficult to achieve due to the translation architecture itself. You can see an example when you need to translate the "Last Updated by" field and so on on each module that declares a new model.

dreispt:
@pedrobaeza ...unless :bulb: an OCA script could translate empty strings copying from what was done on Odoo.

pedrobaeza:
Yeah, maybe we can check and make any injection, but for now that's not feasible.

dreispt:
:+1: 

By the way, this project [is already](https://www.transifex.com/oca/OCA-project-service-8-0/) on Transifex, but I'm not sure if the OCA Transifex sync is working.

pedrobaeza:
Github > Transifex is already working. Transifex > Github is on the way here: https://github.com/OCA/maintainer-tools/pull/69

sysadminmatmoz:
Should I close this PR?

pedrobaeza:
No, it's OK for now to do it this way.

PR #113, opened 2015-09-04
[IMP] project_task_materials: Develop consume materials from tasks.

sergio-teruel:


project_task_materials/README.rst
```diff
@@ -0,0 +1,81 @@
...
+activated and consume materials must be checked in stage task.
+
+Usage
+=====
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/140/8.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
+.. branch is "8.0" for example
```
dreispt, line 24: You can remove this comment

project_task_materials/__openerp__.py
```diff
@@ -1,43 +1,22 @@
...
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# (c) 2015 Antiun Ingeniería S.L. - Sergio Teruel
```
dreispt, line 2: You should also keep the original author's copyright

project_task_materials/__openerp__.py
```diff
@@ -1,43 +1,22 @@
...
+# (c) 2015 Antiun Ingeniería S.L. - Carlos Dauden
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
 {
     'name': 'Project Task Materials',
     'summary': 'Record products spent in a Task',
     'version': '1.0',
```
dreispt, line 9: Use `8.0.1.1.0` -> signals no migration, module upgrade needed

project_task_materials/__openerp__.py
```diff
@@ -1,43 +1,22 @@
...
-Field Service, often require you to keep a record of the materials spent.
-
-This module adds the ability to also this material spending.
-To use it, make sure the "Log work activities on tasks" Project setting is
-activated.
-
-Note that only a simple record is made and no accounting or stock moves are
-actually performed.""",
-    'depends': ['project', 'product'],
+    'description': "",
```
dreispt, line 15: remove description key

project_task_materials/__openerp__.py
```diff
@@ -1,43 +1,22 @@
...
+    'depends': ['project', 'stock', 'stock_account', 'analytic'],
```
dreispt, line 16: We have a problem here: we can't add the `stock` additional dependencies here.
It will force people that don't use it (+accounting) to have it installed.

You should propose this as a separate module.
I suggest `project_task_materials_stock`.
You can make it `auto_install` if both `project`and `stock`are installed.

Feel free to at the same time propose changes to `project_task_materials` that make things simpler for the new module to extend it.

sergio-teruel:
Following @dreispt suggest we create project_task_materials_stock. #115

rafaelbn:
> We have a problem here: we can't add the stock additional dependencies here.
> It will force people that don't use it (+accounting) to have it installed.

@dreispt OK but for you has any sense work with materials without stock?

dreispt:
You can use Project without using Odoo stock or accounting. I have such a case, where stock is handled by an external ERP.

PR #115, opened 2015-09-04
[ADD] project_task_materials_stock: New module to create stock and 

sergio-teruel:
analytic moves.

This PR close #113

### DIFF
--- a/project_task_materials_stock/README.rst
+++ b/project_task_materials_stock/README.rst
@@ -17,6 +17,9 @@ Configuration
 Check option 'Consume Material' in Task Stage to generate a stock move when
 the task is in that stage.
 
+Go to Setting -> Configuration - > Projects and enable option
+"Manage time estimation on tasks"
+
 Usage
 =====
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
--- a/project_task_materials_stock/README.rst
+++ b/project_task_materials_stock/README.rst
@@ -1,0 +1,84 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+============================
+Project Task Materials Stock
+============================
+Project Tasks allow to record time spent, but some activities, such as
+Field Service, often require you to keep a record of the materials spent.
+
+This module extends project_task_material module to consume products spent in 
+a task and create analytic lines to manage costs and create invoices.  
+
+Configuration
+=============
+
+Check option 'Consume Material' in Task Stage to generate a stock move when
+the task is in that stage.
+
+Usage
+=====
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/140/8.0
+
+There is a new field named "Consume Material" in "Task stage" to indicate
+when you want to consume materials.
+
+Known issues / Roadmap
+======================
+
+* Add support for budgeted/expected materials.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/project/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/project/issues/new?body=module:%20project_task_materials_stock%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+License
+=======
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published
+by the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/agpl-3.0-standalone.html>.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Rafael Blasco <rafabn@antiun.com>
+* Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+* Carlos Dauden <carlos@incaser.es>
+* Sergio Teruel <sergio@incaser.es>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
--- a/project_task_materials_stock/__openerp__.py
+++ b/project_task_materials_stock/__openerp__.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Antiun Ingeniería S.L. - Sergio Teruel
+# (c) 2015 Antiun Ingeniería S.L. - Carlos Dauden
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    'name': 'Project Task Materials Stock',
+    'summary': 'Create stock and analytic moves from '
+               'record products spent in a Task',
+    'version': '8.0.1.0.0',
+    'category': "Project Management",
+    'author': 'Antiun Ingeniería S.L., '
+              'Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'depends': ['stock_account', 'project_task_materials'],
+    'data': [
+        'data/project_task_materials_data.xml',
+        'views/project_view.xml',
+    ],
+    'installable': True,
+    'auto_install': True,
+}
--- a/project_task_materials_stock/data/project_task_materials_data.xml
+++ b/project_task_materials_stock/data/project_task_materials_data.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data noupdate="1">
+
+        <record id="analytic_journal_sale_materials" model="account.analytic.journal">
+            <field name="code">SALMAT</field>
+            <field name="name">Tasks Materials</field>
+            <field name="type">general</field>
+        </record>
+
+    </data>
+</openerp>
--- a/project_task_materials_stock/models/project.py
+++ b/project_task_materials_stock/models/project.py
@@ -1,0 +1,175 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Antiun Ingeniería S.L. - Sergio Teruel
+# (c) 2015 Antiun Ingeniería S.L. - Carlos Dauden
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from openerp import models, fields, api
+from openerp.tools.float_utils import float_round
+
+
+class ProjectTaskType(models.Model):
+    _inherit = 'project.task.type'
+
+    consume_material = fields.Boolean(
+        string='Consume Material',
+        help="If you mark this check, when a task goes to this state,"
+             "it will consume the associated materials")
+
+
+class Task(models.Model):
+    _inherit = "project.task"
+
+    def _compute_stock_move(self):
+        self.stock_move_ids = self.mapped('material_ids.stock_move_id')
+
+    def _compute_analytic_line(self):
+        self.analytic_line_ids = self.mapped('material_ids.analytic_line_id')
+
+    @api.depends('stock_move_ids.state')
+    def _check_stock_state(self):
+        if not self.stock_move_ids:
+            self.stock_state = 'pending'
+        elif self.stock_move_ids.filtered(lambda r: r.state == 'confirmed'):
+            self.stock_state = 'confirmed'
+        elif self.stock_move_ids.filtered(lambda r: r.state == 'assigned'):
+            self.stock_state = 'assigned'
+        elif self.stock_move_ids.filtered(lambda r: r.state == 'done'):
+            self.stock_state = 'done'
+
+    stock_move_ids = fields.Many2many(
+        comodel_name='stock.move', compute='_compute_stock_move',
+        string='Stock Moves')
+    analytic_line_ids = fields.Many2many(
+        comodel_name='account.analytic.line', compute='_compute_analytic_line',
+        string='Analytic Lines')
+    consume_material = fields.Boolean(related='stage_id.consume_material')
+    stock_state = fields.Selection(
+        [('pending', 'Pending'),
+         ('confirmed', 'Confirmed'),
+         ('assigned', 'Assigned'),
+         ('done', 'Done')], compute='_check_stock_state', string='Stock State')
+
+    @api.multi
+    def unlink_stock_move(self):
+        moves = self.mapped('stock_move_ids')
+        moves.filtered(lambda r: r.state == 'assigned').do_unreserve()
+        moves.filtered(
+            lambda r: r.state in ['waiting', 'confirmed', 'assigned']
+        ).write({'state': 'draft'})
+        moves.unlink()
+
+    @api.multi
+    def write(self, vals):
+        for task in self:
+            res = super(Task, self).write(vals)
+            if 'stage_id' in vals:
+                if task.stage_id.consume_material:
+                    task.material_ids.create_stock_move()
+                    task.material_ids.create_analytic_line()
+                else:
+                    task.unlink_stock_move()
+                    task.analytic_line_ids.unlink()
+        return res
+
+    @api.multi
+    def unlink(self):
+        self.unlink_stock_move()
+        self.analytic_line_ids.unlink()
+        return super(Task, self).unlink()
+
+    @api.multi
+    def action_assign(self):
+        self.mapped('stock_move_ids').action_assign()
+
+    @api.multi
+    def action_done(self):
+        self.mapped('stock_move_ids').action_done()
+
+
+class ProjectTaskMaterials(models.Model):
+    _inherit = "project.task.materials"
+
+    stock_move_id = fields.Many2one(
+        comodel_name='stock.move', string='Stock Move')
+    analytic_line_id = fields.Many2one(
+        comodel_name='account.analytic.line', string='Analytic Line')
+    product_uom = fields.Many2one(
+        comodel_name='product.uom', string='Unit of Measure')
+
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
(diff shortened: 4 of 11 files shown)


project_task_materials_stock/README.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Known issues / Roadmap
+======================
+
+* ...
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/{project_repo}/issues>`_.
```
pedrobaeza: You have to replace `{project_repo}` with `project`

project_task_materials_stock/README.rst
```diff
@@ -0,0 +1,74 @@
...
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/{project_repo}/issues/new?body=module:%20{module_name}%0Aversion:%20{version}%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
pedrobaeza: Replace placeholders by their values

project_task_materials_stock/README.rst
```diff
@@ -0,0 +1,74 @@
...
+along with this program. If not, see <http://www.gnu.org/licenses/agpl-3.0-standalone.html>.
+
+
+Credits
+=======
+
+Contributors
+------------
+
+* Rafael Blasco <rafabn@antiun.com>
```
pedrobaeza: As I designed this, I should be here also.

project_task_materials_stock/__openerp__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Antiun Ingeniería S.L. - Sergio Teruel
+# (c) 2015 Antiun Ingeniería S.L. - Carlos Dauden
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    'name': 'Project Task Materials Stock',
+    'summary': 'Create stock and analytic moves from '
+               'record products spent in a Task',
+    'version': '1.0',
```
pedrobaeza: Use version 8.0.1.0.0

project_task_materials_stock/data/project_task_materials_data.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data noupdate="1">
+
+        <record id="analytic_journal_sale_materials" model="account.analytic.journal">
+            <field name="code">SALMAT</field>
+            <field name="name">Consumed Materials</field>
```
pedrobaeza: I would call it "Tasks Materials"

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,151 @@
...
+from openerp import models, fields, api
+from openerp.tools.float_utils import float_round
+
+
+class ProjectTaskType(models.Model):
+    _inherit = 'project.task.type'
+
+    consume_material = fields.Boolean(
+        string='Consume Material',
+        help="""If you select this option, when a task is in this state
```
pedrobaeza: Don't use triple quota, but:

```
    help="If you select this option, when a task is in this state"
         "consumes materials associated"
```

pedrobaeza: Rephrase it to: "If you mark this check, when a task goes to this state, it will consume the associated materials"

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,151 @@
...
+        consumes materials associated""")
+
+
+class Task(models.Model):
+    _inherit = "project.task"
+
+    def _get_stock_move(self):
```
pedrobaeza: Call all compute methods with `compute` word, not `get` or `check`

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,151 @@
...
+        comodel_name='account.analytic.line', string='Analytic Line')
+
+    def _prepare_stock_move(self):
+        product = self.product_id
+        res = {
+            'product_id': product.id,
+            'name': product.partner_ref,
+            'state': 'confirmed',
+            'product_uom': product.uom_id.id,
+            'product_uos': product.uos_id and product.uos_id.id or False,
```
pedrobaeza: If you are not putting the same UoM in both fields (uom and uos), you cannot put the same quantity without calculating the converted quantity. Please put here also `product.uom_id.id` for simplicity (you are not going to work with another UoM in this case I think). You can even mention this question in known issues: "You cannot use another UoM in tasks except the product stock one".

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,151 @@
...
+            'product_uom_qty': self.quantity,
+            'product_uos_qty': self.quantity,
+            'origin': self.task_id.name,
+            'location_id':
+            product.product_tmpl_id.property_stock_procurement.id,
+            'location_dest_id': self.env.ref(
+                'stock.stock_location_customers').id,
+        }
+        if product.uos_id and product.uom_id and \
```
pedrobaeza: hehe, I have say nothing in the previous comment. Here you do the conversion, so perfect in this way. But now I see that you should ask for which UoM to use, because maybe people want to work with another one than UoS.

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,151 @@
...
+            'project_task_materials_stock.analytic_journal_sale_materials')
+        res = {
+            'name': self.task_id.name + ': ' + product.name,
+            'ref': self.task_id.name,
+            'product_id': product.id,
+            'journal_id': journal.id,
+            'unit_amount': self.quantity,
+            'account_id': self.task_id.project_id.analytic_account_id.id,
+            'to_invoice':
+            self.task_id.project_id.analytic_account_id.to_invoice.id,
```
pedrobaeza: This field is provided by _hr_timesheet_invoice_ module, so you shouldn't assume that exists. If you don't want to do a glue module, you can do 2 things:
- See if there's a `_prepare_*` or an onchange method on standard that puts these values via inheritance when an analytic account is passed.
- Use attrgetter method to bypass errors in case there's no such value.

project_task_materials_stock/README.rst
dreispt: I see your name 3 lines below :smile_cat:

pedrobaeza: This has been added afterwards, and that's why the comment is not gone :wink:

project_task_materials_stock/README.rst
```diff
@@ -0,0 +1,75 @@
...
+Usage
+=====
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/140/8.0
+
+Known issues / Roadmap
+======================
+
+* ...
```
dreispt: Roadmap: add support for budgeted/expected materials.

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,169 @@
...
+            if move.state == 'assigned':
+                move.do_unreserve()
+            if move.state in ['waiting', 'confirmed', 'assigned']:
+                move.state = 'draft'
+            move.unlink()
+
+    @api.one
+    def write(self, vals):
+        res = super(Task, self).write(vals)
+        if 'stage_id' in vals:
```
dreispt: The logic in `super()` could change the Stage.
We probably should rebrowse the current records and then use `if rec.stage_id in vals:`

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,169 @@
...
+            if self.consume_material:
```
dreispt: I fell this is a little obfuscating. It's longer but clearer to use `if self.stage_id.consume_material`

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,169 @@
...
+        return res
+
+    @api.multi
+    def action_assign(self):
+        self.stock_move_ids.action_assign()
+
+    @api.multi
+    def action_done(self):
+        self.stock_move_ids.action_done()
+        pass
```
dreispt: These `action_*` methods are not used anymore.
What is expected for them? Can they simply be removed?

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,169 @@
...
+            'account.analytic.line')
+        journal = self.env.ref(
+            'project_task_materials_stock.analytic_journal_sale_materials')
+        res = {
+            'name': self.task_id.name + ': ' + product.name,
+            'ref': self.task_id.name,
+            'product_id': product.id,
+            'journal_id': journal.id,
+            'unit_amount': self.quantity,
+            'account_id': self.task_id.project_id.analytic_account_id.id,
```
dreispt: For "Service Desk" use cases, the Analytic Account comes from an analytic_id in the Task object:
Projects are "Service Teams", and Tasks are Work Orders for specific contracts.
If possible, I would like to factor that in here: check for the existence on an `analytic_id` field in the Task, and use it if it exists.

dreispt:
Travis error is unrelated, so CI :ok:

project_task_materials_stock/__openerp__.py
```diff
@@ -0,0 +1,21 @@
...
+    'category': "Project Management",
+    'author': 'Antiun Ingeniería S.L., '
+              'Odoo Community Association (OCA)',
+    'license': 'AGPL-3',
+    'depends': ['stock_account', 'project_task_materials'],
+    'data': [
+        'data/project_task_materials_data.xml',
+        'views/project_view.xml',
+    ],
+    'installable': True,
```
dreispt, line 20: Is `'auto_install': True,` a bad idea?

project_task_materials_stock/models/project.py
carlos-incaser: This action_* methods are created as hooks, but I can remove this...

project_task_materials_stock/models/project.py
carlosdauden: The aim of this is that the moves are only created or deleted when the stage of the task has been modified.
Otherwise moves would be created in each write.

project_task_materials_stock/models/project.py
pedrobaeza: These methods are triggered by buttons, and they're intended for making the task to act as a "picking", where you have the check availability and transfer buttons to perform these operations for all the materials added. You should remove `pass` though.

project_task_materials_stock/README.rst
carlosdauden: Can you explain this support? :smile:

project_task_materials_stock/__openerp__.py
carlos-incaser: I'm sorry. Copy/Paste error...

project_task_materials_stock/models/project.py
carlosdauden: :+1:

project_task_materials_stock/__openerp__.py
sergio-teruel, line 20: auto-install :+1:

project_task_materials_stock/models/project.py
dreispt: I can see that, the point is that the related field shortcut makes it needlessly less clear to read.

project_task_materials_stock/models/project.py
dreispt: I see now. I was confused by the method names, can we make them closer to what they are for?
- "action_assign" is for  "Check (Stock) Availability"
- "action_done" is for "Transfer Materials"

project_task_materials_stock/models/project.py
dreispt: I was thinking about it, and I guess that the Timesheets integrated with Tasks surely have the same problem.
Given this, I'm wondering if the design approach for the Service Desk feature is the best one.
Probably this deserves a discussion on it's own...

project_task_materials_stock/models/project.py
sergio-teruel: Yes...  of course :grimacing:

rafaelbn:
Rebuilding for testing.

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,170 @@
...
+        string='Consume Material',
+        help="If you mark this check, when a task goes to this state,"
+             "it will consume the associated materials")
+
+
+class Task(models.Model):
+    _inherit = "project.task"
+
+    def _compute_stock_move(self):
+        move_ids = [line.stock_move_id.id for line in self.material_ids
```
pedrobaeza: Use `self.mapped('material_ids.stock_move_id')`

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,170 @@
...
+                    if line.stock_move_id]
+        self.stock_move_ids = self.env['stock.move'].browse(move_ids)
+
+    def _compute_analytic_line(self):
+        line_ids = [line.analytic_line_id.id for line in self.material_ids
```
pedrobaeza: Use `self.mapped('material_ids.analytic_line_id')`

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,170 @@
...
+        if not self.stock_move_ids:
+            self.stock_state = 'pending'
+        elif self.stock_move_ids.filtered(lambda r: r.state == 'confirmed'):
+            self.stock_state = 'confirmed'
+        elif self.stock_move_ids.filtered(lambda r: r.state == 'assigned'):
+            self.stock_state = 'assigned'
+        elif self.stock_move_ids.filtered(lambda r: r.state == 'done'):
+            self.stock_state = 'done'
+
+    stock_move_ids = fields.One2many(
```
pedrobaeza: Shouldn't this be Many2many?

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,170 @@
...
+        comodel_name='account.analytic.line', compute='_compute_analytic_line',
+        string='Analytic Lines')
+    consume_material = fields.Boolean(related='stage_id.consume_material')
+    stock_state = fields.Selection(
+        [('pending', 'Pending'),
+         ('confirmed', 'Confirmed'),
+         ('assigned', 'Assigned'),
+         ('done', 'Done')], compute='_check_stock_state', string='Stock State')
+
+    @api.one
```
pedrobaeza: Don't make it @api.one because is deprecated in v9. Use @api.multi and iterate over the recordset:

```
for record in self:
```

pedrobaeza: Or in this case, you can directly use mapped for iterating in all stock.move:

```
moves = self.mapped('stock_move_ids')
for move in moves:
```

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,170 @@
...
+    def unlink_stock_move(self):
+        for move in self.stock_move_ids:
+            if move.state == 'assigned':
+                move.do_unreserve()
+            if move.state in ['waiting', 'confirmed', 'assigned']:
+                move.state = 'draft'
+            move.unlink()
```
pedrobaeza: Do this in one operation at the end: `moves.unlink()` (code with the new approach)

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,170 @@
...
+
+    @api.one
```
pedrobaeza: Don't use @api.one also. And in this case, it's worst because your are forcing to serialize a write over multiple records.

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,170 @@
...
+                self.material_ids.create_stock_move()
+                self.material_ids.create_analytic_line()
+            else:
+                self.unlink_stock_move()
+                self.analytic_line_ids.unlink()
+        return res
+
+    @api.multi
+    def action_assign(self):
+        self.stock_move_ids.action_assign()
```
pedrobaeza: This will raise error when multiple records. Use instead:

```
self.mapped('stock_move_ids').action_assign()
```

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,170 @@
...
+
+    @api.multi
+    def action_done(self):
+        self.stock_move_ids.action_done()
```
pedrobaeza: The same

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,170 @@
...
+            'product_uos_qty': self.quantity,
+            'origin': self.task_id.name,
+            'location_id':
+            product.product_tmpl_id.property_stock_procurement.id,
+            'location_dest_id': self.env.ref(
+                'stock.stock_location_customers').id,
+        }
+        return res
+
+    @api.one
```
pedrobaeza: Don't use @api.one

rafaelbn:
Tested: (you can check here: http://3116482-115-2d3873.runbot1.odoo-community.org)

1 - Add instruction un usage to explain that there is a new field named "Consume Material" in "Task stage" to indicate when you want to consume materials.

2 - Error. Not cosuming material.
- Create a product "motor" stockable and set in the inventory 10 unit, you have 10 motors.
- Create a task "repair car" in a project "car shop" (for example!)
- In task, Input 1 hour
- In task, input 1 motor
- Change stage in task which has stage marked as "consume material"
- You can check that products are not cosumed (they must be)

rafaelbn:
@dreispt @pedrobaeza your comments have beed done and fixed, please review if we can remove label "needs fixing" and set "need review". Thanks

pedrobaeza:
That's because move has waiting availability state. One thing you can do is to raise a Warning if the material can be consumed due to it hasn't been reserved.

sergio-teruel:
@rafaelbn @pedrobaeza Fixed sotck location.

project_task_materials_stock/README.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Configuration
+=============
+
+Check option 'Consume Material' in Task Stage to generate a stock move when
+the task is in that stage.
+
+Usage
+=====
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
```
rafaelbn, line 25: Add instruction un usage to explain that there is a new field named "Consume Material" in "Task stage" to indicate when you want to consume materials

sergio-teruel, line 25: @rafaelbn Change done.

pedrobaeza:
Please rebase to get green tests

sergio-teruel:
@pedrobaeza Rebase done

project_task_materials_stock/README.rst
```diff
@@ -0,0 +1,84 @@
...
+
+This module extends project_task_material module to consume products spent in 
+a task and create analytic lines to manage costs and create invoices.  
+
+Configuration
+=============
+
+Check option 'Consume Material' in Task Stage to generate a stock move when
+the task is in that stage.
+
```
rafaelbn, line 19: - Go to Setting -> Configuration - > Projects and enable option "Manage time estimation on tasks"

rafaelbn:
Tested in runbot :+1: , please change readme with comment and check travis, thanks.

sergio-teruel:
@rafaelbn changes done..

rafaelbn:
In my opinion this module is ready to be merged but we need more reviews. Thanks

ingenieriacloud:
:+1:  functional

janetrp:
tested in runbot :+1:

dreispt:
Congratulations! It got through the review pinhole :smile:

PR #119, opened 2015-10-13
Warn if Project has no calendar set up. Fixes #73

### LINKED ISSUE #73: SLA calculation hangs whens no Working calendar is provided
I traced it to `project_sla_control.py`, function `_compute_sla_date()`.
Maybe I'll fix it when porting to the new API.
The workaround is to ensure a Working Calendar is set for the SLA controlled Projects.

dreispt:


project_sla/project_sla_control.py
```diff
@@ -171,6 +171,13 @@ def datetime2str(dt_value, fmt):  # tolerant datetime to string
                    safe_getattr(doc, 'project_id.analytic_account_id.sla_ids'))
         if not sla_ids:
             return res
+        cal_obj = safe_getattr(doc, 'project_id.resource_calendar_id')
+        if not cal_obj:
+            _logger.warning('Project %s has no calendar!', doc.project_id.name)
```
pedrobaeza: Log message instead of raising an exception?

project_sla/project_sla_control.py
```diff
@@ -171,6 +171,13 @@ def datetime2str(dt_value, fmt):  # tolerant datetime to string
                    safe_getattr(doc, 'project_id.analytic_account_id.sla_ids'))
         if not sla_ids:
             return res
+        cal_obj = safe_getattr(doc, 'project_id.resource_calendar_id')
```
pedrobaeza: Shouldn't this be called `calendar`?

dreispt:
The module is still in the old API., so Travis fails because of the new pylint-odoo PR checks.
The conversion is not trivial, will require refactoring and testing, and I don't have the time to do it know.
However, I believe that the fix is worth merging asap, and it wont' break the branch build.

pedrobaeza:
pylint-odoo is still not enabled, but although it would be, it only raises warning, but don't make PRs to fail. The error is in codacy

dreispt:
OK, should be good to go then!

dreispt:
I addressed Codacy's complain.

yvaucher:
:+1:

dreispt:
Ping!

pedrobaeza:
:+1:

PR #121, opened 2015-10-16
[ADD][8.0] sale_service_project: Module to create projects and task with

sergio-teruel:
works and materials.


dreispt:
This sits in the borderline, but since it adds a feature to `sale` linking it to `project` I believe it's best hosted in the sales repo.
More opinions?

sergio-teruel:
@dreispt Ok. I move it to sale-workflow

PR #122, opened 2015-10-19
 action_create_task

ahilali:


pedrobaeza:
This is not new API, so the call is correct

:-1:

dreispt:
@ahilali Though it would be a nice contribution to port this to the new API :smile_cat:

bealdav:
@pedrobaeza @dreispt 
The bug is real in v8
  File "/workspace/parts/odoo/addons/web/controllers/main.py", line 948, in call_button
    action = self._call_kw(model, method, args, {})
  File "/workspace/parts/odoo/addons/web/controllers/main.py", line 936, in _call_kw
    return getattr(request.registry.get(model), method)(request.cr, request.uid, _args, *_kwargs)
  File "/home/david/.voodoo/shared/shared_odoo/odoo8/openerp/api.py", line 250, in wrapper
    return old_api(self, _args, *_kwargs)
  File "/workspace/parts/project-service-fix-context/project_issue_task/project_issue.py", line 44, in action_create_task
    rec.write({'task_id': task_id}, context=context)
  File "/home/david/.voodoo/shared/shared_odoo/odoo8/openerp/api.py", line 248, in wrapper
    return new_api(self, _args, *_kwargs)
TypeError: write() got an unexpected keyword argument 'context'

When you deal with a browse, here `rec` context should not be set like args (cr, uid) even in old api

pedrobaeza:
You have to check because the problem is in other part, not here (maybe in Odoo itself). In the inheritance chain, one method is declared without `context=None` argument, only `context`.

PR #123, opened 2015-11-03
[FIX] Make project_functional_block installable

tungtt:
Fix inheritance of search view of project task.


dreispt:
Thank you @tungtt , TravisCI testes are not running properly yet on 9.0 unfortunately.
But in the meanwhile, could you please also fix the style checks?

tungtt:
My purpose was just to make it installable. However, results produced by TravisCI look cool so I will fix the styles.

@dreispt how can I assign this pull request to me so that when everything is OK, I will reassign it to you?

And sorry for this messy pull request (this is my first pull request in OCA).

dreispt:
Nothing messy, it's all good.
No need for the reassignments. Just comment here when it's ready for reviewers.

dreispt:
@tungtt OK, 9.0 tests are now fixed.
However we are asking for v9 modules to be in the new API, that's one of the reason why pylint checks are [failing](https://travis-ci.org/OCA/project/jobs/89060660#L258).
Could you please convert to the new API and check the remaining style errors?

dreispt:
PS: please post here if you need help or advice.

tungtt:
@dreispt: It is ready to review, finally. This is a summary of my changes
- Convert to new API.
- Update directory structure to make the code cleaner. Each file contains only one model.
- Remove field `sequence` as I think we don't need it. Ordering functional blocks by complete names is enough.
- Add smart buttons to display number of child blocks and number of tasks.
- Refactoring the implementation of field `complete_name` to make it faster.
  - Functional blocks data are kind of stable so I think it is not worthy to do the computation of `complete_name` every time we read a functional block. Therefore, I removed the implementation in `name_get`.
  - I make it a normal field and handle the value update within `create` and `write` functions. I tried to make it a functional field but failed with the recursion update.
    - Tried to depend on `parent_id.complete_name` but failed because the update of functional field does not go through `write` function.
    - Tried to put a function in `@api.multi(_get_related_fblocks)` to modify the value of `self` passed in the compute function but failed. The code looks like this

```
@api.multi
def _get_related_fblocks(self):
    # code here, returns browse records

@api.multi(_get_related_fblocks)
@api.depends('name', 'parent_id.name')
def _compute_complete_name(self):
    # code here

complete_name = fields.Char(compute='_compute_complete_name', store=True, string='Complete Name', index=True, size=512)
```

```
- Tried to do this but failed
```

```
@api.multi
@api.depends('name', 'parent_id.name')
def _compute_complete_name(self):
    # search for fblocks which are in self and children of self
    # update complete names of all above records

complete_name = fields.Char(compute='_compute_complete_name', store=True, string='Complete Name', index=True, size=512)
```

However, current implementation of mine leads to an issue that `complete_name` is not translated when `name` is translated. Not sure how to fix this properly.

Regards,

P.S.: what is the keyword to replace `description` in `__openerp__.py` ?

dreispt:
> P.S.: what is the keyword to replace description in **openerp**.py ?

The README.rst file replaces the description key.

jgrandguillaume:
Hi @tungtt 

Thanks for the good work and congrats for your first PR in the OCA world. One last thing : you can take this model here for the Readme.rst https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

More generally, you can find an OCA official template module here (useful to get some best practice): https://github.com/OCA/maintainer-tools/tree/master/template/module

I'm giving you my  :+1:  once the Readme.rst is ok

Regards,

Joël

project_functional_block/models/project_functional_block.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @api.multi
+    def write(self, vals):
+        res = super(ProjectFunctionalBlock, self).write(vals)
+        # Update complete_name for these records and child records
+        if 'updating_complete_name' not in self._context:
+            ctx = self._context.copy()
+            ctx.update({'updating_complete_name': True})
+            self.with_context(ctx)._update_complete_name()
+        return res
```
dreispt: I feel that the logic in create() and writ()e could be replaced by a converting complete_name into a computed field.
But I won't pick on that, just add it to the Known Issues/Roadmap section of the README please.

tungtt:
Done

dreispt:
@tungtt You still need to address Joël's comment above to "collect" his +1.

tungtt:
@jgrandguillaume, @dreispt `README.rst` is updated.

I also correct W503 warnings. However, I see TravisCI says this

```
======== Testing test_pylint ========
From https://github.com/OCA/project
 * [new branch]      9.0        -> origin/9.0
warning: refname 'origin/9.0' is ambiguous.
Start lint check just in modules changed
************* Module project_functional_block.__openerp__
project_functional_block/__openerp__.py:22: [C8102(manifest-required-key), ] Missing required key "images" in manifest file
```

What is the key `images` for ?

project_functional_block/__openerp__.py
```diff
@@ -19,25 +19,28 @@
...
-     'project_view.xml',
- ],
- 'demo': [
-     'project_demo.xml',
- ],
- 'installable': False,
- 'application': True,
+{
+    'name': 'Project requiring functional blocks',
+    'version': '1.2',
```
dreispt: Set to `9.1.1.2.0`

project_functional_block/models/project_functional_block.py
```diff
@@ -0,0 +1,152 @@
...
+
+class ProjectFunctionalBlock(models.Model):
+    _name = 'project.functional.block'
+    _description = 'Functional block to organize projects tasks'
+    _order = 'complete_name'
+    _rec_name = 'complete_name'
+
+    # Columns
+    name = fields.Char('Name', size=64, required=True, translate=True)
+    code = fields.Char('Code', size=8)
```
dreispt, line 36: The current best practice is to remove `size` from Char fields

project_functional_block/models/project_functional_block.py
```diff
@@ -0,0 +1,152 @@
...
+    # TODO: this field doesn't work with multi-languages
+    complete_name = fields.Char(string="Complete Name", index=True)
```
dreispt: Add a TODO comment: to convert into a computed field

project_functional_block/views/project_functional_block_view.xml
```diff
@@ -0,0 +1,83 @@
...
+            <field name="name">project.functional.block.search</field>
+            <field name="model">project.functional.block</field>
+            <field name="type">search</field>
+            <field name="arch" type="xml">
+                <search string="Functional Blocks">
+                    <field name="name" />
+                    <field name="complete_name" />
+                    <field name="parent_id" />
+                    <group expand="0" string="Group By...">
+                        <filter string="Parent" icon="terp-partner"
```
dreispt: Here `icon` is deprecated, and I believe that `expand`also.

project_functional_block/views/project_functional_block_view.xml
```diff
@@ -0,0 +1,83 @@
...
+                            <button class="oe_stat_button" name="action_view_tasks"
+                                type="object" icon="fa-tasks">
+                                <field string="Tasks" name="tasks_count" widget="statinfo" />
+                            </button>
+                        </div>
+                        <div class="oe_title">
+                            <label class="oe_edit_only" for="name" string="Name"/>
+                            <h1><field name="name" placeholder="Name"/></h1>
+                        </div>
+                        <group col="4">
```
dreispt: `col`is deprecated

project_functional_block/views/project_task_view.xml
```diff
@@ -0,0 +1,48 @@
+<openerp>
+    <data>
+        <!-- FUNCTIONAL BLOCK TO TASK -->
+        <record id="view_task_functional_block_form" model="ir.ui.view">
+            <field name="name">project.task.functional_block.form</field>
+            <field name="model">project.task</field>
+            <field name="type">form</field>
```
dreispt: Views `type`attribute is deprecated - remove form other views also.

project_functional_block/views/project_task_view.xml
```diff
@@ -0,0 +1,48 @@
...
+            <field name="inherit_id" ref="project.view_task_search_form"/>
+            <field name="arch" type="xml">
+                <xpath expr="//field[@name='name']" position="after">
+                    <field name="functional_block_id" select="1"/>
+                </xpath>
+                <xpath expr="//filter[@name='project']" position="after">
+                	<separator orientation="vertical"/>
+                    <filter string="Functional Block"
+                        name="groupby_functional_block_id"
+                        icon="terp-gtk-jump-to-ltr"
```
dreispt: remove icon

dreispt:
Well done @tungtt.
Some minor fixes a it's ready to go.
PS: if you can make `complete_name` a computed field that would be great!

dreispt:
@tungtt Ping!

project_functional_block/views/project_functional_block_view.xml
tungtt: done

project_functional_block/views/project_functional_block_view.xml
tungtt: done. But I find it inconvenient because I need to add extra `group`s.

project_functional_block/views/project_task_view.xml
tungtt: done

project_functional_block/views/project_task_view.xml
tungtt: done

project_functional_block/models/project_functional_block.py
tungtt, line 36: I am not sure that it is the best practice. For me, and also in my company, the best practice is to specifying explicitly the size of each `Char` field so that
- We have a clear understanding how much characters we can input in a field
- We can reduce the size of the database.

I will wait for your feedback to do this modification.

project_functional_block/__openerp__.py
tungtt: done.

project_functional_block/__openerp__.py
```diff
@@ -19,25 +19,28 @@
...
-     'project_view.xml',
- ],
- 'demo': [
-     'project_demo.xml',
- ],
- 'installable': False,
- 'application': True,
+{
+    'name': 'Project requiring functional blocks',
+    'version': '9.1.1.2.0',
```
dreispt, line 24: Version should begin with the Odoo target version,  9.0

project_functional_block/models/project_functional_block.py
```diff
@@ -0,0 +1,146 @@
...
+class ProjectFunctionalBlock(models.Model):
+    _name = 'project.functional.block'
+    _description = 'Functional block to organize projects tasks'
+    _order = 'parent_id desc, name'
+    _rec_name = 'complete_name'
+    _parent_store = True
+    _parent_order = 'name'
+
+    # Columns
+    name = fields.Char('Name', size=64, required=True, translate=True)
```
dreispt, line 35: Remove `size`from all Char fields

project_functional_block/models/project_functional_block.py
```diff
@@ -0,0 +1,146 @@
...
+        'project.functional.block', 'Parent Block',
+        ondelete='cascade', index=True)
+    parent_left = fields.Integer('Left Parent', index=True)
+    parent_right = fields.Integer('Right Parent', index=True)
+    tasks_count = fields.Integer(string='Tasks Count', compute='_count_tasks')
+    child_fblocks_count = fields.Integer(
+        string='Child Blocks Count', compute='_count_child_fblocks')
+
+    @api.multi
+    def _compute_complete_name(self):
```
dreispt, line 50: I think this can be implemented with a single line:

```
fir fblock in self:
    fblock.complete_name = '%s / %s' % (fblock.parent_id.complete_name, fblock.name)
```

project_functional_block/models/project_functional_block.py
```diff
@@ -0,0 +1,146 @@
...
+        for func_block in todo_fblocks:
+            complete_name = func_block.name
+            parent = func_block.parent_id
+            while parent:
+                complete_name = parent.name + ' / ' + complete_name
+                parent = parent.parent_id
+            func_block.complete_name = complete_name
+
+    @api.multi
+    def _count_tasks(self):
```
dreispt: If you add a `child_ids` you can then:

    for fblock in self:
        fblock.tasks_count = sum(x.tasks_count fro x in fblock.child_ids)

project_functional_block/models/project_functional_block.py
```diff
@@ -0,0 +1,146 @@
...
+        return {
+            'name': action.name,
+            'help': action.help,
+            'type': action.type,
+            'view_type': action.view_type,
+            'view_mode': action.view_mode,
+            'target': action.target,
+            'context': str(ret_ctx),
+            'res_model': action.res_model,
+            'domain': "[('id', 'in', (" + task_ids_str + ",))]"
```
dreispt, line 98: I don't think that domain and context must be strings.
AFAIK Python list and dict can be used directly..

project_functional_block/models/project_functional_block.py
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+    @api.multi
+    def _count_child_fblocks(self):
```
dreispt, line 102: If you add a `child_ids` you can then:

```
for fblock in self:
    fblock.child_fblocks_count = sum(x.child_fblocks_count fro x in fblock.child_ids)
```

dreispt:
I repeated the review. A few fixes and improvement suggestions made.

PR #127, opened 2015-11-13
[ADD] gap_analysis

xie8899:


gap_analysis/README.rst
```diff
@@ -0,0 +1,79 @@
...
+* ...
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/
+{project_repo}/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback `here <https://github.com/OCA/
+{project_repo}/issues/new?body=module:%20
```
elicoidal: Modify accordingly to include the repo/module name

gap_analysis/README.rst
```diff
@@ -0,0 +1,79 @@
...
+{gap_analysis}%0Aversion:%20
+{version}%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+
+
+Credits
```
elicoidal: Add button "Try me on runbot"

gap_analysis/README.rst
```diff
@@ -0,0 +1,79 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==============
+gap_analysis
+==============
+
+basic models on the business analysis and time estimation.
```
elicoidal: This module is part of a set ("Gap analysis") and  provides the 
basic models for business analysis and time estimation.

gap_analysis/README.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Installation
+============
+
+To install this module, you need to:
+
```
elicoidal: No specific steps required

gap_analysis/README.rst
```diff
@@ -0,0 +1,79 @@
...
+
+
+Configuration
+=============
+
+To configure this module, you need to:
+
```
elicoidal: No configuration steps required

gap_analysis/__openerp__.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
```
elicoidal: Please simplify your headers following OCA templates
https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3

gap_analysis/README.rst
```diff
@@ -0,0 +1,79 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==============
+gap_analysis
+==============
```
elicoidal: remove additional ==

gap_analysis/__openerp__.py
```diff
@@ -0,0 +1,41 @@
...
+{
+    'name': 'Gap Analysis',
+    'category': 'project',
+    'summary': 'Gap Analysis',
+    'version': '8.0.0.0.2',
+    'website': 'www.elico-corp.com',
+    "author": "<AUTHOR(S)>, Odoo Community Association (OCA)",
+    'depends': [
+        'base',
+        'web_ckeditor4',
```
rafaelbn: web_ckeditor4 has a bug, check https://github.com/OCA/web/issues/151

elicoidal: @rafaelbn Thanks to the pointer. We might switch to standard odoo editor or etherpad (maybe as a separate module).

gap_analysis/models/business.py
```diff
@@ -0,0 +1,212 @@
...
+        'Customer Story',
+        readonly=True,
+        states={'draft': [('readonly', False)]}
+    )
+    odoo_scenario = fields.Text(
+        'Odoo Scenario',
+        readonly=True,
+        states={'draft': [('readonly', False)]}
+    )
+    odoo_gap = fields.Text(
```
dreispt: I dislike the `odoo` specific field names. Can you use just `scenario` and `gap`?

elicoidal: Good point: it will be more generic.

gap_analysis/models/business.py
```diff
@@ -0,0 +1,212 @@
...
+        string='Total Estimated Time',
+        store=True,
+    )
+    state = fields.Selection(
+        selection="_get_states",
+        string='State',
+        default='draft',
+        readonly=True,
+        states={'draft': [('readonly', False)]}
+    )
```
dreispt: I like Stages as a configurable wrapper on top of static States. Maybe you can add that to the Roadmap section.

gap_analysis/models/business.py
```diff
@@ -0,0 +1,212 @@
...
+
+    business_requirement_ids = fields.One2many(
+        comodel_name='business.requirement',
+        inverse_name='parent_id',
+        string='Rough Estimation Lines',
```
dreispt: This label in incorrect I believe.

gap_analysis/models/business.py
```diff
@@ -0,0 +1,212 @@
...
+        ondelete='set null',
+        domain="[('id', '!=', id)]"
+    )
+    level = fields.Integer(
+        compute='_get_level',
+        string='Level',
+        store=True
+    )
+    summary_estimation = fields.Boolean(
+        string='Summary Estimation?',
```
dreispt: This one is worth a `help` explanation.

elicoidal: True: I will review the business case here.

gap_analysis/models/business.py
```diff
@@ -0,0 +1,212 @@
...
+        readonly=True,
+        states={'draft': [('readonly', False)]}
+    )
+
+    @api.one
+    @api.depends(
+        'rough_estimation_lines.estimated_time')
+    def _get_estimated_time_total(self):
+        time_total = 0
+        for line in self.rough_estimation_lines:
```
elicoidal: s/rough/draft (everywhere)

gap_analysis/security/security.xml
```diff
@@ -0,0 +1,16 @@
...
+<openerp>
+    <data noupdate="0">
+            <record id="group_gap_analysis_user" model="res.groups">
+                <field name="name">Gap Analysis User</field>
+                <field name="implied_ids" eval="[(4, ref('base.group_user'))]"/>
+                <field name="category_id" ref="base.module_category_project_management"/>
+            </record>
+
+            <record id="group_gap_analysis_manger" model="res.groups">
+                <field name="name">Gap Analysis Manger</field>
```
elicoidal: s/Manger/Manager

gap_analysis/models/business.py
```diff
@@ -0,0 +1,212 @@
...
+            time_total += line.estimated_time
+        self.estimated_time_total = time_total
```
dreispt: Make my day ans use a on liner:

```
time_total = sum(line.estimated_time for line in self.rough_estimation_lines)
```

gap_analysis/models/business.py
```diff
@@ -0,0 +1,212 @@
...
+        for br in self:
+            level = br._compute_level()
+            br.level = level
+
+    @api.multi
+    def _compute_level(self):
+        level = 1
+        if self.parent_id:
+            level += self.parent_id._compute_level()
+        return level
```
dreispt: SInce `_compute_level` is just a helper for `_get_level` I would prefer for it to be nested insode `_get_level`.
It would make the API cleaner.

gap_analysis/models/business.py
```diff
@@ -0,0 +1,212 @@
...
+        self.write({'state': 'done'})
+
+    @api.one
+    def action_button_cancel(self):
+        self.write({'state': 'cancel'})
+
+
+class BusinessRequirementCategory(models.Model):
+    _name = "business.requirement.category"
+    _description = "Bus. Req. Category"
```
dreispt: Is the abbreviation really needed?

gap_analysis/reports/business_report.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+        
+    </data>
+</openerp>
```
dreispt: Remove this file?

gap_analysis/views/business_view.xml
```diff
@@ -0,0 +1,208 @@
...
+                        <button name="action_button_done" states="approved" string="Done" type="object" groups="gap_analysis.group_gap_analysis_manger" class="oe_highlight"/>
+                        <button name="action_button_cancel" string="Cancel" type="object" attrs="{'invisible': [('state', 'in', ('done','cancel'))]}" groups="gap_analysis.group_gap_analysis_user" class="oe_highlight"/>
+                        <field name="state" widget="statusbar" statusbar_visible="draft,confirmed,approved,done,cancel"/>
+                    </header>
+                    <sheet>
+                        <h1>
+                            <field name="name" class="oe_inline" readonly="1"/>
+                        </h1>
+                        <group>
+                            <group cols="2">
```
dreispt: cols is deprecated

gap_analysis/views/business_view.xml
```diff
@@ -0,0 +1,208 @@
...
+                        <notebook>
+                            <page string="Business Analysis">
+                                <group string="Customer Story">
+                                     <field name="business_requirement" widget="text_ckeditor4" nolabel="1"/>
+                                </group>
+                                <group string="Odoo Scenario">
+                                    <field name="odoo_scenario" widget="text_ckeditor4" nolabel="1"/>
+                                </group>
+                                <group string="Odoo Gap">
+                                     <field name="odoo_gap" widget="text_ckeditor4" nolabel="1"/>
```
dreispt: As mentioned before, I would drop the "Odoo" work, to make it more generic.

gap_analysis/views/business_view.xml
```diff
@@ -0,0 +1,208 @@
...
+            </field>
+        </record>
+
+        <record id="view_business_estimation_type_form" model="ir.ui.view">
+            <field name="name">business.estimation.type.form</field>
+            <field name="model">business.estimation.type</field>
+            <field name="type">form</field>
+            <field name="priority">20</field>
+            <field name="arch" type="xml">
+                <form string="Bus. Estimation Type">
```
dreispt: Don't abbreviate

gap_analysis/views/business_view.xml
```diff
@@ -0,0 +1,208 @@
...
+            </field>
+        </record>
+
+        <record id="view_business_estimation_type_tree" model="ir.ui.view">
+            <field name="name">business.estimation.type.form</field>
+            <field name="model">business.estimation.type</field>
+            <field name="type">tree</field>
+            <field name="priority">20</field>
+            <field name="arch" type="xml">
+                <tree string="Bus. Estimation Type">
```
dreispt: Don't abbreviate

gap_analysis/README.rst
```diff
@@ -0,0 +1,75 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==========
+gap_analysis
```
dreispt: Use the human readable title

gap_analysis/README.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/140/8.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
+.. branch is "8.0" for example
```
dreispt: remove these two lines

gap_analysis/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# © <YEAR(2015)>
+# <AUTHOR(Elico Corp, contributor: Eric Caudal, Alex Duan, Xie XiaoPeng)>
```
dreispt: Is there a reason for the `<...>` format used?

gap_analysis/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+# © <YEAR(2015)>
+# <AUTHOR(Elico Corp, contributor: Eric Caudal, Alex Duan, Xie XiaoPeng)>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Gap Analysis',
+    'category': 'project',
+    'summary': 'Gap Analysis',
+    'version': '8.0.0.0.2',
+    'website': 'www.elico-corp.com',
+    "author": "<AUTHOR(S)>, Odoo Community Association (OCA)",
```
dreispt: Fill in the authors

gap_analysis/__openerp__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# © <YEAR(2015)>
+# <AUTHOR(Elico Corp, contributor: Eric Caudal, Alex Duan, Xie XiaoPeng)>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Gap Analysis',
+    'category': 'project',
```
dreispt: I'm not sure about this: will open discussion below.

gap_analysis/__openerp__.py
dreispt: I think this module should stick to the standard HTML fields/widgets, and then provide ckeditor as an option, through an additional module.

elicoidal: agree

gap_analysis/models/business.py
elicoidal: only at menu level (initial menu were not fully visible)

dreispt:
Great contribution, thanks!
Now, on the big picture: modules (and models) should use a prefix for their main "functional area"/"app".

So, this is not exactly a "project" module, so it does not have to use the `project` prefix.
But "Gap Analysis" as a functional area sounds restrictive to me. I would prefer something like Business Requirements Management. I also understand that this would be a big change for little benefit (other than neatness), so let me know what you think.

Whatever the choice, all models should use that prefix: we shouldn't have some with `gap_analysis` and others with `business`.

At this point I would suggest to adopt a "Business Requirements Management" module category and use the `business` prefix for the gap analysis models. The module should then be names `business_gap_analysis` and we could later add a `business_requirements` App top module.

What do you think?

elicoidal:
@dreispt we had some discussions on the contributor list about  the repo and decided (more because of laziness) to contribute it by default in project repo. We thought initially to have project_implementation (or similar).
I agree with you that Business Requirement Management category: what kind of other module could it contain though?
We are at the beginning of the development process so it is easy to change now for neatness as long as we do not go back and forth later. I will send a mail to the interested users to get their opinion as well.
After the decision is made we will modify all the modules.
BTW: any recommendation to get travis and runbot running knowing that these are sets of modules?

dreispt:
Funny, I don't recall seeing that discussion on the ML. But IMO this is the right repo for that feature, but I think the modules should use a prefix other than `project`.

I can see this link together the lifecycle for an Output/System/Project - estimation, sale, implementation, maintenance. For example: you could later identify the "system feature" a support Issue relates to.

Now, in project-speak this is in fact called a Product Breakdown Structure (PBS). I'll check Jordi Ballester's work to see what he has on this...

> BTW: any recommendation to get travis and runbot running knowing that these are sets of modules?

I'm not following. What would be the point?

elicoidal:
@dreispt 

> I can see this link together the lifecycle for an Output/System/Project - estimation, sale, implementation, maintenance. For example: you could later identify the "system feature" a support Issue relates to.

Agree. This is why we thought of the project_implementation. The BR are actually linked later to the tasks (and when ready in the reports) to allow strict follow up between planned and actual hours for project control.

>    BTW: any recommendation to get travis and runbot running knowing that these are sets of modules?
> I'm not following. What would be the point?
> Travis and runbot are failing because the PR of the set are depending on each other. Anyway to fix this?

gap_analysis/reports/business_report.xml
xie8899: I will add the reports here in the future.

gap_analysis/__init__.py
xie8899: The right format is like this? <\Elico Corp, contributor: Eric Caudal, Alex Duan, Xie XiaoPeng(S)>\

gap_analysis/models/business.py
xie8899: I will do it.

gap_analysis/reports/business_report.xml
dreispt: No need for an empty file nevertheless.
I could add that note to the Roadmap section.

dreispt:
> Now, in project-speak this is in fact called a Product Breakdown Structure (PBS). I'll check Jordi Ballester's work to see what he has on this...

I think Requirements/gap Analysis is different from both a PBS (Product Breakdown Structure) or a WBS (Work Breakdown Structure). They can be mapped at the project planning stage, but are different things. So no collision with [Jordi's work](https://github.com/Eficent/eficent-odoo-addons) at the moment.

So, what prefix to use for the Module and for the Models?
I like the `business` prefix you are already using. It points us to "Business Analysis" which includes requirements analysis but is much wider. We can also use that as a new module Category.

So I propose to rename the module to `business_gap_analysis` and keep all the Model names as they are.

@elicoidal @xie8899  ?

elicoidal:
Agree

JordiBForgeFlow:
@elicoidal @dreispt I like very much your approach as a tool to manage a repository of requirements. That would constitute the PDB that @dreispt is talking about. So, describe the future state of a product in terms of requirements.  The approval process is good and much needed.

I can see, that the current module can introduce conflicts with our approach to project management following ISO 21500, because here you are introducing estimates.

Virtually any project manages estimates, so the solution to enter planned time & materials to complete a deliverable or work package should be common for all projects, wouldn't you agree?

An interesting idea could be that you can enter estimates into the requirements, but these estimates are also liked/incorporated to the project's (WBS) estimates, so that you can analyze the planned costs from the project (and compare plan vs. actuals if needed).

So I can see that there's an inmediate need for a centralized project tool to enter estimates (material/time) for a project, that should not be restricted to the Business Analysis.

The module "analytic_resouce_plan" provides precisely this tool. We have moved the project repository to 
https://github.com/projectexpert/pmis, and it is available for v8 (old API, unfortunately).

So, it would be great if the requirements can link to a project (WBS element), and have the possibility to enter the estimates as analytic resource plan lines. That would better integrate Business Requirements with Projects.

I think it is time for us to start to move our Project repo to OCA @sysadminmatmoz and me are very involved in this now.

dreispt:
@jbeficent I see them as complementary. In a Project you can surely use a WBS supported in an Analytic Account tree, and have there budget and estimated times.

But at the Gap Analysis stage we are way before that. We could be just responding to an RFQ. We need to enumerate the customer requirements, ipsis verbis, to then evaluate gaps (an extension could link this to a PBS) and estimate costs. No point in using Analytic Accounts at that stage.
Mapping the gap analysis a WBS could happen later, when (and if) the Project is initiated.

I can also see a future "Business Analysis" possible use case where you keep a system installation components and features map using some sort of PBS. It would support evolution maintenance, where small projects have a small requirement set and deliver additional features to an extiting PBS (installed system).

JordiBForgeFlow:
@dreispt When you estimate costs based on your analysis you are in fact providing estimates for a future (possible) project. The results of the Gap Analysis results generally into an Approved Budget for the project, which constitutes itself the starting baseline of costs for the project.

So, say you start a gap analysis for a project "Odoo implementation at Apple Inc.".
WBS:
- Project [P / A001]: "Odoo implementation at Apple Inc."
  -- Phase [P / A001 /1]: Gap Analysis

Then you create a requirements repository with the business requirements, and the solution requirements (gaps). You link the requirements to [P / A001](the requirements are related to the project implementation, the gap analysis represents the the work you do to elicit the requirements of the overall project).

Then you enter the estimates for each gap requirement using Analytic Plan Lines (better than resource plan lines, as this is only going to be used to assess the cost). As the Requirement is linked to the project [P / A001], so are the Analytic Plan Lines that you enter in the Requirement.

You now have the cost of each gap, represented as analytic plan lines.

And in project [P / A001] you have also the estimated cost for the project implementation, based on the results of the gap analysis.

The Analytic Plan module (https://github.com/projectexpert/pmis/tree/8.0/analytic_plan) supports planning versions. In this case the Requirements would use a planning version "Gap analysis", so that you can create other planning versions in the future (for example, once the project is won), and compare costs between versions.

JordiBForgeFlow:
@dreispt It important to highlight that Business Analysis is always part of a project. Not creating a project during the business analysis would be an error IMHO. The project does not need to be fully decomposed in the WBS. But just to the level that it can support the business analysis process.

elicoidal:
@jbeficent thanks a lot for your highly valuable input!

> An interesting idea could be that you can enter estimates into the requirements, but these estimates are also liked/incorporated to the project's (WBS) estimates, so that you can analyze the planned costs from the project (and compare plan vs. actuals if needed).

This is interesting and I agree with the general idea. Actually this is somehow the plan (be aware that there are another 5 PR associated): estimates are necessary during the negotiation phase but I think should not be allowed when creating the tasks from BR and replaced by accurate estimation (constraint to be added in the wizard to create the tasks).

> The module "analytic_resouce_plan" provides precisely this tool. We have moved the project repository to https://github.com/projectexpert/pmis, and it is available for v8 (old API, unfortunately).

I will have a look but I think the AA should only be introduced when the project is validated and tasks associated to the project created. This is when I think convenient to create some AA entry.
My understanding of cost followup of the project was much straight forward: create a BI view joining BR/Tasks/Timesheets with 4 dimensions: BR time vs TS time and BR value vs TS cost (to get the gross profit).

> So, it would be great if the requirements can link to a project (WBS element), and have the possibility to enter the estimates as analytic resource plan lines. That would better integrate Business Requirements with Projects.

For the me the estimate is just a intermediary state until we finalize the negotiation. At that moment we get some kind of final figure that could be used as Standard to evaluate the project (for invoicing or cost control). Even if you are in times and material the final value will be the one to dictate the negotiation with the customer ("you said 100 and you spent 150")

> Then you enter the estimates for each gap requirement using Analytic Plan Lines (better than resource plan lines, as this is only going to be used to assess the cost). As the Requirement is linked to the project [P / A001], so are the Analytic Plan Lines that you enter in the Requirement.

This is nice indeed but my feeling is that it is overkill for many of the small prospects or projects. I think your way is the way to go for large projects which was not actually my original objective.
I am not against the idea because I see the beauty of it but also the complexity associated which might not suit small businesses (my original primary target).

> It important to highlight that Business Analysis is always part of a project. Not creating a project during the business analysis would be an error IMHO. The project does not need to be fully decomposed in the WBS. But just to the level that it can support the business analysis process.

I agree. Actually project is created since the beginning and linked to the Opportunity but not actively used until the end of the process (see below)

@dreispt Thanks a lot as well for your input! 

> But at the Gap Analysis stage we are way before that. We could be just responding to an RFQ. We need to enumerate the customer requirements, ipsis verbis, to then evaluate gaps (an extension could link this to a PBS) and estimate costs. No point in using Analytic Accounts at that stage.
> Mapping the gap analysis a WBS could happen later, when (and if) the Project is initiated.

Fully agree, at least in most of our project implementations.

Both points of view are in my opinion complementary and not far from my original idea
Actually my vision is to integrate/funnel the sales down to the execution:
- the opportunity ("Original customer interest/lead", start of the project process) 
- the Business requirement ("Rough or detailed expression of the need")
- the estimation ("Gap analysis")
  (proposed way with cost structure is one possible way, another being adding standard objects/estimation for the BR)
- the creation of the SQ with or without estimate ("Customer negotiation")
- the validation process by management ("Internal cost/ sales approval")
- the validation of the SO ("Closing the contract")
- the creation of all related tasks in the project ("Execution") and timesheets.
- Make sure the budget is respected ("Reporting and control")

That said, we can think about a new module (which gives another option to the current set) to link the AA (through the modules proposed) and create the tasks in a project as during the gap (the project is somehow "in process"/"building") which could be complementary to the current wizard which creates the tasks at the end.

Another important point is the current scope of this modules was very small or small ERP/service implementation project (from 5K euro to 300K eur). 
This is narrow as you can see but my idea as well was to create an object base that could be used for other type of projects (Engineering or construction for example), adding new bricks. So if we need to be more generic and allow other additional modules to be plugged in, I am adaptable.

Again thanks for the valuable input

elicoidal:
An additional point from my last comment: current estimate is task creation oriented (with relevant information: user/ resource/ time), which I think is too restrictive.
We should allow to chose between:
- task,
- service or
- stockable product, 

with corresponding necessary fields. I think it would be quite generic for most cases.
eg a project to build a website would require:
- 1 server to host the website (stockable)
- 1 task to design the website (task)
- 1 template etc. (service)

The information would be pushed via the wizard to the SQ for the negotiation, and validated SO for final approval and in fine:
1. The stockable requirement type would trigger a PO or tender (through MTO/MTS)
2. The task type would create a task in the project (through wizard). (current covered case)
3. The service type would trigger an invoicing after validation of the SO/invoicing 
NB: actually knowing that in v8 service products are not triggering invoicing in some cases, maybe to be finetuned

sysadminmatmoz:
@elicoidal , @jbeficent 
The first thing to coordinate is the project definition, since a lot of people mix it with contract and consider it created only after a RFQ was accepted; and that's wrong.

A project starts with the lead; the opportunity assessment is already in the project initiation stage - we are already spending time and resources by studying the lead and we're already executing, creating and delegating tasks. The RFQ creation process already needs at least some rough-cut resource plan, time plan and we already need the project scope definition and a WBS for at least the deliverables. Our product at this stage is the project charter, which can be considered as the offer (sale quotation + all other documentation needed to assess the offered job).

If the offer is accepted (and we get a sale order) we continue to the next stage: project planning, otherwise the project is cancelled; but it was there, we had costs and we spent time, even if we didn't sell a thing! The project plan handles the fine tuned work plan and control plan.

Only after that stage we enter in the operations stage (execution and control) after which the closure stage starts.
**Until there I think we're on the same page.**

The main approach difference is in handling the tasks. For a professional PM, a task is the lowest level of the project plan (worker A completes the job B), everything else is handled on the project level (as low in the WBS hierarchy as needed). Materials needed to do the job are handled through the resource plan (own timesheets, external services, purchases).

If I reuse the website building project, the structure is:
- parent project (class: project): Build a website
  - sub project 1 (class: deliverable): Website hosting (resources: server rental, manpower for setup, manpower for maintenance, ...)
  - sub project 2 (class: deliverable): Website design (resources: workstation assets, software assets, manpower assets, ...)
  - sub project 3 (class: deliverable): Website template (resources: workstation assets, software assets, manpower assets, ...)
    etc...

**The resource plan lines wizard generates the purchase requests, the billing plan lines generate the sale requests.**

The planning is done on the project level, tasks are just for the daily to-do lists.

JordiBForgeFlow:
@elicoidal I propose to separate the part of estimates to a separate module, so that one module handles purely the requirements management part, and the other handles the estimation. This would add the possibility for alternative estimation methods for requirements management.

elicoidal:
@jbeficent we can do that if it helps on getting more flexibility in the deployments of the modules.
Do you have already in mind something as alternate solution or is it just in case?

elicoidal:
@sysadminmatmoz I understand your PoV and it is valid for large projects.
I think we could give the option to generate either task or project during the final step.
to be noted that the time estimate can not be recorded at project level as is but a link to the BR can be done.
Would it be OK?

dreispt:
AFAICS we agree that there are several approaches that can be used once the SO is closed: Project Tasks, Analytic Account WBS, or even a Manufacturing BoM.
Those should be option to use later in the process, provided bu specific extension modules.

At this point this module just adds support for Business Requirements analysis.
They can later be mapped into a PBS and WBS, but out of scope here.

It also adds BR Estimation Lines. I think it's OK to do so.
These lines would later be copied to a Sale Order, so they should have a similar structure.
And in SO lines we can have stockable items and services. AFAIK there are already features to converting SO Service lines into Projects or Tasks - they could be reused and extended. 

So, for this PR I'm good with the design in general, just need to have a closer look at estimation lines.
Maybe the name should be just `business_requirement`?

sysadminmatmoz:
Although there are modules converting SO lines in tasks, our (project.expert) approach goes in the opposite direction; project planning lines are converted in SO lines; that's the main difference in the approach.

We start a project before we even started with the quotation.

Thus we'll have to find a way to connect the billing plan lines with the BR estimation lines, since it seems they're basically the same thing.

dreispt:
@sysadminmatmoz How do you represent the billing plan?

JordiBForgeFlow:
@dreispt  The billing plan is represented by billing plan lines. Each billing plan line refers to an analytic account, date, the product that will be used to bill and the quantity and amount. Basically represents the amount that will be billed to the customer when a certain milestone of the WBS element is achieved.  Billing plan lines can create sales orders.

So you see, the Billing Plan is more oriented towards fixed price projects, where there's a contractual agreement on the milestones and billed amounts.

I created the Billing Plan to resemble SAP's approach: https://help.sap.com/saphelp_46c/helpdata/en/4c/22745c46e611d189470000e829fbbd/content.htm

elicoidal:
Thanks @dreispt @sysadminmatmoz @jbeficent for your valuable input. 

I would like to keep the estimated lines in the BR as a container that we can use later to link to the Billing Plan lines.
Indeed I see 2 levels of usage here.
Simple one: 
1. we create the BR, 
2. Fill in estimation 
3. Create SO
4. Create final project tasks for the BR

Elaborated one:
1. Project is created
2. Billing plan is created/linked
3. we create the BR (and maybe sub-projects are created on the go), 
4. Fill in estimation (linked to BP lines)
5. Create SO
6. Create final project(s) for the BR

Again, the idea is not to get all features today but to be able to have the generic objects so that we can move later on without having to rebuild/migrate.

Your thoughts?

dreispt:
@elicoidal Both processes are valid, and can be used in the same organization in different cases.
Go with the simple implementation first.
Estimation lines should be part of the module, so that they can be a common link to be used by both approaches.

Now, what should this module be named?

JordiBForgeFlow:
@elicoidal I can see from other modules that you're making other areas (CRM, Sales, Financials) circumvent the Business Requirement. IMHO this approach is dangerous. Creating financial planning for BR will make BR take the role that project has. Thus, when some needs to handle financials for projects, link to sales, etc... will just need to create duplicate structures.

I see BR's as artifacts that are created around a project. They serve a project, and not the other way around. The project must be the cornerstone of it all.

dreispt:
I would really like to unblock this particular module please.

@elicoidal I think this work deserves it's own Roadmap, where we can discuss how the other pieces fit together. I suggest it to include a link to a Google Doc where we can better collaborate around each item.

JordiBForgeFlow:
@dreispt @elicoidal Agreed. My only request would be that the estimates are split from the Business Requirements I would appreciate it. I'm eager to adopt Business Requirements, but as as a standalone component. We'll use a different approach to estimating. Modularity for me is in this case important.

elicoidal:
@dreispt I am fine with business_gap_analysis or better business_requirement (as we said that gap analysis was maybe to restrictive).
@jbeficent We split the modules to allow modularity and generic behaviour.  I think we need to keep the estimate as a container (that you can fill in/super with your own structure).

> I can see from other modules that you're making other areas (CRM, Sales, Financials) circumvent the Business Requirement. IMHO this approach is dangerous. Creating financial planning for BR will make BR take the role that project has. Thus, when some needs to handle financials for projects, link to sales, etc... will just need to create duplicate structures.

That is the original idea: allow a full link between CRM=>BR=>project so that we can control the commercial activity down to the execution. This is actually the same idea followed by Odoo 9 (with SQ and milestones).
Now I can remove in separate module the financial estimate and other modules could be handled later.
I will need to move forward in this aspect for my business (which is not necessary compatible with large project management) and might simply publish the other set of modules in our company repo.

dreispt:
I concur with @elicoidal.
Business Requirements are just a wishlist from the customer.
Each BR can map to a PBS (an ERP feature) or to a WBS (a task, such as "train the users").
Whatever PBS and WBS approach you choose to use, you need some data structure to link them with BRs.

So I see that we will always need the BR Estimates model, and don't the disadvantage to having it bundled with BRs.
@jbeficent  @sysadminmatmoz ?

elicoidal:
@dreispt Google is not my friend in China... But we will have to make it go ;)

@xie8899 @duanyp1991 Can you separate the code of the estimation in new module, change the name of all the modules and apply the full modifications we have been discussing in this thread (CKEditor, etc.)

JordiBForgeFlow:
@dreispt WBS is a deliverable oriented decomposition of the work to be performed. Means that the PBS is in fact embedded in the upper levels of the WBS (Deliverables), while the lower levels (Work Packages) describe how to achieve the first ones.

Every project has estimates. Requirements Management is, in projects, the effort required to describe the outputs of the Project. So, the time you spend to describe precisely the Scope of the project, and the Planning activities to identify the Cost of the project.

The statement "It is too early to create a project" or "but we have not yet even sold the project", does not mean that BR analysis should not be part of a project. During BR Analysis you are identifying the overall outputs for the project and their cost.

You normally start with the Cost Plan that was identified during the initial Business Analysis (also referred to as Gap Analysis), and as the project progresses and more information arises the Cost Plan can be refined (or due to change requests incorporated).

Therefore, the Project Cost Plan is first elaborated as an output of the Business Analysis, and then modified with Change Orders during Project Execution.

elicoidal:
@jbeficent 

> Therefore, the Project Cost Plan is first elaborated as an output of the Business Analysis, and then modified with Change Orders during Project Execution.

Looks like the estimation lines to me: can you elaborate how they would differ?
Note that currently we placed the project at CRM level as I consider the CRM opportunity the top container of the structure (Project<>CRM => BR =>estimation=>SO=>task in project).
I was considering having a Business Requirement Analysis object to contain BR lines but I did not follow as I wanted it simple (and linked to Opport.).
We might reconsider as currently, link is done at opportunity level and I am not sure this always relevant.

dreispt:
@elicoidal whatever may work as a collaborative document; maybe an Etherpad would do.

JordiBForgeFlow:
On Mon, Nov 23, 2015 at 1:46 PM, Eric Caudal notifications@github.com
wrote:

> @jbeficent https://github.com/jbeficent
> 
> Therefore, the Project Cost Plan is first elaborated as an output of the
> Business Analysis, and then modified with Change Orders during Project
> Execution.
> 
> Looks like the estimation lines to me: can you elaborate how they would
> differ?
> 
> You are placing estimates assigned to BR's.  Considering
> Project->CRM->BR->estimation, it's pretty hard to say how much is the
> baselined cost for a project that had an Initial Baseline + 10 approved
> changed orders. IMHO Estimates are directly linked to a project, as cost
> planning is one of the key aspects of Project Management.
> 
> Note that currently we placed the project at CRM level as I consider the
> CRM opportunity the top container of the structure (Project<>CRM => BR
> =>estimation=>SO=>task in project).
> 
> That's too far to the project IMHO, as per what I indicated above.
> 
> I was considering having a Business Requirement Analysis object to contain
> BR lines but I did not follow as I wanted it simple (and linked to Opport.).
> 
> You did well not to create the BRA object. I find it useless as well. If
> only BR lines could be linked to the Project, that would make soo much
> sense to me.
> 
> We might reconsider as currently, link is done at opportunity level and I
> am not sure this always relevant.
> 
> IMHO first is the link to Project. Then others can follow as well. For
> instance, we have an entity called Change Order. That could well contain BR
> lines that this CR refers to, but I think that you always need to be clear
> on what is the project associated to this BR.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OCA/project/pull/127#issuecomment-158924827.

## 

Jordi Ballester Alomar
Founder | Eficent
(+34) 629530707 | jordi.ballester@eficent.com | http://www.eficent.com
Twitter: https://twitter.com/jbeficent_erp | Linkedin:
https://www.linkedin.com/in/jordiballesteralomar

dreispt:
I'm setting work in progress. Let us know when we can start the review on this PR.
The reviews on the other features should then follow - a roadmap Issue could guide us in this process.

elicoidal:
@jbeficent 

> You are placing estimates assigned to BR's. Considering Project->CRM->BR->estimation, it's pretty hard to say how much is the baselined cost for a project that had an Initial Baseline + 10 approved changed orders. IMHO Estimates are directly linked to a project, as cost planning is one of the key aspects of Project Management.

I do not see the difficulty: BR are linked to tasks when they are created so you get you valuation at project level. For change orders, see below. We plan to have a report to get est vs actual based on TS (see https://github.com/OCA/project/pull/127#issuecomment-158051834)

> IMHO first is the link to Project. Then others can follow as well. For instance, we have an entity called Change Order. That could well contain BR lines that this CR refers to, but I think that you always need to be clear on what is the project assohttps://github.com/OCA/project/pull/127#issuecomment-158051834ciated to this BR.

Any change order is a new commercial opportunity for which you add new opportunity and then create new BRs for this new opportunity. You quote to the customer and finally, when accepted, integrate to the same project as tasks. 
In the end:
- Commercially you have a follow up and possible quotation.
- The project contains BR from 2 opportunities, which is fine (or it could be 2 projects as well so this is flexible)
- All tasks are linked to the BR so if you need the initial value of the BR, this is easy.
- Change orders are managed properly
  Now, I currently proposed Project <> CRM => BR, I do not see any issue to switch to CRM <> Project => BR if that helps getting more standard/generic management

All in all, I do not think the proposed solution is far from what you need BUT I understand that this might be not standard management as per other software (cf: SAP) and I'd rather modify it to stick to some good practices if necessary.

@dreispt I think we need to decide on this last topic and we could be good to go. etherpad is a good option as well

sysadminmatmoz:
> I do not see the difficulty: BR are linked to tasks when they are created so you get you valuation at project level. For change orders, see below. We plan to have a report to get est vs actual based on TS

Project is an analytic account, task is just one of the journals. The cost planning is done on the account level, not on journal. Task is not a necessity for PM, specially on the planning level. At the initiation stage we do not plan tasks, just project... and it's not related only to big projects, it's just a common PM practice.

> Any change order is a new commercial opportunity for which you add new opportunity and then create new BRs for this new opportunity. You quote to the customer and finally, when accepted, integrate to the same project as tasks. 

A change request is not a commercial opportunity. A CR can be started by any project stakeholder (team member, supplier, authority, law, situation, ...) and not only by the customer. You shouldn't mix change management with customer relations management. The closest thing to the change management is the claim management (but not the same thing nevertheless). A CR can be just an idea to do things faster, a proposal to do things differently, a last minute change in plan, etc... and it is not directly tied to quotations although it could result in one sometimes, but rarely. It's a process undergoing while the Project is way in progress.

We're not trying to manage things like other software do, but we're trying to manage things like they are supposed to be managed in real life and on real projects. That's the real good practice (I spent my whole life managing projects, companies, groups of companies, ...).

From my point of view: BR should be linked to analytic planning lines and not to tasks. Cost planning and control should be done on project level and not on task, actually everything should be done on project level. A client request (a lead) starts a project - the only thing that's variable is, when that projects ends. It could be on the initiation level (after the quotation, which is not accepted), it could be on the planning level (the plan is not accepted), it could be on the execution level (contract is cancelled while works were already in progress) or it could be brought home through the whole closure level, but in any case, a project started and it existed, even if only for a while.

elicoidal:
@sysadminmatmoz Well noted and I think we have some kind of final understanding.
Basically instead of linking the BR to CRM we link it to project. 
We can still have the opportunity linking the project (and BR links) and if necessary generate a SQ based on it (separate module).
For change request, we should have field "Change request" (m2o) in the BR that could be used for traceability and later to print additional quotation on demand (Change request tab to be managed at Project level)

dreispt:
@elicoidal I don't think BRs should be hardlinked to Projects. I want to be able to use BRs without using Projects, or without using SOs, for that matter.
This PR should be agnostic to that.

elicoidal:
@dreipt Let's then stick for the moment to what we have but we might 
think about having one module creating the link later between BR and 
projects.

JordiBForgeFlow:
@dreispt @elicoidal I agree. The more flexibility, the better.

@elicoidal Some other iteas around the business analysis. Sometimes there's not only a parent-child relationships with regards to business requirements. Sometimes you need to link requirements, identifying what is the type of relationships (depends on, solves,..). For instance, the Business Analysis Body of Knowledge classifies requirements into the following categories: business requirements, stakeholder requirements, solution requirements. 

Example:
- Business requirement: I want to travel faster
- Stakeholder requirement: I want a faster horse
- Solution requirement: Build an automobile incorporating an engine that prompts wheels to be moved by the force generated by the engine.

Solution requirements (the gap) sometimes do not have a 1:1 relationship with the stakeholder requirement. The business analyst must capture the Business and stakeholder requirements, and then work closely with the project team to build the Solution requirements. Traceability between them all ensures that the solution that is build implements correctly the needs identified during the business analysis. 

While this imposes a more complex set-up, it would open up for a better adoption by organizations wishing to adopt the standards of Business Analysis into their organization using Odoo. After all alignment with existing standards IS a selling point. @rafaelbn  will probably agree on that.

elicoidal:
@jbeficent Do you mean to create 3 fields for "gap" or to have a m2o to allow greater flexibility (but less readibility of current form actually)

JordiBForgeFlow:
@elicoidal A solution requirement would be separate from the stakeholder requirement, but linked. As an example of a good way to link, see the module 'partner_relations'. https://github.com/OCA/partner-contact/tree/8.0/partner_relations. Building a module 'business_requirement_relations' would do that.

Then, the parent_id (that Odoo S.A. tends to move away from) would not be needed, because in fact it depicts a specific case of relationship between BR's.

dreispt:
I created a pad for this: http://pad.odoo.com/p/oca-project-br

elicoidal:
@jbeficent I understand: I would keep it for a separate module for the moment.
@dreispt thanks for the pad. Is there any example or guideline somewhere about how to fill it?
I will gather the thoughts from this page first + the original gap analysis functional description altogether in the pad.

JordiBForgeFlow:
Thanks @dreispt! the pad is very clear, and a very good starting point to define the future roadmap.

elicoidal:
@dreispt Indeed I didnot see that you have done all the job ;) many thanks

dreispt:
Just did some copy-pasting :smile:

elicoidal:
@jbeficent @dreispt @sysadminmatmoz 
I have been reading again all the comments and checked the 2 main additional projects that might be useful to complete this module 
- https://github.com/projectexpert/pmis/tree/8.0/analytic_plan and related and
- https://github.com/OCA/partner-contact/tree/8.0/partner_relations

I fully agree on your approach but I fear the the gap might take time to bridge so I suggest to get it done in several versions.
For the **Analytic plan**, I think we could actually get it done immediately having:
- the object project directly linked to the BR and
- the estimation lines pushed to another module (selective install)
- you could install the analytic plan module without conflict

For the **relationship** you suggested, I think it is mandatory but would imply a major refactoring I am not sure we are able to do right now so I prefer to keep a simple parent/child relationship with a roadmap to migration later.

Here is what I compiled as modifications to be done short term (could be achieved tomorrow):
- Module name: business_requirement and business_requirement_*
- module category:  Business Requirements Management
- Make sure all models starts with business_
- Remove estimation and estimation lines from BR objects
- Move all classes/methods linked to estimation into separate module
- BR not directly linked to CRM anymore, only to project (they might be still indirectly linked through selected Project in CRM).
- Add a tab of smart link BR in the project form.
- Add a checkbox “Change Request” at BR level (for traceability to be used later)

This basic module will have Project => BR
For estimation/deliverable module (simplistic alternative to Analytic plan)
- name: business_requirement_deliverable
- “Estimation” to be renamed “Deliverable”.
- In Estimation/Deliverable: user and task type are not mandatory (if present this is a potential task).
- If “user” in estimation/deliverable then a task is created from the deliverables list thanks to wizard
- All deliverables (could be task/service/stockable) can be potentially used in the SO as SOL

This additional module will add BR => Deliverables. 
Other modules might depends on it (CRM etc.)

Additional notes for other modules:
- Wizard Task (to be moved to the project) creates the tasks in the project based on the deliverables with users
- Wizard create SO (located in CRM) creates the SO with Tasks+stockable+Services from the BR of the linked project

Quick roadmap for business_requirement (check http://pad.odoo.com/p/oca-project-br):
1. Use improved relationships
(@jbeficent I might need your input)
2. Reporting
3. Fully integrate Analytic Plan with the rest of the modules (reports, SO, BI)

Questions:
1. Project link: actually we could create a sub project every time we create a BR and link it to the BR on one side and the parent project on the other side. It seems to me a little bit too specific though right now and can be anyway achieved by manually selecting the project in the BR.
2. In the deliverable/Analytic plan, conceptually are we talking about the cost of our resources for the integrator (as in COGS for integrator) or for the customer (as Revenue for the integrator)

Thanks again for all your input making clear this functional requirement
Let me know if I forgot or misinterpreted anything

JordiBForgeFlow:
On Thu, Nov 26, 2015 at 1:58 PM, Eric Caudal notifications@github.com
wrote:

> @jbeficent https://github.com/jbeficent @dreispt
> https://github.com/dreispt @sysadminmatmoz
> https://github.com/sysadminmatmoz
> I have been reading again all the comments and checked the 2 main
> additional projects that might be useful to complete this module
> - https://github.com/projectexpert/pmis/tree/8.0/analytic_plan and
>   related and
> - https://github.com/OCA/partner-contact/tree/8.0/partner_relations
> 
> I fully agree on your approach but I fear the the gap might take time to
> bridge so I suggest to get it done in several versions.
> For the _Analytic plan_, I think we could actually get it done
> immediately having:
> - the object project directly linked to the BR and
> - the estimation lines pushed to another module (selective install)
> - you could install the analytic plan module without conflict
> 
> For the _relationship_ you suggested, I think it is mandatory but would
> imply a major refactoring I am not sure we are able to do right now so I
> prefer to keep a simple parent/child relationship with a roadmap to
> migration later.
> 
> Here is what I compiled as modifications to be done short term (could be
> achieved tomorrow):
> - Module name: business_requirement and business_requirement_*
> - module category: Business Requirements Management
> - Make sure all models starts with business_
> - Remove estimation and estimation lines from BR objects
> - Move all classes/methods linked to estimation into separate module
> - BR not directly linked to CRM anymore, only to project (they might
>   be still indirectly linked through selected Project in CRM).
> - Add a tab of smart link BR in the project form.
> - Add a checkbox “Change Request” at BR level (for traceability to be
>   used later)
> 
> Not really necessary, but can be handy in a simplified version. We have an
> object "Change Request" (
> https://github.com/projectexpert/pmis/tree/8.0/change_management) where we
> can add BR's. Then, in the BR you would have the related Change Request.
> 
> This basic module will have Project => BR
> For estimation/deliverable module (simplistic alternative to Analytic plan)
> - name: business_requirement_deliverable
> - “Estimation” to be renamed “Deliverable”.
> - In Estimation/Deliverable: user and task type are not mandatory (if
>   present this is a potential task).
> - If “user” in estimation/deliverable then a task is created from the
>   deliverables list thanks to wizard
> - All deliverables (could be task/service/stockable) can be
>   potentially used in the SO as SOL
> 
> I see a conflict of terminologies. A deliverable is a description of an
> output resulting of some work (using resources). So I do not see a
> deliverable = estimation at all.
> I see the deliverable as the output. E.g. a product/service that is
> obtained as end result. That deliverable can then be used as the basis for
> the creation of the SO Line, or it can be the SO line itself.
> A deliverable can be achieved as the result of work / materials / services
> executed.

The key thing is how do you put a price to a deliverable based on the work
/ materials / services executed. SAP uses a concept called Dynamic Item
Processor (
http://help.sap-ag.de/saphelp_globext607_10/helpdata/en/d4/9b86687fee11d3a74400104b59c1ec/content.htm),
that will use the planned costs (materials, services,...) to determine the
end product and sale price.

We see deliverables being a separate entity (e.g. object "deliverable").
Project deliverables can be used as the basis for the creation of SO's.

I still see Estimates as the Resource Plan (
https://github.com/projectexpert/pmis/tree/8.0/analytic_resource_plan).
Analytic plan would be lower level, since it only deals with costs, and
cannot trigger procurements.

> This additional module will add BR => Deliverables.
> 
> Then I see as BR => Deliverables => Estimates
> Example:
> BR: I need a faster horse
> Deliverable: Car
> Estimates: 1 Engine, 4 Wheels,....
> 
> Other modules might depends on it (CRM etc.)
> 
> Additional notes for other modules:
> - Wizard Task (to be moved to the project) creates the tasks in the
>   project based on the deliverables with users
> - Wizard create SO (located in CRM) creates the SO with
>   Tasks+stockable+Services from the BR of the linked project
> 
> I see the wizard as Deliverable => SO Lines
> 
> Quick roadmap for business_requirement (check
> http://pad.odoo.com/p/oca-project-br):
> 1. Use improved relationships
> (@jbeficent https://github.com/jbeficent I might need your input)
> 2. Reporting
> 3. Fully integrate Analytic Plan with the rest of the modules (reports,
> SO, BI)
> 
> Questions:
> 1. Project link: actually we could create a sub project every time we
> create a BR and link it to the BR on one side and the parent project on the
> other side. It seems to me a little bit too specific though right now and
> can be anyway achieved by manually selecting the project in the BR.
> 
> I can better see a sub-project created for each Deliverable. That would be
> more in-line with PMBOK/ISO 21500.
> 1. In the deliverable/Analytic plan, conceptually are we talking about the
>    cost of our resources for the integrator (as in COGS for integrator) or for
>    the customer (as Revenue for the integrator)
> 
> There would be two objects:
> Deliverable Plan Item => When confirmed, creates planned revenue in the
> Analytic Plan
> Resource Plan Item => When confirmed, creates Planned costs in the Analytic
> Plan.

From the Resource Plan you could work out a method to derive the
Deliverables. For example, create a single deliverable for a Project
considering quantities for all Resources planned, but using the sale price
for those resources.

Thanks again for all your input making clear this functional requirement

> Let me know if I forgot or misinterpreted anything
> 
> It is important to understand what drives cost and what drives revenue.
> Deliverables drive revenue. Resources drive cost. Cost and revenue are
> analytic lines in real, and analytic planning lines in plan.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/OCA/project/pull/127#issuecomment-159906429.

## 

Jordi Ballester Alomar
Founder | Eficent
(+34) 629530707 | jordi.ballester@eficent.com | http://www.eficent.com
Twitter: https://twitter.com/jbeficent_erp | Linkedin:
https://www.linkedin.com/in/jordiballesteralomar

elicoidal:
@jbeficent 

> Not really necessary, but can be handy in a simplified version. We have an
> object "Change Request" (
> https://github.com/projectexpert/pmis/tree/8.0/change_management) where we
> can add BR's. Then, in the BR you would have the related Change Request.

To be added to the roadmap (besides your modules have not yet landed in OCA)

>  I see a conflict of terminologies. A deliverable is a description of an
> output resulting of some work (using resources). So I do not see a
> deliverable = estimation at all.
> I see the deliverable as the output. E.g. a product/service that is
> obtained as end result. That deliverable can then be used as the basis for
> the creation of the SO Line, or it can be the SO line itself.
> A deliverable can be achieved as the result of work / materials / services
> executed.
> 
> The key thing is how do you put a price to a deliverable based on the work
> / materials / services executed. SAP uses a concept called Dynamic Item
> Processor (
> http://help.sap-ag.de/saphelp_globext607_10/helpdata/en/d4/9b86687fee11d3a74400104b59c1ec/content.htm),
> that will use the planned costs (materials, services,...) to determine the
> end product and sale price.
> We see deliverables being a separate entity (e.g. object "deliverable").
> Project deliverables can be used as the basis for the creation of SO's.
> I still see Estimates as the Resource Plan (
> https://github.com/projectexpert/pmis/tree/8.0/analytic_resource_plan).
> Analytic plan would be lower level, since it only deals with costs, and
> cannot trigger procurements.

I think we are online but correct me if I am wrong:
- Deliverable is valued at "Sales Price"/revenue and is added to SOL
- estimate is valued at Cost price and might be added to the AA (for budgeting)
  Now I just want a module to answer my current constraint which is to be able to generate quotations out of BR and I think my approach is correct. 
  I do not want to make it more complex because I do not want to rewrite the Analytic Plan which is perfect for deliverable/resource/estimation purpose. Would you agree?

>  This additional module will add BR => Deliverables.
> Then I see as BR => Deliverables => Estimates
> Example:
> BR: I need a faster horse
> Deliverable: Car
> Estimates: 1 Engine, 4 Wheels,....

Correct (but again I have no intention now to go that deep, see above)

> I can better see a sub-project created for each Deliverable. That would be more in-line with PMBOK/ISO 21500.

I like the idea: To be added to the roadmap (it is now possible manually)

> It is important to understand what drives cost and what drives revenue. Deliverables drive revenue. Resources drive cost. Cost and revenue are analytic lines in real, and analytic planning lines in plan.

Indeed and thanks a lot for your valuable input. This is now crystal clear :)
I will proceed with the development part to propose a new PR hopefully today, maybe on Monday

xie8899:
@elicoidal, i fix the error of cannot install the module in our trunk.

business_requirement/README.rst
```diff
@@ -0,0 +1,72 @@
...
+Configuration
+=============
+
+No configuration steps required
+
+
+
+Usage
+=====
+
```
dreispt, line 159: As a guideline, here I expect to see steps that guide me to evaluate the feature, after I click on the Try Me button.

business_requirement/data/business_sequnece.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dreispt, line 1: This file is not in the manifest and has a typo in the name.

business_requirement/models/business.py
```diff
@@ -0,0 +1,152 @@
...
+        'Gap',
+        readonly=True,
+        states={'draft': [('readonly', False)]}
+    )
+    drop = fields.Boolean(
+        string='Drop?',
+        default=False,
+        readonly=True,
+        states={'draft': [('readonly', False)]},
+        help="""Determine whether this business requirement is to be kept
```
dreispt: You have a state field. Can't it be used for this?

business_requirement/models/business.py
```diff
@@ -0,0 +1,152 @@
...
+from openerp import api, fields, models
+
+
+class BusinessRequirement(models.Model):
+    _name = "business.requirement"
+    _description = "Business Requirement"
+
+    sequence = fields.Char(
+        'Sequence',
+        readonly=True,
```
dreispt: index=True

business_requirement/models/business.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @api.model
+    def create(self, vals):
+        if vals.get('name', '/') == '/':
+            vals['name'] = self.env['ir.sequence'].get('business.requirement')
+        return super(BusinessRequirement, self).create(vals)
+
+    @api.multi
+    @api.depends(
+        'parent_id')
```
dreispt: why this line break?

business_requirement/models/business.py
```diff
@@ -0,0 +1,152 @@
...
+    def _get_level(self):
+        def _compute_level(br):
+            level = 1
+            if br.parent_id:
+                level += _compute_level(br.parent_id)
+            return level
```
dreispt: All good, but I can't help suggesting an alternate one line implementation:

```
return parent_id and parent_id.level + 1 or 1
```

business_requirement/models/business.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @api.model
+    def _get_states(self):
+        states = [
+            ('draft', 'Draft'),
+            ('confirmed', 'Confirmed'),
+            ('approved', 'Approved'),
+            ('done', 'Done'),
+            ('cancel', 'Cancel'),
+        ]
```
dreispt, line 206: In case you feel "cancel" is not expressive enough for dropped requirements, consider adding a "drop" state.

business_requirement/views/business_view.xml
```diff
@@ -0,0 +1,149 @@
...
+                                <group string="Gap">
+                                     <field name="gap" widget="html" nolabel="1"/>
+                                </group>
+                            </page>
+                            <page string="Sub Business Requirement">
+                                <field name="business_requirement_ids" readonly="1" context="{'form_view_ref':'view_business_requirement_form', 'tree_view_ref':'view_business_requirement_tree'}"/>
+                            </page>
+                            <page string="Other Information">
+                                <group col="4">
+                                    <field name="categ_id"/>
```
dreispt: I would place Categories in the document header, outside the notebook.

business_requirement/views/business_view.xml
```diff
@@ -0,0 +1,149 @@
...
+                                    <field name="parent_id"/>
+                                    <field name="level"/>
+                                    <field name="drop"/>
+                                    <field name="change_request"/>
```
dreispt: Change request flag should definitely be in the header

business_requirement/views/business_view.xml
```diff
@@ -0,0 +1,149 @@
...
+                            </page>
+                            <page string="Sub Business Requirement">
+                                <field name="business_requirement_ids" readonly="1" context="{'form_view_ref':'view_business_requirement_form', 'tree_view_ref':'view_business_requirement_tree'}"/>
+                            </page>
+                            <page string="Other Information">
+                                <group col="4">
+                                    <field name="categ_id"/>
+                                    <field name="parent_id"/>
+                                    <field name="level"/>
+                                    <field name="drop"/>
```
dreispt: As mentioned before, dropped should go away as it's the same as the cancelled state

business_requirement/views/business_view.xml
```diff
@@ -0,0 +1,149 @@
...
+
+        <record id="view_business_requirement_tree" model="ir.ui.view">
+            <field name="name">business.requirement.tree</field>
+            <field name="model">business.requirement</field>
+            <field name="arch" type="xml">
+                <tree string="Business Requirement">
+                    <field name="sequence"/>
+                     <field name="name"/>
+                     <field name="description"/>
+                     <field name="categ_id"/>
```
dreispt, line 123: add change request

business_requirement/views/business_view.xml
```diff
@@ -0,0 +1,149 @@
...
+        <record id="view_business_requirement_search" model="ir.ui.view">
+            <field name="name">business.requirement.search</field>
+            <field name="model">business.requirement</field>
+            <field name="arch" type="xml">
+                <search string="Business Requirement">
+                    <field name="name" string="Name"/>
+                    <field name="sequence" string="Sequence"/>
+                    <group expand="0" string="Group By">
+                        <filter string="Category" icon="terp-go-home"  context="{'group_by':'categ_id'}"/>
+                        <filter string="State" icon="terp-go-home" context="{'group_by':'state'}"/>
```
dreispt, line 144: I think "icon" is deprecated

business_requirement/models/business.py
```diff
@@ -0,0 +1,152 @@
...
+        compute='_get_level',
+        string='Level',
+        store=True
+    )
+    change_request = fields.Boolean(
+        string='Change Request?',
+        default=False,
+        readonly=True,
+        states={'draft': [('readonly', False)]}
+    )
```
dreispt, line 100: I think we need an additional related field and model, to organize BRs around a "Solution" under design. As sales time the Solution can be a bid/SO, at execution time can be a Project.

dreispt:
Just made a full code review, mostly minor comments.
But I think it needs to have the  ability to organize BRs in "Solutions" or equivalent.

business_requirement/models/business.py
elicoidal, line 206: Agree

business_requirement/README.rst
elicoidal, line 159: We need to add screenshots: I'll wait for the module to be stabilized before doing it

business_requirement/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+# © 2015
+# Elico Corp, contributor: Eric Caudal, Alex Duan, Xie XiaoPeng
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': 'Business Requirement',
+    'category': 'Business Requirements Management',
+    'summary': 'Business Requirement',
+    'version': '8.0.1.0.0',
+    'website': 'www.elico-corp.com',
+    "author": "Elico Corp(S), Odoo Community Association (OCA)",
```
elicoidal: remove "(S)"

business_requirement/__openerp__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# © 2015
```
elicoidal: for all company copyrights:
© 2015 Elico Corp (http://www.elico-corp.com)

business_requirement/data/business_sequnece.xml
```diff
@@ -0,0 +1,20 @@
...
+<openerp>
+    <data noupdate="1">
+
+        <!-- Sequences for sale.order -->
```
elicoidal: Remove comment or adapt it

business_requirement/models/business.py
elicoidal: Agree.
@xie8899 please remove this field and create a state "drop"

business_requirement/models/business.py
elicoidal, line 100: This is still a WIP. We need to add the project link in BR (Required) and add a smart button in project to see all BR.
Later for CRM purpose we will add the link of the project in the opportunity.

business_requirement/views/business_view.xml
elicoidal: agree

business_requirement/views/business_view.xml
elicoidal, line 123: @dreispt Would not be better to add change request in a separate module (link to the @jbeficent module?) or later on. All the more that those modules have not  yet landed in OCA AFAIK

elicoidal:
@dreispt thanks for the review. Note that that this is still half baked. I have the concept clear but we need to finish a first clean version for business_requirement and business_requirement_deliverable
We are still working internally on it (hopefully today or tomorrow we should have a much cleaner version).

dreispt:
OK, I'll mark these modules as WIP.
Please move them to Needs Review as soon as they can be reviewed.

elicoidal:
@dreispt We have cleansed a lot and probably today or tomorrow I can get it under review

elicoidal:
@jbeficent @dreispt This PR starts to be mature. I have tested it functionally and it could still be improve but I think it is usable now.
it is linked to PR #138 (deliverable - mature), #143 (crm - stable but improving) and #144 (project - stable but improving)

elicoidal:
@jbeficent @dreispt @pedrobaeza can you help me finishing the review here?

dreispt:
Mr.Travis is not happy, needs some love.

dreispt:
AFAICT my original review comments are still relevant and unaddressed.

elicoidal:
@dreispt Thanks for the check up. This is national holidays here and we have made multiple modifications that we will push only next week. I will work on the README files as well.
For your info we are already using it in production and this is a huge step forward in terms of internal process even if features are not yet fully stable

business_requirement/models/business.py
```diff
@@ -0,0 +1,481 @@
...
+    partner_id = fields.Many2one(
+        related='project_id.partner_id',
+        store=True,
+        readonly=True,
+    )
+    sub_br_count = fields.Integer(
+        string='Count',
+        compute='_sub_br_count'
+    )
+    linked_project = fields.Many2one(
```
elicoidal, line 117: @victormartinelicocorp This should be push to module business_requirement_project

business_requirement/models/business.py
```diff
@@ -0,0 +1,481 @@
...
+        string='Linked project',
+        comodel_name='project.project',
+        groups='project.group_project_manager',
```
elicoidal, line 120: make it available for project user

business_requirement/models/business.py
```diff
@@ -0,0 +1,481 @@
...
+        readonly=True,
+        states={'draft': [('readonly', False)]}
+    )
+    gap = fields.Text(
+        'Gap',
+        readonly=True,
+        states={'draft': [('readonly', False)]}
+    )
+    categ_id = fields.Many2many(
+        'project.category',
```
elicoidal, line 60: Not sure about this one: shouldnot it be business_requirement.category?

elicoidal, line 60: Besides, the tags should be searchable in tree view

martintorresvm, line 60: @elicoidal depends if you want to have the same tags (project.category) for `project.task` and `business.requirement` or have separated ones.
Other comment, yes the tags should be searchable in tree view.

elicoidal, line 60: @victormartinelicocorp Current problem is that the views in the BR menu does not allow to handle the category in the BR.
Side note: I think BR categories and Project categories are 2 different things so I am comfortable with the idea of having BR category in the BR.

elicoidal:
replaced by #179

PR #136, opened 2015-11-25
[ADD] module 'create_project_from_analytic'

JordiBForgeFlow:
This module allows to create a Project from an existing Analytic Account. This creation is only possible for Analytic Accounts of type 'Contract or Project', that are not already assigned to an existing Project.

In the Analytic Accounts list view you can select the analytic accounts that you want, and then go to 'More' and 'Create Project'.

### DIFF
--- a/create_project_from_analytic/README.rst
+++ b/create_project_from_analytic/README.rst
@@ -1,0 +1,69 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+====================================
+Create Project from Analytic Account
+====================================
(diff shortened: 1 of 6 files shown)


JordiBForgeFlow:
Unless I'm missing something fundamental on projects, if you delete a project the analytic account stays, but you cannot reverse the situation to again have a project for that AA. This module makes it possible to do that.

This is useful in cases where the analytic account already has analytic lines, and the project was deleted by a human error. Or also useful when a company started to use an analytic account and after a time changed their mind and now want to convert it to a project.

pedrobaeza:
Are you sure? If you uncheck "Tasks", save, and mark again?

JordiBForgeFlow:
@pedrobaeza Thanks. I'm closing this PR now.. (brrrrr I need a coffee)

PR #144, opened 2015-12-16
[ADD] business_requirement_project

xie8899:


elicoidal:
@jbeficent @dreispt This PR starts to be mature. I have tested it functionally and it could still be improve but I think it is usable now.
it is linked to PR 127 (deliverable - mature), 138 (deliverable - mature), 143 (crm - stable but improving).
We should push a PR today.

elicoidal:
replaced by https://github.com/OCA/business-requirement/pull/9

PR #145, opened 2015-12-19
[FIX] project_task_materials_stock: Error when create a project from template

sergio-teruel:
[FIX] project_task_materials_stock: Duplicate entries if user configure more than one stage to consume material
@rafaelbn needs functional review 
@carlos-incaser review please

### DIFF
--- a/project_task_materials_stock/models/project.py
+++ b/project_task_materials_stock/models/project.py
@@ -50,24 +50,29 @@ class Task(models.Model):
 
     @api.multi
     def unlink_stock_move(self):
+        res = False
         moves = self.mapped('stock_move_ids')
-        moves.filtered(lambda r: r.state == 'assigned').do_unreserve()
-        moves.filtered(
-            lambda r: r.state in ['waiting', 'confirmed', 'assigned']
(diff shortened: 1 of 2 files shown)


sergio-teruel:
Pending.. review tests

rafaelbn:
Tested functional :+1: super fix thanks!

We must fix travis, runbot is not related

sergio-teruel:
@rafaelbn @dreispt travis fixed. Thanks

rafaelbn:
@dreispt this is a FIX and it's ready to be merged, do we need another review? could you make it? thanks

yajo:
Could you provide a test for this fix?

yajo:
Functional review works, just waiting for the test to armor this.

sergio-teruel:
@Yajo the module already has tests.

yajo:
I mean a test that proves that this fix is needed.

JavierIniesta:
Tested in runbot :+1:

JMODaERP:
I´ve tested in runbot :+1:

PR #149, opened 2016-02-02
* [ADD] Adding new module : project_issue_code

mikevhe18:

PR #150, opened 2016-02-04
[8.0][project_recalculate]Error W503

mikevhe18:
I've got this error on my PR #149 

screenshoot-travis-error

It's because the preferred place to break around a binary operator is after the operator, not before it.
So, I fix it in this PR.


pedrobaeza:
:+1: 

Fast-tracking so you can rebase the other PR.

PR #151, opened 2016-02-05
[8.0][ADD] project_earned_value_task

JordiBForgeFlow:
This module allows you to manage the progress of your projects using the
Earned Value Management technique. For a detailed explanation, see the
`Slides <https://www.odoo.com/es_ES/slides/slide/project-management-with-the
-lights-on-earned-value-management-using-odoo-283>`_ and `Video <https://www
.youtube.com/watch?v=lvCodX1TK5E>`_ of the conference at Odoo Experience
2015.


project_earned_value_task/__openerp__.py
```diff
@@ -0,0 +1,27 @@
...
+    "name": "Project Earned Value Task",
+    "summary": "Manage the progress of your projects using the Earned Value "
+               "Management technique.",
+    "version": "8.0.1.0.0",
+    "author": "Eficent Business and IT Consulting Services SL,"
+              "Odoo Community Association (OCA)",
+    "website": "www.eficent.com",
+    "category": "Generic Modules/Projects & Services",
+    "depends": ["project", "hr_timesheet"],
+    "init_xml": [],
```
dreispt: deprecated, remove

project_earned_value_task/data/project_data.xml
```diff
@@ -0,0 +1,33 @@
...
+        </record>
+        <record id="project.project_tt_deployment" model="project.task.type">
+            <field name="poc">100</field>
+        </record>
+        <record id="project.project_tt_cancel" model="project.task.type">
+            <field name="poc">0</field>
+        </record>
+    </data>
+
+</openerp>
```
dreispt, line 33: Why is this needed?

project_earned_value_task/models/project.py
```diff
@@ -0,0 +1,472 @@
...
+    _inherit = 'project.task.type'
+    poc = fields.Float('% Completed')
+
+
+class Project(models.Model):
+    
+    _inherit = "project.project"
+
+    @staticmethod
+    def _get_evm_ratios(ac, pv, ev, bac):
```
dreispt: Can you add a docstring to shed light on the arguments?

project_earned_value_task/models/project.py
```diff
@@ -0,0 +1,472 @@
...
+
+        return res
+
+    def _get_earliest_latest_dates(self):
+        project = self
+        cr = self.env.cr
+        # Get the earliest and latest dates for the associated tasks
+        cr.execute("""SELECT MIN(date_start)
+                   FROM project_task
+                   WHERE project_id=%s""", (project.id, ))
```
dreispt: shouldn't it be `%d`?

project_earned_value_task/models/project.py
```diff
@@ -0,0 +1,472 @@
...
+        min_date_start = cr.fetchone()[0] or 0
+        if min_date_start == 0:
```
dreispt: I think the `== 0` is unnecessary

project_earned_value_task/models/project.py
```diff
@@ -0,0 +1,472 @@
...
+            datetime_start = datetime.today()
+        else:
+            datetime_start = datetime.strptime(min_date_start,
+                                               "%Y-%m-%d %H:%M:%S")
```
dreispt: It would be best to use import the server date format from the ORM

project_earned_value_task/models/project.py
```diff
@@ -0,0 +1,472 @@
...
+        WHERE PT.project_id = %s""", (self.id, ))
+        res = cr.fetchone()
+        if res:
+            return res[0] or 0.0
+        else:
+            return 0.0
+
+    def _get_plan_cost_to_date(self):
+        cr = self.env.cr
+        today = datetime.today().strftime('%Y-%m-%d %H:%M:%S')
```
dreispt: What if we want to measure this at a different point in time, such as end of last month when preparing reporting?
I suggest accepting an additional optional argument with the date.

project_earned_value_task/models/project.py
```diff
@@ -0,0 +1,472 @@
...
+        if max_date_end_2 == 0:
+            datetime_end_2 = datetime.today()
+        else:
+            datetime_end_2 = datetime.strptime(max_date_end,
+                                             "%Y-%m-%d %H:%M:%S")
+        datetime_end = max(datetime_end, datetime_end_2)
+
+        return datetime_start, datetime_end
+
+    def _get_budget_at_completion(self):
```
dreispt: This is the same as `_get_plan_cost_to_date` using end date as measuring date. You should reuse that function.

project_earned_value_task/models/project.py
```diff
@@ -0,0 +1,472 @@
...
+        AND ip.res_id=CONCAT('product.template,',PRT.id)
+        AND ip.company_id=PT.company_id)
+        WHERE PT.project_id = %s""", (self.id, ))
+        res = cr.fetchone()
+        if res:
+            return res[0] or 0.0
+        else:
+            return 0.0
+
+    @api.one
```
dreispt, line 242: deprecated for v9

dreispt:
Travis needs some love, there's also an [XML error](https://travis-ci.org/OCA/project/jobs/107192364#L429), and some minor comment from my side.
But overall LGTM :shipit:

project_earned_value_task/models/project.py
elicoidal: Agree: probably start_date and end_date

dreispt:
@jbeficent Ping!

JordiBForgeFlow:
@dreispt Sorry! Been juggling with too many balls lately. I'm now looking into it.

project_earned_value_task/data/project_data.xml
JordiBForgeFlow, line 33: This is just a proposal of the percentage completed for the task stages that are predefined in the system. If we do not set it, the POC will default to 0 for all stages as a default.

project_earned_value_task/models/project.py
JordiBForgeFlow: Agreed. I'm changing it in all the cases found.

project_earned_value_task/models/project.py
JordiBForgeFlow: the SELECT MIN statement will return 0 if the date_start is not maintained. That's the reason for this condition.

project_earned_value_task/models/project.py
JordiBForgeFlow: You mean tools.DEFAULT_SERVER_DATETIME_FORMAT, right? Now changing it

project_earned_value_task/data/project_data.xml
dreispt, line 33: Ah! OK

project_earned_value_task/models/project.py
dreispt: Unless you want to specifically exclude `None`values, it's more pythonic to write `if min_date_start:`

project_earned_value_task/models/project.py
dreispt: Or better yet: use `fields.Datetime.from_string(min_date_start)`

project_earned_value_task/models/project.py
JordiBForgeFlow: @dreispt regarding measurement of EVM at a different point of time, I suggest the same approach as in the analytic chart of accounts: https://github.com/odoo/odoo/blob/8.0/addons/analytic/analytic.py

if context.get('from_date', False):
            where_date += " AND l.date >= %s"

Because the evm values are used in function fields, and we cannot pass values to those.

project_earned_value_task/models/project.py
JordiBForgeFlow: Got it. Done

project_earned_value_task/models/project.py
JordiBForgeFlow: Got it!

project_earned_value_task/models/project.py
JordiBForgeFlow, line 242: Should I then just remove it?

project_earned_value_task/models/project.py
dreispt: ok

project_earned_value_task/models/project.py
dreispt, line 242: This targets v8 so I guess you can keep it

elicoidal, line 242: @dreispt if deprecated, better find a better alternative. isnot there anything to improve this?
We need to think ahead about migration.

elicoidal:
@jbeficent Any progress on this?

PR #155, opened 2016-02-11
add a few german translations

rschnapka:


dreispt:
Hi @rschnapka , thanks for the contribution.

I can't comment on the translation itself, but I noticed the PO fileshave a recent export date and advertise `Project-Id-Version: OpenERP Server 7.0`. Did you export them from Odoo?

Also, in case you're not aware, translations may be easier to contributed at Transifex: https://www.transifex.com/oca/OCA-project-8-0/

rschnapka:
Hello @dreispt, 
if present I just edited the existing de.po. Mostly updating with the module.pot file, if present. Incidentally in this case I did not check the header data, will update the project-id later on...
And yes I know about and use Transifex. Yesterday I was very active with repositories mostly not having Repos in Transifex and just tripped over 'project' in that effort. Usually I would use Transifex as well.
Thank You much.

dreispt:
I'll wait a few days for other reviews.
If none happens, ping me back and I'll merge it.

PR #156, opened 2016-02-18
8.0 move project department

dreispt:
Following discussion on the Contributors ML: https://odoo-community.org/groups/contributors-15/contributors-31944


pedrobaeza:
Shouldn't we do this on v9?

dreispt:
This will be done only for 9.0.

PR #158, opened 2016-02-18
8.0 move project task

dreispt:


dreispt:
This will be done only for 9.0.

PR #159, opened 2016-02-24
[ADD] project_task_programmed module 8.0

astirpe:
Plan task as a reminder for any date field as deadline.
Any date field can be used. How much time in advance the task should be created can be specified.


rafaelbn:
Hi @astirpe , thanks for this contribution.

Maybe you are interested, take a look in:
- Base module: https://github.com/yelizariev/addons-yelizariev/tree/8.0/reminder_base
- For tasks: https://github.com/yelizariev/addons-yelizariev/tree/8.0/reminder_task_deadline
- Others:
  - https://github.com/yelizariev/addons-yelizariev/tree/8.0/reminder_crm_next_action
  - https://github.com/yelizariev/addons-yelizariev/tree/8.0/reminder_issue_deadline
  - https://github.com/yelizariev/addons-yelizariev/tree/8.0/reminder_phonecall

We are planning a more general approach for alerts in Odoo as @yelizariev did. But we need some more time as is a big work.

Also say that v9 already integrates alerts with web browser. Could be interesting analyze them and maybe a back-port.

rafaelbn:
cc @jbeficent

astirpe:
Thank you @rafaelbn !
The module task_alert is meant to only remind a deadline by creating a task rather than an event.
Of course,a more general approach for alerts would be preferred.

rafaelbn:
So, you have a task with deadline: 2016/04/01, and then with this module you can schedule that a second task will be created automatically one month before (2016/03/01). Is it correct?
Then, module names doesn't fit to me. But I'm not sure... I would like to test it functionally. Could you fix travis and runbot please?

astirpe:
The functionality is a bit more generic, lets describe it this way:
- you have a generic object with deadline: 2016/04/01; doesn't matter if the object is an invoice, an order, a task, an issue, etc...
- then with this module you can schedule that a new task will be created automatically one month before (2016/03/01).

The module name was only meant to be technical, but I agree it doesn't give a quick overview of the functionality.

task_alert/README.rst
```diff
@@ -0,0 +1,28 @@
...
+
+A cron job named "Create alert tasks" will be created by installing this module. This job can be found via:
+
+	**Settings > Technical > Automation > Scheduled Actions**
+
+This job will run daily by default.
+
+|
+
+**Contact us:**
```
rafaelbn: Delete please

task_alert/README.rst
```diff
@@ -0,0 +1,28 @@
...
+
+When you have any remark about this module, please let us know on http://www.onestein.eu/feedback
```
rafaelbn: Delete please. This should be done through OCA.
Please read https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md
And use OCA template https://github.com/OCA/maintainer-tools/tree/master/template/module

task_alert/__openerp__.py
```diff
@@ -0,0 +1,42 @@
...
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
```
rafaelbn: You only need https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L2

task_alert/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
```
rafaelbn: As said check tamplate
Example: https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py#L2

rafaelbn:
Hi @astirpe , do you mean that Invoices hava deadline? 

> you have a generic object with deadline: 2016/04/01; doesn't matter if the object is an invoice, an order, a task, an issue, etc...

And when that deadline from invoice will generate a task?

> then  with this module you can schedule that a new task will be created automatically one month before (2016/03/01)

I will try to understand testing it, so please as said in my comments review module template and in the README add secction usage where you can explain more detailed how to use this module.

Thanks!

task_alert/README.rst
dreispt: @rafaelbn True, but I wonder if it wouldn't be fair to let the original author to also add a link offering support. Off topic , of course, but could be an interesting discussion.

astirpe:
@rafaelbn Thank you for the review! Still unit tests to be done...
@dreispt Good idea, for the time being I removed all the references to the Company.

dreispt:
As a matter a fact I am about to use automatically created Tasks for approaching dates/deadlines.
My use case is the renovation of documentation, such as professional certifications: x days before the expiration date I will create a Task to handle the renovation.
Why a Task instead of a Calendar Event? Because it's a non trivial task, involving interaction with official entities and I wand to document the status and evolution of the process. 

Now, my solution is to use standard Automated Actions for this.
What does this module provide that an Automated Action can't do?

dreispt:
...I can think of one limitation for Automated Actions: time Actions are triggered only once for each Document. This means that it would work for the first renovation only, and won't work for the later ones...

task_alert/README.rst
elicoidal: @dreispt @rafaelbn it could be a good discussion to open (I know some contributors do not like it some wants to).
In the meantime, let's stick to the current rule (no company ad)

task_alert/doc/index.rst
```diff
@@ -0,0 +1,21 @@
+Create a project to which the tasks will be related:
```
elicoidal: What is this file for?
This is not standard OCA AFAIK: please confirm.

astirpe: This file was already deleted with the last commit

task_alert/README.rst
rafaelbn: But that's why there is an "Contributors" section. There you have an email.
I agree i will be a nice discussion.
@dreispt @elicoidal Would be correct for example to add in "Cotnributors" section add a line like:
- Onestein Feedback (info@onestein.nl)

??

dreispt:
The module name should begin with `project_`, and I'm not sure if `_alert` is the best choice.
How about `project_task_programmed`?

astirpe:
@dreispt I prefer _project_task_programmed_ as a module name, definitely! I will rename the module when the unit tests will be committed. In the meanwhile let's the discussion go on, maybe some other good candidates will be proposed. Thanks!

task_alert/README.rst
elicoidal: @rafaelbn @dreispt 
Actually, the way to provide feedback should be via github (issues/PR), not via one of the contributors.
We cannot open this door: feedback should be centralized to offer standard response to issues/questions/modifications.
This section is to acknowledge the contributors and their work, not to provide a way of posting bugs (there is another section for that with standard Github management)

rafaelbn: Correct. So as said, this lines must be deleted.

dreispt:
I'm glad you liked that name.
I think my concerns about the relevance of the module are solved: this feature needs to work recurrently, and Automated Actions can't do that out of the box (though they could be improved for that...).
So let's go on with the review.

dreispt:
I'm glad you liked that name.
I think my concerns about the relevance of the module are solved: this feature needs to work recurrently, and Automated Actions can't do that out of the box (though they could be improved for that...).
So let's go on with the review.

project_task_programmed/models/project_task_alert.py
```diff
@@ -0,0 +1,104 @@
...
+             "field to send out the alert.",
+        required=True
+    )
+    date_field_id = fields.Many2one(
+        'ir.model.fields',
+        string="Date field",
+        domain=[('ttype', 'in', ('date', 'datetime'))],
+        required=True
+    )
+    task_name = fields.Char(
```
dreispt: To follow conventions, I would use `name` for this.

project_task_programmed/models/project_task_alert.py
```diff
@@ -0,0 +1,104 @@
...
+            days_delta = timedelta(days=task_alert.days_delta)
+            to_date = (datetime.now() + days_delta).strftime(DF)
+            args = [
+                (task_alert.date_field_id.name, '<=', to_date),
+                (task_alert.date_field_id.name, '!=', False)
+            ]
+            if not task_alert.last_run:
+                task_alert.last_run = datetime.now()
+            last_run_dt = datetime.strptime(task_alert.last_run, DF)
+            last_run = (last_run_dt).strftime(DF)
```
dreispt: You do convert the date from text to date, and the from date back to text.
What is the point?

project_task_programmed/models/project_task_alert.py
```diff
@@ -0,0 +1,104 @@
...
+            model_name = task_alert.date_field_id.model_id.model
+            rec_ids = self.env[model_name].search(args)
+            for rec in rec_ids:
+                # Check if the task has already been created
+                if self.env['project.task'].search([
+                    ('alert_res_id', '=', rec.id),
+                    ('alert_model_name', '=', rec._name),
+                    ('alert_field_name', '=', task_alert.date_field_id.name)
+                ]):
+                    continue
```
dreispt: We don't want this, or would have the same issue as Automated Actions.
Use case: ff an Employee's ID document is renewed, we write the new date on the same Employee form, and want a second alert to be created for the same Document ID.

project_task_programmed/models/project_task_alert.py
```diff
@@ -0,0 +1,104 @@
...
+                    'name': task_name,
+                    'project_id': task_alert.project_id.id,
+                    'user_id': userid,
+                    'description': description,
+                    'alert_model_name': rec._name,
+                    'alert_res_id': rec.id,
+                    'alert_field_name': task_alert.date_field_id.name,
+                }
+                self.env['project.task'].create(task_data)
+            task_alert.last_run = datetime.now()
```
dreispt: You should use `fields.Datetime.now()` instead.
In general, you instead of `import datetime` you should  use the equivalent functions provided by the ORM `fields.Datetime`

dreispt:
Tests are failing.

project_task_programmed/models/project_task_alert.py
astirpe: @dreispt thank you for the review! I added a check on the dates to handle this case.

project_task_programmed/models/project_task_alert.py
```diff
@@ -57,15 +55,12 @@ def _merge_placeholders(self, incoming, rec):
     def create_task_alerts(self):
         for task_alert in self:
             days_delta = timedelta(days=task_alert.days_delta)
-            to_date = (datetime.now() + days_delta).strftime(DF)
+            to_date = fields.Datetime.now() + days_delta
```
dreispt: Careful, this won't work: `fields.Datetime.now()` returns a string in the server expected format.
It´s perfect for a field default value, but not for date math.

ehahouimohssine:
I have tested it (The functionality works well) and got some feedback points:
- In the Task Alerts view, you see a field called DateField. Many names in ODOO have the same name (example: Create date) so you can not easily see which model it is. Maybe you can show the name of the field and the model of the field (Example Create_date / Employee).
- If a task is created, I can not see which item the Task alerts are triggered. I only see the taskname. 

The above feedback points are nice to have.

antoniocanovas:
It works as expected

dreispt:
Hey @astirpe , I did some tests and found that this feature can be provided by standard Automated Actions: if you edit a "deadline date" for a later time, the AA will be triggered again when the time threshold is met. The issue of AAs not being triggered again only happens when the reference date field is kept unchanged.

If we can achieve the goal with a standard feature, no point for a module, right?

astirpe:
I'm going forward with the improvements of the module [project_task_programmed](https://github.com/onesteinbv/addons-onestein/tree/8.0/project_task_programmed). You will find the latest code in this repository: [github.com/onesteinbv/addons-onestein](https://github.com/onesteinbv/addons-onestein).

In case OCA is still interested in this module, I'm ready to share the latest code.
Also any suggestion/improvement will be appreciated and well accepted.

Thanks to you all!

dreispt:
@astirpe Thank you for submitting your code to the OCA. Please continue doing that every time you have a feature that might be interesting to other people.

astirpe:
I think is time to close this old PR

PR #161, opened 2016-02-26
Change color of issue depending of state_sla

AlexandroTapia:
Change color of issue depending of state_sla
state               color
warning    = green
will fail      = orange
Failed       = red

### DIFF
--- a/project_sla/project_sla_control.py
+++ b/project_sla/project_sla_control.py
@@ -99,8 +99,19 @@ class SLAControl(orm.Model):
             for sla in self.browse(cr, uid, ids, context=context):
                 doc = self.pool.get(sla.doc_model).browse(
                     cr, uid, sla.doc_id, context=context)
+                colors = {
+                    '1': 0,
+                    '3': 4,
+                    '4': 3,
+                    '5': 2,
+                }
                 if doc.sla_state < new_state:
-                    doc.write({'sla_state': new_state})
+                    doc.write(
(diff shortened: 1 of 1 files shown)


dreispt:
Thanks for the PR @AlexandroTapia.
IMO this is too opinionated and removes users the ability to choose how to use colors.

I must say I'm using something like this is production, but I use Automated Action rules for it.
That has the advantage of being more customizable.
Maybe instead we could add sample Automated Actions for this, either as demo or default data.

AlexandroTapia:
Could you give me an example of one of your Automated Actions?

dreispt:
Of course:

Create an Automated Action:
-  When to Run: `On Update`
-  Before Update Filter: create filter `[('sla_state', 'not in', ['4', '5']), ('state', 'not in', ['done','cancelled'])]`
-  Filter: create filter `[('sla_state', 'in', ['4', '5']), ('state', 'not in', ['done','cancelled'])]`
- Actions: create and use an Action with
  - Base Model: `Task`
  -  Action To Do: `Write on a Record`
  -  Condition: `obj.color not in [2]`
  -  Update Policy:  `Update the current record`
  - Value Mapping: `Color Index | Value | 2`

snap 2016-03-03 at 13 04 49

It's easier than in sounds.
A similar Automated Action can be created to set to Yellow (color=3) on the SLA Warning state.

PR #163, opened 2016-03-23
[9.0] [ADD] project_timesheet_time_control

pedrobaeza:
# 
# Project Timesheet Time Control

This module adds a button at account analytic line level to compute minutes
lasted from start date to the current moment.

It adds a datetime field that replaces `date` field in tree view, and write

It also adds a filter by user and some groups by task and by user.

Finally, it allows to open and close tasks from account analytic lines.


pedrobaeza:
@sergio-incaser @carlosdauden @elicoidal @rafaelbn Ready to be reviewed

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -0,0 +1,71 @@
...
+                     ('fold', '=', False)], limit=1)
+                line.task_id.write({'stage_id': stage.id})
+
+    @api.multi
+    def button_close_task(self):
+        for line in self.filtered('task_id'):
+            if line.task_id.project_id:
+                stage = self.env['project.task.type'].search(
+                    [('project_ids', '=', line.task_id.project_id.id),
+                     ('fold', '=', True)], limit=1)
```
danimaribeiro: What about raise an UserError when there is no fold stage? I did one test and the task goes 'undefined' 
undefined

project_timesheet_time_control/views/account_analytic_line_view.xml
```diff
@@ -0,0 +1,55 @@
...
+            </field>
+        </record>
+
+        <record id="account_analytic_line_search" model="ir.ui.view">
+            <field name="name">account.analytic.line.search</field>
+            <field name="model">account.analytic.line</field>
+            <field name="arch" type="xml">
+                <search string="Search task work">
+                    <field name="user_id"/>
+                    <group expand="0" string="Group By...">
```
danimaribeiro: I coudn't find these filters in the view. Shouldn't this inherit from "hr_timesheet.hr_timesheet_line_search"

project_timesheet_time_control/README.rst
```diff
@@ -0,0 +1,65 @@
...
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/140/9.0
+
+Known issues / Roadmap
+======================
+
+* *sale_service* module adds a check at project stage called "Is a closed
+  stage", but we can't use this check, because it has a lot of dependencies
+  (including sale module), so we use the check "Folded in Tasks Pipeline" for
+  considering if a task is closed or not.
```
dreispt: Wise choice.

project_timesheet_time_control/__openerp__.py
```diff
@@ -0,0 +1,23 @@
...
+# © 2016 Pedro M. Baeza
+# © 2016 Sergio Teruel
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    'name': 'Project timesheet time control',
+    'version': '9.0.1.0.0',
+    'category': 'Project',
+    'author': 'Serv. Tecnol. Avanzados - Pedro M. Baeza, '
+              'Antiun Ingeniería S.L.',
```
dreispt: Missing the OCA

project_timesheet_time_control/hooks.py
```diff
@@ -0,0 +1,12 @@
...
+# © 2016 Pedro M. Baeza
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+
+def post_init_hook(cr, registry):
+    """Put the date with 00:00:00 as the date_time for the line."""
+    cr.execute("""
+        UPDATE account_analytic_line
+        SET date_time = to_timestamp(date || ' 00:00:00',
+                                     'YYYY/MM/DD HH24:MI:SS')
```
dreispt: Add a WHERE clause, so that does not reset hours on a module reinstall

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @api.onchange('task_id')
+    def onchange_task_id(self):
+        if self.task_id:
+            self.account_id = self.task_id.project_id.analytic_account_id.id
+
+    def eval_date(self, vals):
+        if 'date_time' in vals and 'date' not in vals:
+            vals['date'] = fields.Date.from_string(vals['date_time'])
+        return vals
```
dreispt, line 41: Since this is not a model specific feature, I would rather have it as a function outside the class.
I suggest the name `dates_from_strings()`.

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -0,0 +1,71 @@
...
+    @api.multi
+    def button_end_work(self):
+        end_date = datetime.now()
+        for line in self:
+            date = fields.Datetime.from_string(line.date_time)
+            line.unit_amount = (end_date - date).total_seconds() / 3600
+        return True
+
+    @api.multi
+    def button_open_task(self):
```
dreispt: I was expecting this to be named `button_start_task`.

dreispt:
Reviewed: 
- Design looks good, but IMO it would be more robust if we added the `base_stage_stage` dependency, and used it's states to identify the "open" and "close" states. Currently it depends on an error prone heuristic - the first not folded, the first folded. To get in trouble it is as simple as putting the "Cancelled" stage before the "Done" one.
- Some minor notes on the implementation.

project_timesheet_time_control/models/account_analytic_line.py
pedrobaeza, line 41: I prefer not, because it can't be easily accessed (you need a complicated import for using it in an inherited class).

pedrobaeza:
cc @Tecnativa

rafaelbn:
HI @dreispt , which is the fix? It works fine. We are working on this daily :smile: 

We can add your comment in known issue (although I heard "wise choice"). For a future improvement.

@antespi @yajo take a look please

IMHO this is a :+1:

project_timesheet_time_control/README.rst
sbidoul: I personally think relying on the folded state for this is fragile. If you later install sale_service  there is two mechanism to indicate a task is closed and it becomes very confusing.

You could depend on [project_stage_closed](https://github.com/OCA/project/tree/8.0/project_stage_closed) to have a more robust solution.

pedrobaeza: Yes, good idea. We'll work on this.

dreispt:
No issue on my side :+1:

project_timesheet_time_control/__openerp__.py
rvalyi: :-1: still missing

pedrobaeza: Yeah, we haven't still touched this. Even the credits are outdated :wink:

I'll do it this afternoon

project_timesheet_time_control/views/account_analytic_line_view.xml
yajo: You probably meant `analytic.view_account_analytic_line_filter`.

project_timesheet_time_control/hooks.py
yajo: Actually it's OK because `date_time` field is created by this addon, and AFAIK `post_init_hook` does not get executed on upgrades.

project_timesheet_time_control/i18n/es.po
```diff
@@ -0,0 +1,82 @@
...
+#. module: project_work_time_control
+#: model:ir.ui.view,arch_db:project_work_time_control.work_time_control_timesheet_account_analytic_line_tree
+msgid "Open task"
+msgstr "Abrir tarea"
+
+#. module: project_work_time_control
+#: model:ir.ui.view,arch_db:project_work_time_control.work_time_control_analytic_line_search
+#, fuzzy
+msgid "Search task work"
+msgstr "Actualizar Modo Trabajo"
```
yajo: This is wrong.

project_timesheet_time_control/i18n/es.po
```diff
@@ -0,0 +1,82 @@
...
+
+#. module: project_work_time_control
+#: model:ir.ui.view,arch_db:project_work_time_control.work_time_control_timesheet_account_analytic_line_tree
+msgid "Stop"
+msgstr "Stop"
```
yajo: Detener

project_timesheet_time_control/i18n/es.po
```diff
@@ -0,0 +1,82 @@
...
+#. module: project_work_time_control
+#: model:ir.model,name:project_work_time_control.model_project_task
+#: model:ir.ui.view,arch_db:project_work_time_control.work_time_control_analytic_line_search
+msgid "Task"
+msgstr "Tarea"
+
+#. module: project_work_time_control
+#: model:ir.model,name:project_work_time_control.model_project_task_type
+msgid "Task Stage"
+msgstr "Estado actual"
```
yajo: Etapa actual

project_timesheet_time_control/i18n/es.po
```diff
@@ -0,0 +1,82 @@
...
+
+#. module: project_work_time_control
+#: model:ir.model.fields,help:project_work_time_control.field_account_analytic_line_folded
+msgid ""
+"This stage is folded in the kanban view when there are no records in that "
+"stage to display."
+msgstr ""
```
yajo: Missing

project_timesheet_time_control/security/project_security.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
```
yajo: `<odoo>`

project_timesheet_time_control/security/project_security.xml
```diff
@@ -0,0 +1,22 @@
...
+        <field name="model_id" ref="model_account_analytic_line"/>
+        <field name="domain_force">['|',
+                                        ('user_id', '=', user.id),
+                                        '|',
+                                        ('task_id.user_id', '=', user.id),
+                                        '|',
+                                            ('task_id.project_id.privacy_visibility', 'in', ['public', 'employees']),
+                                            '&amp;',
+                                                ('task_id.project_id.privacy_visibility', '=', 'followers'),
+                                                ('task_id.message_follower_ids', 'in', [user.partner_id.id]),
```
yajo: You have the `user.partner_id.ids` shortand. Not critical though :wink:

project_timesheet_time_control/security/project_security.xml
```diff
@@ -0,0 +1,22 @@
...
+<data noupdate="0">
+
+    <record id="account_analytic_line_visibility_rule" model="ir.rule">
+        <field name="name">Account Analytic Line: employees: public or employee or following or assigned</field>
+        <field name="model_id" ref="model_account_analytic_line"/>
+        <field name="domain_force">['|',
+                                        ('user_id', '=', user.id),
+                                        '|',
+                                        ('task_id.user_id', '=', user.id),
+                                        '|',
```
yajo: Convention seems to be starting the domain by all the ORs `['|', '|', '|', ...]`. It works the same anyway, so nevermind, but please fix the indentation here, is quite confusing.

project_timesheet_time_control/views/account_analytic_line_view.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
```
yajo: `<odoo>`?

project_timesheet_time_control/views/account_analytic_line_view.xml
```diff
@@ -0,0 +1,55 @@
...
+                            icon="STOCK_STOP"
+                            attrs="{'invisible': [('unit_amount', '>', 0)]}"
+                    />
+                </field>
+                <field name="task_id" position="after">
+                    <field name="folded" invisible="1"/>
+                    <button name="button_close_task"
+                            string="Close task"
+                            type="object"
+                            icon="gtk-apply"
```
yajo: I thought gtk icons were getting deprecated in favor of fontawesome ones.

yajo: You could use http://fontawesome.io/icon/folder-o/ for example here...

project_timesheet_time_control/views/account_analytic_line_view.xml
```diff
@@ -0,0 +1,55 @@
...
+                            attrs="{'invisible': [('folded', '=', True)]}"
+                            />
+                    <button name="button_open_task"
+                            string="Open task"
+                            type="object"
+                            icon="gtk-redo"
```
yajo: ... and http://fontawesome.io/icon/folder-open-o/ here. Or maybe the other way around. Not sure what's clearer...

project_timesheet_time_control/views/project_task_view.xml
```diff
@@ -0,0 +1,28 @@
...
+                    <attribute name="invisible">1</attribute>
+                </xpath>
+                <xpath expr="//field[@name='timesheet_ids']//field[@name='date']" position="after">
+                    <field name="date_time"/>
+                </xpath>
+                <xpath expr="//field[@name='timesheet_ids']//field[@name='unit_amount']" position="after">
+                    <button name="button_end_work"
+                            string="Stop"
+                            type="object"
+                            icon="STOCK_STOP"
```
yajo: Options here:

http://fontawesome.io/icon/hourglass-end/
http://fontawesome.io/icon/times-circle/

yajo:
I see a bug in runbot:
1. Log in as admin.
2. Leave demo user with these permissions:
   
   captura de pantalla de 2016-08-26 16-40-52
3. Log in as demo.
4. Go to Timesheets > Activities > Create.
5. Fill activity name & task (do not fill the project, let it be autofilled).
   
   ```
   Odoo Warning - Access Error
   The requested operation cannot be completed due to security restrictions. Please contact your system administrator.
   
   (Document type: project.project, Operation: read)
   ```

I tested the same workflow in core Odoo runbot v9 and it works (well, instead of autofilling, you simply cannot choose a task before a project. With this PR if you do that, you see no tasks listed, that could be another related bug).

rafaelbn:
Could you please fix this branch to go ahead with this PR please?

pedrobaeza:
I have fixed all the comments. Please review @dreispt @Yajo @luismontalba @chienandalu

luismontalba:
Testing it on runbot I see it's OK but I can't see the "filter by user":

seleccion_001

pedrobaeza:
The filter appears typing in the search box. Anyway, it's in the standard finally, so I don't add it in this module. Read again the new README.

luismontalba (approved):
Tested on runbot

yajo (changes requested):
Code: 👍 
Functional: 👎  because of this:

1. Log in as admin and set demo user's permissions only to Project/User and HR/Employee.
2. Log in as Demo User.
3. Go to Timesheets > Activities and try to create one.

```
Odoo Warning - Access Error
Sorry, you are not allowed to modify this document. Only users with the following access level are currently allowed to do that:
- Project/Manager

(Document model: project.task.type)
```

I'm pretty sure you shouldn't need to be a project manager to enter an activity.

pedrobaeza:
Again a damn related field without readonly that is overwritten although there's no change... Fixed.

Please check again, @Yajo.

project_timesheet_time_control/hooks.py
dreispt: But it gets executed on reinstalls (`-i`), so it wouldn't hurt to add that condition.

project_timesheet_time_control/README.rst
```diff
@@ -0,0 +1,56 @@
...
+  date field with datetime field value.
+* It also adds a task filter and a group by task options in the search.
+* Finally, it allows to open and close tasks from account analytic lines.
+  The selected closed stage is the first one that is found with the mark
+  "Closed" checked.
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
```
dreispt: The current text talks about "account analytic line", but end users will be more familiar with "Project Tasks" and "Timesheet Lines", so I find it unclear as is.
It would help to add here a list of steps for the end user that show the typical use case.

project_timesheet_time_control/hooks.py
pedrobaeza: Are you sure about this? I don't think so.

pedrobaeza: Are you sure about this? I don't think so.

yajo: Hmm I don't think reinstalls actually exist in Odoo world. Either you update, or you uninstall and install.
However adding the `WHERE` is not much complex, so thinking twice about it, seems legit.

yajo (changes requested):
Still failing.

1. Set the same permissions (project user and employee) to demo user and log in as him.
2. Go to activities and create a row.
3. Enter something in the description and choose the task "Document history management".

```
Odoo Warning - Access Error
The requested operation cannot be completed due to security restrictions. Please contact your system administrator.

(Document type: project.project, Operation: read)
```

pedrobaeza:
All comments fixed and @Yajo's problem is also solved. @dreispt would you do the honor of merging for continuing with the last PR?

dreispt:
Done!

pedrobaeza:
Thank you!

PR #167, opened 2016-04-12
[ADD] project_double_alias

pedrobaeza:
# 
# Double alias in projects

Odoo by default allows to define an alias to receive mails, but you can only
select to create tasks or claims related to the project, but not both.

This module makes possible to create an alias for each.
# Installation

On installation time, this module will move all the existing project aliases
that point to issues to the second alias. If the configured alias is for tasks,
an alias for issues is searched in the system anyway for being linked.

This operation will be reverted back when uninstalling (move second aliases to
the first one when no alias is defined for tasks).

In case you have projects which have configured both aliases for tasks and
issues, both are preserved on uninstallation of this module, but you will
only have accessed to the task one.
# Configuration
- Enable developer mode.
- Go to _Settings > Parameters > System parameters_.
- Create or edit the parameter with the name "mail.catchall.domain", and put
  as value the domain of your mail server.
# Usage
- Go to a project
- Access the "E-mails" tab.
- Define a mail alias for the tasks and a mail alias for issues


pedrobaeza:
cc @rafaelbn @carlosdauden @sergio-teruel

carlosdauden:
:+1: Code review

project_double_alias/__openerp__.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# © 2016 Pedro M. Baeza <pedro.baeza@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Double alias for project",
+    "summary": "Define an alias for tasks and another alias for issues",
+    "version": "8.0.1.0.0",
```
sergio-teruel: Version 8.0 or 9.0? in readme file at line 57 the version is 9.0

pedrobaeza: Indeed, it's version 9!

project_double_alias/views/project_project_view.xml
```diff
@@ -0,0 +1,54 @@
...
+                    <field name="second_alias_id" class="oe_read_only oe_inline" />
+                    <div class="oe_edit_only oe_inline" name="edit_second_alias" style="display: inline;" >
+                        <field name="second_alias_name" class="oe_inline"/>@<field name="alias_domain2" class="oe_inline"/>
+                    </div>
+                </div>
+            </div>
+        </field>
+    </record>
+
+    <record model="ir.ui.view" id="view_project_kanban">
```
sergio-teruel: First id key

sergio-teruel:
@pedrobaeza review done, check it.

project_double_alias/views/project_project_view.xml
pedrobaeza: It doesn't matter. We can change it only for readability. In fact, I copy/paste the Odoo view, but you know that they don't really follow conventions at all.

pedrobaeza:
@sergio-teruel, changes done

sergio-teruel:
Perfect... so :+1:

dreispt:
Functionally sounds good. Technically looks well built. :+1:

PR #169, opened 2016-04-26
Added feature to stop generating invoices after contract goes in overdue

agb80:


dreispt:
@agb80 I think the right place for this module is OCA/contract. 
Do you agree @pedrobaeza ?

pedrobaeza:
@dreispt I agree, but remember that the contract repo was created after 7.0, and this module still resides here for 7.0 version. If we move the module, existing installations will break. That's why we keep it here.

dreispt:
@pedrobaeza Sorry, I didn't notice it was a 7.0 module. Agreed then.

account_analytic_analysis_recurring/account_analytic_analysis_recurring.py
```diff
@@ -66,7 +66,7 @@ def product_id_change(
             self, cr, uid, ids, product, uom_id, qty=0, name='',
             partner_id=False, price_unit=False, pricelist_id=False,
             company_id=None, context=None):
-        context = context or {}
+        context = dict(context or {})
```
dreispt, line 69: I don't see the point for this.

account_analytic_analysis_recurring/account_analytic_analysis_recurring.py
```diff
@@ -123,6 +123,16 @@ class AccountAnalyticAccount(orm.Model):
...
         'recurring_next_date': fields.date('Date of Next Invoice'),
+        'state': fields.selection(
+            [('template', 'Template'),
+             ('draft', 'New'),
+             ('open', 'In Progress'),
+             ('pending', 'To Renew'),
+             ('overdue', 'Overdue'),
+             ('close', 'Closed'),
+             ('cancelled', 'Cancelled')], 'Status', required=True,
+            track_visibility='onchange'
```
dreispt, line 134: It would be nice if you could add comments identifying what this redefinition changes from the standard.

dreispt:
Question: shouldn't we have a flag in the Contract to allow avoiding specific contracts to have invoicing suspended? There are always special cases...

dreispt:
Travis build fails because of `analytic_hours_block` :confused:

dreispt:
...but tests pass on module [unit test](https://travis-ci.org/OCA/project/jobs/139804342#L2055) :smile:

account_analytic_analysis_recurring/account_analytic_analysis_recurring.py
```diff
@@ -166,7 +166,10 @@ def _prepare_invoice_line(self, cr, uid, line, invoice_id, context=None):
         contract = line.analytic_account_id
         fpos = contract.partner_id.property_account_position or False
         account_id = fpos_obj.map_account(cr, uid, fpos, account_id)
-        taxes = product.taxes_id or False
+        taxes = [
+            x for x in product.taxes_id if
+            x.company_id.id == context['force_company']
```
dreispt, line 171: What if no `force_company` key is in the context?

pedrobaeza:
Comments hasn't been attended for a long time, so I close.

PR #170, opened 2016-04-27
[9.0] extract analytic_base_department

yvaucher:
Moving https://github.com/OCA/department/pull/56 here


dreispt:
@yvaucher Maybe this PR is more appropriate for OCA/contract.

yvaucher:
@dreispt Maybe more in https://github.com/oca/account-analytic ? As we already have a dependency on this repo in oca_requirements.txt ?

dreispt:
Sounds good to me. 
Maybe we should also check with some of the contributors there.
@sbidoul @lepistone @guewen : does adding departments to analytic account fit well into the OCA/account-analytic repo?

yvaucher:
Moved in https://github.com/OCA/account-analytic/pull/57

PR #176, opened 2016-05-19
[8.0][NEW] project_task_analytic_partner: New module

carlosdauden:
This module adds the partner on task work lines and propagate it to the auto-generated analytic entries on the "Other partner" field.


project_task_analytic_partner/README.rst
```diff
@@ -0,0 +1,64 @@
...
+   :alt: License: AGPL-3
+
+=====================
+Partner in task works
+=====================
+
+This module adds the partner on task works and propagate it to the
+auto-generated analytic entries on the "Other partner" field.
+
+
```
pedrobaeza, line 13: Add Installation section mentioning the dependencies to OCA/account-analytic

project_task_analytic_partner/README.rst
```diff
@@ -0,0 +1,64 @@
...
+=============
+
+* Go to *Settings -> Configuration -> Project -> Project Management*.
+* Enable *Record timesheet lines per tasks*.
+
+
+Usage
+=====
+
+Go to Project > Tasks, and you will find the partner field in work lines.
```
pedrobaeza: Mention also about the purpose of this field.

pedrobaeza:
Little comments, but :+1:

project_task_analytic_partner/__openerp__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# © 2016 Carlos Dauden <carlos.dauden@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Partner in task works',
+    'summary': 'Select alternative partner in task works',
```
dreispt: I think it should be "task work" or "task work lines"

project_task_analytic_partner/README.rst
dreispt, line 13: AFAICS this module requires HR Timesheet to work.
This is not clear from the title, and should be explained here.

project_task_analytic_partner/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+    'website': 'http://www.tecnativa.com',
+    'depends': [
+        'analytic_partner_hr_timesheet_invoice',
+        'project_timesheet',
+    ],
+    'data': [
+        'views/project_view.xml',
+    ],
+    'installable': True,
+    'autoinstall': True,
```
dreispt: Why?

dreispt:
The README needs some improvements.
Other than that it's OK.

carlosdauden:
Changes done

project_task_analytic_partner/__openerp__.py
```diff
@@ -0,0 +1,22 @@
...
+              "Odoo Community Association (OCA)",
+    'website': 'http://www.tecnativa.com',
+    'depends': [
+        'analytic_partner_hr_timesheet_invoice',
+        'project_timesheet',
+    ],
+    'data': [
+        'views/project_view.xml',
+    ],
+    'installable': True,
```
dreispt, line 21: @pedrobaeza Should this one also be auto installable?

pedrobaeza, line 21: Well, this is where you can have the control to enable/disable the feature. We can put it if we want, as the same explanation of the other applies also to this one. What you prefer.

dreispt, line 21: No opinion here. I'm not familiar with the `analytic_partner_hr_timesheet_invoice` module or the related use case.
Whatever makes more sense to you. If in doubt, don't use autoinstall.

rafaelbn:
Amazing, thanks! :+1: Tested in runbot!

PR #182, opened 2016-07-04
[ADD] project_task_category: Add categories on the Tasks

martintorresvm:


martintorresvm:
@elicoidal

project_task_category/README.rst
```diff
@@ -0,0 +1,48 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+==================================
+Project Task Configurable Category
+==================================
+
+This module extends the functionality of Project to add a category on the task and support filter and group.
+
```
elicoidal, line 10: The difference with tag is that this category is m2o and can be used for filtering and more important grouping or reporting.

project_task_category/README.rst
```diff
@@ -0,0 +1,48 @@
...
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Rona Lin <rona.lin@elico-corp.com>
+* Victor-M Martin Torres <victor.martin@elico-corp.com>
```
elicoidal, line 35: Eric Caudal eric.caudal@elico-corp.com

elicoidal:
small details in README.
:+1:

project_task_category/views/project_task_view.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+        <record id="view_task_form2" model="ir.ui.view">
+            <field name="name">project_task_form_view_type</field>
+            <field name="model">project.task</field>
+            <field name="inherit_id" ref="project.view_task_form2"/>
+            <field name="arch" type="xml">
+                <field name="project_id" position="before">
+                    <field name="task_categ_id" sequence='0'/>
```
dreispt: Hmm, I didn't know about the `sequence` view attribute.
What does it do?

dreispt:
I have a suggestion: add a `sequence` field to the Categories, and use that for sort order instead.

Review: :+1:

martintorresvm:
@elicoidal @dreispt

rafaelbn:
Hi I have tested in runbot but I don't find that field to set the category. Please check my screenshoot:



Runbot: http://3150033-182-e2fb64.runbot1.odoo-community.org/web?db=3150033-182-e2fb64-all#view_type=kanban&model=ir.module.module&menu_id=53&action=37

project_task_category/models/task_category.py
```diff
@@ -11,8 +11,9 @@ class TaskCategory(models.Model):
     name = fields.Char('Name', required=True, translate=True, select=True)
     description = fields.Char('Description', translate=True)
     active = fields.Boolean('Active', default=True)
+    sequence = fields.Integer('Sequence')
```
dreispt, line 14: Thanks!

martintorresvm:
@elicoidal @dreispt I made an improvement based on your recommendation.

@rafaelbn I'll need to check why is not visible the Category field on Runbot.

elicoidal:
@victormartinelicocorp once @rafaelbn issue is fixed, let's merge

dreispt:
I confirm @rafaelbn 's issue.
Digging into it, the `task_categ_id` [is already added](https://github.com/OCA/project/blob/8.0/project_categ/project_categ_model.py#L80) by the `project_categ` module.

That module allows you to organize tags into domain areas.
- it adds a tree hierarchy to Tags (`project.category`)
- allows to provide a "Root Category" in projects
- then only child Tags are selectable for Tasks in that Project

That usage is compatible with what we want with this module.
My suggestion is to reuse it, and relabel the "Category Root" to "Category" and make it editable.

elicoidal:
@dreispt not sure about your meaning: are you suggesting to make this module dependent of the 'project_categ' or just rename the ids so that both can live together.
We looked into that module time ago but it didnot fulfil our needs which is why we added this one.
I don't think this is good to have them depend on each other.

dreispt:
@elicoidal Slightly changing the field name here would solve the problem. But I don't see why the `project_categ` can´t do what we need here with my suggested adjustments there.

elicoidal:
@dreispt 
I took some time to test the module and I agree that it could be easily compatible.
Nevertheless I have an issue with project_categ module: it makes the tag mandatory (which in my opinion should be solved in another module).
If we inherit, we would have 2 fields for category and tags and both would be mandatory. I think (but I could be wrong/subjective):
- Category should be mandatory as this defines reporting or could trigger other method (as it is unique)
- Tags should be not mandatory as it is more relevant for search/reminders/communication

your thoughts?

dreispt:
@elicoidal My idea is to reuse it, not to inherit it. IMO it's OK to change it so that tags are not Tags required. That sjhould be a configuration to be added later.

elicoidal:
@dreisp, I am sorry but I am not sure to follow what you want to do
Do you mean changing the original module and adding our category to it?
Do you put in our module the code from the project_categ module?
If both modules have the same feature, I dont see it (whoever would have 
both modules would have duplicated features).
I am OK to converge whatever the way, just need to understand your point

dreispt:
The Category field is already provided by the existing module.
You just need to make it editable (and make the Tags field optional again).
No need for a new module.

elicoidal:
OK Clear now!
We will amend the existing one and base the business_requirement_deliverable_category module on this one.
Thanks for your patience
( @victormartinelicocorp)

dreispt:
So it's OK to close this PR?

elicoidal:
Yes closing: we will adjust the other module

PR #183, opened 2016-07-04
[IMP] project_task_materials_stock: Warning to prevents remove invoiced lines

carlosdauden:


project_task_materials_stock/models/project.py
```diff
@@ -73,7 +73,14 @@ def write(self, vals):
                         task.material_ids.create_analytic_line()
                 else:
                     if task.unlink_stock_move():
-                        task.analytic_line_ids.unlink()
+                        if any(task.material_ids.mapped(
+                                'analytic_line_id.invoice_id')):
+                            raise exceptions.Warning(
+                                _(
+                                    "You can't move to not consume stage "
```
pedrobaeza: You can't move to a not consume stage if there are already invoiced analytic lines

pedrobaeza:
:+1:

rafaelbn:
Tested in runbot :+1:

sergio-teruel:
:+1:

dreispt:
This is a PR from a person from Tecnativa, fast merged based on positive reviews by three reviewers from Tecnativa.
There's nothing explicitly against this in the OCA guidelines, but I believe that is not in the spirit of the OCA. There should be two or three independent reviews. In this case, I feel that the PR should have waited for at least one other (independent) opinion.

pedrobaeza:
Are you saying that we are not independent enough to give a critical review to our colleagues? You can see all our PRs to check our method. Please don't doubt of our integrity.

dreispt:
Please don't get me wrong: my comment is about the process, not about your judgement or integrity. 

I just believe that it's not a good principle for a PR to be validated only by work colleagues.
It should be given the change for people outside a team to also reviews.
Thus I believe that there should be at least two or three independent reviews.
Otherwise, it can set a precedence that in the future can be abused to push changes into the OCA.

In this case, it would be best if someone outside your team had the chance to also comment on the code before it was merged.

pedrobaeza:
I don't think so. OCA is about peer review, not about "peer that are not from your company" review. If the criteria of any of the reviewers are somehow abused, being for PRs of their colleagues or other PRs (it can happen also if I want a module that someone develops to be ASAP in OCA), then that should be penalized.

If we had lots of reviewers, then this wouldn't be needed, and we will get our work reviewed by other people, but it's not the case just right now.

rafaelbn:
What @dreispt said is correct,  It could happens. But Daniel, as @pedrobaeza said then PSC must penalize it. I could be agree with you if the interest of merging is first and the interest of a good code and stable module is second. How to know this? The PSC will detect this.
We must take in care that all this work is public so Reviewer / Programmer 's reputation is at stake, so I don't think anyone want to loose this because a simple PR.

elicoidal:
I trust @pedrobaeza and his team but @dreispt has a point.
Having an independent view is part of the peer review and we should try (and I say try in a good effort perspective) to get to this point

project_task_materials_stock/models/project.py
```diff
@@ -83,7 +83,13 @@ def write(self, vals):
                         task.material_ids.create_analytic_line()
                 else:
                     if task.unlink_stock_move():
-                        task.analytic_line_ids.unlink()
+                        if any(task.material_ids.mapped(
+                                'analytic_line_id.invoice_id')):
+                            raise exceptions.Warning(
+                                _("You can't move to a not consume stage if "
```
elicoidal, line 89: "You cannot move back to a not-consumed stage if analytic lines have been invoiced"

pedrobaeza, line 89: It's not "move back", but only "move", because the order of the stages can be twisted in any way.

elicoidal, line 89: OK: just adapt accordingly the sentence as the original one was a little bit strange.

PR #186, opened 2016-07-27
[9.0][project_description] Migrate from v8

yajo:
Almost no changes.

@Tecnativa @rafaelbn


dreispt:
:+1:

rafaelbn:
:+1:

pedrobaeza:
:+1:

dreispt:
Pedro, although this seems a trivial PR, please don't merge so fast.
We should leave some time for other people to comment.

pedrobaeza:
Again this conversation? The module has 3 approvals (even one not of my team), so there's no need for that. Don't fall again on the same thing.

dreispt:
Well, you already know my opinion here.
It should really be discussed on the proper place - I'll open the discussion as soon as I find some time for that.

pedrobaeza:
Ok, you can start a discussion to change the current rules if you want, but take into account that:

1) We are following the rules
2) This is totally incoherent with your optimistic merge proposal
3) The current problem with OCA is the lack of contributors for reviewing, which is not a problem in my team, that always review our work between each other. If you try to block this option, you'll be obstructing the OCA itself.

PR #191, opened 2016-08-29
Project project recalculate

alfredoavanzosc:
The field "calculation type" has 2 values: "date_begin", and "date_end", and it is initialized to False. In the "project recalculate" function asks for the following condition: "if project.calculation_type != 'none'", this condition is never given, because the field is initialized to False.

### DIFF
--- a/project_recalculate/models/project_project.py
+++ b/project_recalculate/models/project_project.py
@@ -64,10 +64,9 @@ class ProjectProject(models.Model):
                     project.date):
                 raise Warning(_("Cannot recalculate project because your "
                                 "project don't have date end."))
-            if project.calculation_type != 'none':
-                for task in project.tasks:
-                    task.task_recalculate()
-                vals = project._start_end_dates_prepare()
-                if vals:
-                    project.write(vals)
+            for task in project.tasks:
+                task.task_recalculate()
+            vals = project._start_end_dates_prepare()
+            if vals:
+                project.write(vals)
         return True


pedrobaeza:
Please check your PR. It's not correct.

pedrobaeza:
I have fixed it with the new "Change base" feature.

pedrobaeza:
:+1: 

@antespi, please check

oihane:
@antespi @pedrobaeza @alfredoavanzosc  I think it is enough by erasing the line with `if project.calculation_type != 'none'`, as there is a first if that controls if calculation type has a value. The field is not required, so you are allowed to give no value, if you take a look to the view the button will not be visible if value is not set. What do you think?

anajuaristi:
@antespi @pedrobaeza @oihane 
I need this patch merged ASAP because it's blocking one of our custom modules to be merged. Please If you could make it today it would be great.

Thank you:

Ana

oihane:
@anajuaristi  I don't think this patch is correct, if adding 'none' to the selection is the solution there are some things that aren't fixed.

rafaelbn:
Hi, when @oihane give her OK we will review it. Thanks!

antespi:
@alfredoavanzosc this is a bug, right, but @oihane is right, the line to change  is https://github.com/OCA/project/blob/588ee1922d717811fe915e531d225626ebb7f233/project_recalculate/models/project_project.py#L67:

```
if project.calculation_type:
   (...)
```

This will avoid to make the calculation if calculation_type is set to False

@alfredoavanzosc, please undo your changes and change only that line

oihane:
@antespi many many thanks for your response, sorry for any inconveniences.

@alfredoavanzosc the only time that it will raise the warning is if there is not calculation_type set and you try to recalculate it, because you shouldn't, so is more correct not to add the 'none'

oihane:
@antespi @rafaelbn this is another way, LGTM :+1:

project_recalculate/models/project_project.py
```diff
@@ -64,10 +64,9 @@ def project_recalculate(self):
                     project.date):
                 raise Warning(_("Cannot recalculate project because your "
                                 "project don't have date end."))
-            if project.calculation_type != 'none':
-                for task in project.tasks:
-                    task.task_recalculate()
-                vals = project._start_end_dates_prepare()
-                if vals:
-                    project.write(vals)
+            for task in project.tasks:
```
antespi, line 67: Why didn't you add condition here to avoid calculation if `calculation_type` is False?

```
if project.calculation_type:
   for task in project.tasks:
      (...)
   (...)
```

Your current code makes no change

alfredoavanzosc: A ver @antespi , me estoy ahorrando una línea de código, ¿para que voy a poner 
if project.recalculation_type:
   for task in projec.tas:
    (....)

Si antes tengo estas condición:
   if not project.calculation_type:
       raise Warning(_("Cannot recalculate project because your "
                       "project don't have calculation type."))

Si quieres la meto, pero no es para nada necesaria

pedrobaeza: This should be in English.

But if you have the Warning, what is the problem then?

antespi: OK, i didn't see the Warning above.

antespi:
:+1:

pedrobaeza:
I don't see the main goal in the PR, as there's a previous Warning that prevents possible problems.

oihane:
@rafaelbn @pedrobaeza coveralls decreases because there are less lines. Can you take a look and merge it, it has two +1

ghost:
:+1:

antespi:
@pedrobaeza this PR only remove an useless line, that's all.

pedrobaeza:
OK, merging

PR #195, opened 2016-09-28
Multiple default_group_by

sysadminmatmoz:
Without repeating the definition for the stage_ids and defining both stages and timeboxes in _group_by_full, the default fold column option gets messed on the original project.task kanban after installing project_gtd module.

It's intended just as a quick fix, I didn't pep clean the original code thus the travis problems.


dreispt:
I see your point, but if we don't PEP-clean the 8.0 branch will become red 😢

PR #201, opened 2016-10-27
[MIG] 10.0 port project_task_default_stage

astirpe:


project_task_default_stage/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 {
     'name': 'Project Task Default Stage',
     'summary': 'Recovery default task stage projects from v8',
-    'version': '9.0.1.0.0',
+    'version': '10.0.1.0.0',
     'category': 'Project',
     'author': 'Incaser Informatica S.L., '
```
pedrobaeza: Please switch the author to the new company `Tecnativa` instead of `Incaser...`.

pedrobaeza (approved):
Little comment, but is OK (code review)

pedrobaeza:
Thanks

project_task_default_stage/models/__init__.py
```diff
@@ -1,2 +1,3 @@
 # -*- coding: utf-8 -*-
+from . import project_task_type
 from . import project
```
dreispt: It would be best for the Python file to be named after the Model it implements, as originally.

dreispt (approved):

PR #202, opened 2016-11-03
[10.0][ADD][project_task_dependency] Task dependencies on tasks

tarteo:
Enables the user to define dependencies (other tasks) of a task.
e.g. task 'building the roof' is dependent on task 'building walls and pillars'


pedrobaeza:
Please put the module name in singular: `project_task_dependency` according guidelines https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#modules

rafaelbn (approved):
Functional test

Good to know about https://github.com/OCA/project/tree/8.0/project_recalculate module which is very interesting to calculate dates depending in dependencies.

levkar:
It might be better to have two many2many fields (predecessor_ids and successor_ids) which uses same table instead of task_ids.

susport:
You could validate not save tasks with the same name, when a task depends on the other.

bistaray:
What is the best way to review this?  I see no runbot link, and pulling the v10 branch does not provide this module.  (Sorry - am new with GIT)

jcdrubay:
You can pull the code from here: https://github.com/tarteo/project/tree/100_project_task_dependencies

pedrobaeza:
@bistaray, runbot is available clicking on "Show all checks"

bistaray:
@pedrobaeza - thanks - I see this now.  admin / admin isn't working.  How do I discover the credentials?

pedrobaeza:
@bistaray anyone can change the password. I have launched a rebuilt of the instance to have a fresh one with the default password.

project_task_dependencies/models/project_task.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import models, fields
+
+
+class ProjectTask(models.Model):
+    _inherit = 'project.task'
+
+    task_ids = fields.Many2many(
+        comodel_name='project.task', relation='project_task_project_task_rel',
+        column1='task_id', column2='dependent_task_id', string='Dependencies',
```
dreispt: This is not clear to me: the are the Tasks I depend on, or the Tasks the depend on me?
The field name (and string label) should make this clear.

dreispt (changes requested):
There is also a pending request to change the module name.

project_task_dependencies/models/project_task.py
bistaray: "Dependencies" is clear to me, and I'd assert to native English speakers.  See http://www.dictionary.com/browse/dependencies  and https://blogs.msdn.microsoft.com/project/2008/07/28/back-to-basics-understanding-task-dependencies/ 

Another option would be 'Predecessors' (things that come before me)

pedrobaeza:
I remind you that the module names should be in singular form according guidelines.

bistaray:
Using runbot I was able to add task dependencies to another task.  This module only offers the basic data structure.

Possibly users may wish a recursion check - to ensure no circular dependencies.

I will want a way to visualize dependencies across a project - so would need a field 'parent_id' or equivalent - so I can list tasks in a treeview (and group by parent perhaps?) showing the parent task for each task.  @levkar has suggested an improvement that may make visualization easier also.

As it stands it works and is respects the design of Odoo, IMHO.

tarteo:
I've fixed the name and created a mechanism to prevent circular dependencies.

dreispt (approved):

tarteo:
@dreispt Can you remove the needs fixing tag, please?

PR #208, opened 2016-12-18
[9.0] [MIG] project_task_materials_stock module

cubells:
- Updated README
- Updated views
- Updated tests

cc @Tecnativa


project_task_materials_stock/README.rst
```diff
@@ -0,0 +1,73 @@
...
+Field Service, often require you to keep a record of the materials spent.
+
+This module extends project_task_material module to consume products spent in
+a task and create analytic lines to manage costs and create invoices.
+
+Configuration
+=============
+
+#. If you are a project manager, go to Project > Configuration > Stages and
+check option 'Consume Material' in Task Stage to generate a stock move when
```
pedrobaeza: Indent these lines 3 spaces to get a correct visualization

project_task_materials_stock/README.rst
```diff
@@ -0,0 +1,73 @@
...
+the task is in that stage.
+#. Go to Project -> Configuration - > Settings and enable option
+"Manage time estimation on tasks"
+
+Usage
+=====
+
+#. Go to a task, edit, and add materials to be consumed on tab "Materials".
+#. Move task to an stage on consume material are activated and moves and
+analytic lines will be created.
```
pedrobaeza: Indent this

project_task_materials_stock/i18n/project_task_materials_stock.pot
```diff
@@ -0,0 +1,62 @@
+# Translation of Odoo Server.
```
pedrobaeza: Remove this file

pedrobaeza (changes requested):
Minor things to change

pedrobaeza (approved):
Please squash your migration commits. Runbot will be correct when the dependency is merged.

pedrobaeza:
@cubells, can you improve tests for avoiding dropping coverage?

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,180 @@
...
+# Copyright 2016 Tecnativa - Vicent Cubells
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+from openerp import models, fields, api, exceptions, _
+
+
+class ProjectTaskType(models.Model):
+    _inherit = 'project.task.type'
+
+    consume_material = fields.Boolean(
+        string='Consume Material',
```
yajo: Since the string equals the field name, you don't need it.

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,180 @@
...
+        help="If you mark this check, when a task goes to this state,"
```
yajo: Add a space at the end of the string, or it will join without it.

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,180 @@
...
+        for task in self:
+            if not task.stock_move_ids:
+                task.stock_state = 'pending'
+            elif task.stock_move_ids.filtered(
+                    lambda r: r.state == 'confirmed'):
+                task.stock_state = 'confirmed'
+            elif task.stock_move_ids.filtered(lambda r: r.state == 'assigned'):
+                task.stock_state = 'assigned'
+            elif task.stock_move_ids.filtered(lambda r: r.state == 'done'):
+                task.stock_state = 'done'
```
yajo: Probably this would be faster & shorter:

```
if not task.stock_move_ids:
    task.stock_state = 'pending'
else:
    states = task.mapped("stock_move_ids.state")
    for ("confirmed", "assigned", "done") as state:
        if state in states:
            task.stock_state = state
            break
```

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @api.multi
+    def unlink_stock_move(self):
+        res = False
+        moves = self.mapped('stock_move_ids')
+        moves_done = moves.filtered(lambda r: r.state == 'done')
+        if not moves_done:
+            moves.filtered(lambda r: r.state == 'assigned').do_unreserve()
+            moves.filtered(
+                lambda r: r.state in ['waiting', 'confirmed', 'assigned']
```
yajo: Use a set instead of a list, [it's way faster in this case](http://stackoverflow.com/questions/2831212/python-sets-vs-lists#2831242).

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,180 @@
...
+        res = super(Task, self).write(vals)
+        for task in self:
+            if 'stage_id' in vals:
+                if task.stage_id.consume_material:
+                    if not task.stock_move_ids:
+                        task.material_ids.create_stock_move()
+                        task.material_ids.create_analytic_line()
+                else:
+                    if task.unlink_stock_move():
+                        if any(task.material_ids.mapped(
```
yajo: You don't need `any` here AFAIK.

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @api.multi
+    def unlink(self):
+        self.mapped('stock_move_ids').unlink()
+        self.mapped('analytic_line_ids').unlink()
+        return super(Task, self).unlink()
+
+    @api.multi
+    def action_assign(self):
+        self.mapped('stock_move_ids').action_assign()
```
yajo, line 101: Don't you need to `return` it? 🤔

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,180 @@
...
+
+    @api.multi
+    def action_done(self):
+        self.mapped('stock_move_ids').action_done()
```
yajo, line 105: Don't you need to `return` it? 🤔

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class ProjectTaskMaterials(models.Model):
+    _inherit = "project.task.materials"
+
+    stock_move_id = fields.Many2one(
+        comodel_name='stock.move', string='Stock Move')
+    analytic_line_id = fields.Many2one(
+        comodel_name='account.analytic.line', string='Analytic Line')
+    product_uom = fields.Many2one(
```
yajo: Add `_id` suffix. And possibly an `oldname` if this was like this in v8.

project_task_materials_stock/models/project.py
```diff
@@ -0,0 +1,180 @@
...
+            'unit_amount': self.quantity,
+            'account_id': analytic_account.id,
+            'user_id': self._uid,
+            'product_uom_id': self.product_uom.id,
+        }
+        ctx = dict(self._context or {})
+        ctx['uom'] = self.product_uom.id
+        # Compute based on pricetype
+        amount_unit = \
+            self.product_id.with_context(ctx).price_get(
```
yajo: Better do `self.product_id.with_context(uom=self.product_uom.id).price_get(` and remove those `ctx` lines above.

yajo (commented):
Some not required implementation notes, but at least please give your thoughts on them.

dreispt:
Status?

project_task_materials_stock/models/project.py
cubells: It's not needed because this method is introduced in this module

project_task_materials_stock/models/project.py
cubells: It's not needed because this method is introduced in this module

cubells:
All is done @dreispt @pedrobaeza @Yajo

pedrobaeza (approved):

project_task_materials_stock/views/project_view.xml
```diff
@@ -14,46 +13,40 @@
...
+            </field>
+            <field name="material_ids" position="replace">
+                <field name="stock_state" invisible="1"/>
+                <field name="consume_material" invisible="1"/>
+                <group string="Materials consumed" groups="project.group_tasks_work_on_tasks">
+                    <notebook>
+                        <page string="Products">
+                            <field name="material_ids" attrs="{'readonly':[('stock_state','!=', 'pending')]}">
+                                <tree string="Materials used" editable="top">
+                                    <field name="product_id" domain="[('type', 'in', ['consu', 'product'])]"/>
```
yajo, line 37: Would it make sense under some scenarios to be able to put other product types? Such as considering the task has consumed a service product?

yajo (approved):
Code review OK, just a little question...

project_task_materials_stock/views/project_view.xml
pedrobaeza: @carlosdauden, do you remember why we finally discard that option?

cubells: Consume action create an stock move and stock move only can be generated with products that not are services.
That's the reason.

PR #209, opened 2016-12-23
[9.0] [MIG] project_issue_task module

cubells:
- Created README
- Updated file structure
- Updated views

cc @Tecnativa


project_issue_task/__openerp__.py
```diff
@@ -1,53 +1,26 @@
...
-##############################################################################
-
+# Copyright 2015 - 2013 Daniel Reis
+# Copyright 2016 - Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     'name': 'Project Issue related Tasks',
     'summary': 'Use Tasks to support Issue resolution reports',
-    'version': '8.0.1.1.0',
+    'version': '9.0.1.1.0',
```
pedrobaeza: 9.0.1.0.0

project_issue_task/models/project_issue.py
```diff
@@ -0,0 +1,39 @@
...
+from openerp import api, models, _
+
+
+class ProjectIssue(models.Model):
+    _inherit = 'project.issue'
+
+    @api.multi
+    def action_create_task(self):
+        """
+        Create and a related Task for the visit report, and open it's Form.
```
pedrobaeza: Create a task related to the issue(s), and open it(them).

project_issue_task/models/project_issue.py
```diff
@@ -0,0 +1,39 @@
...
+
+            rec_fields = ['project_id', 'analytic_account_id', 'location_id']
+            task_data = dict([(x, getattr(rec, x).id) for x in rec_fields
+                              if hasattr(rec, x) and getattr(rec, x)])
+            task_data['name'] = _('Report for %s') % rec.name
+            task_data['issue_id'] = rec.id
+            task_data['tag_ids'] = [(6, 0, [x.id for x in rec.tag_ids])]
+
+            task_model = self.env['project.task']
+            task_id = task_model.create(task_data)
```
pedrobaeza: Name it `task` only

project_issue_task/models/project_issue.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @api.multi
+    def action_create_task(self):
+        """
+        Create and a related Task for the visit report, and open it's Form.
+        """
+        res = {}
+        for rec in self:
+
+            assert not rec.task_id, _("A Task is already assigned to the "
```
pedrobaeza: Use `raise UserError` instead

project_issue_task/models/project_issue.py
```diff
@@ -0,0 +1,39 @@
...
+            task_data = dict([(x, getattr(rec, x).id) for x in rec_fields
+                              if hasattr(rec, x) and getattr(rec, x)])
+            task_data['name'] = _('Report for %s') % rec.name
+            task_data['issue_id'] = rec.id
+            task_data['tag_ids'] = [(6, 0, [x.id for x in rec.tag_ids])]
+
+            task_model = self.env['project.task']
+            task_id = task_model.create(task_data)
+            rec.write({'task_id': task_id.id})
+            res = {
```
pedrobaeza: This only opens the last created task. Two possibilities:

* Put `self.ensure_one()` and prevent multi-records.
* Add the task to a list, and at the end open a tree or a form view depending on the number of records.

project_issue_task/models/project_issue.py
```diff
@@ -0,0 +1,39 @@
...
+        Create and a related Task for the visit report, and open it's Form.
+        """
+        res = {}
+        for rec in self:
+
+            assert not rec.task_id, _("A Task is already assigned to the "
+                                      "Issue!")
+
+            rec_fields = ['project_id', 'analytic_account_id', 'location_id']
+            task_data = dict([(x, getattr(rec, x).id) for x in rec_fields
```
pedrobaeza: Maybe it's better to just assign directly data for such small number of fields. Please encapsulate this in a `_prepare_...` method.

project_issue_task/models/project_task.py
```diff
@@ -0,0 +1,39 @@
...
+    ref = fields.Char(string='Reference')
+    reason_id = fields.Many2one(comodel_name='project.task.cause',
+                                string='Problem Cause')
+
+    @api.multi
+    def write(self, values):
+        """ On Task Close, also close Issue """
+        stage_obj = self.env['project.task.type']
+        for task in self:
+            if 'stage_id' in values:
```
pedrobaeza: Make the if only one time.

project_issue_task/models/project_task.py
```diff
@@ -0,0 +1,39 @@
...
+                               compute='_fld_issue_id', store=True)
+    ref = fields.Char(string='Reference')
+    reason_id = fields.Many2one(comodel_name='project.task.cause',
+                                string='Problem Cause')
+
+    @api.multi
+    def write(self, values):
+        """ On Task Close, also close Issue """
+        stage_obj = self.env['project.task.type']
+        for task in self:
```
pedrobaeza: Do here `for task in self.filtered('issue_id'):`

project_issue_task/models/project_task.py
```diff
@@ -0,0 +1,39 @@
...
+            if 'stage_id' in values:
+                if stage_obj.browse(values.get('stage_id')).fold:
+                    if self.issue_id:
```
pedrobaeza: Remove this line

project_issue_task/models/project_task.py
```diff
@@ -0,0 +1,39 @@
...
+                        self.issue_id.write(
```
pedrobaeza: `task.issued_id.write(...`

pedrobaeza (changes requested):
Please add tests

project_issue_task/__openerp__.py
```diff
@@ -1,53 +1,26 @@
...
-This module adds some usability improvements:
-
-  * "Create Task" button on the Issue form
-  * Automaticaly Close the Issue when the Task is Closed
-  * Automatically Cancel the Task when Issue is Cancelled
-  * Make the Task also visible to all followers of the related Issue
-""",
-    'author': "Daniel Reis,Odoo Community Association (OCA)",
+    'author': "Daniel Reis,"
+              "Tecnativa, "
```
dreispt, line 11: IMO a port/version migration is not original enough to claim authorship and copyright.

pedrobaeza: I don't agree. We have need the same time to port the module than the original one you need to create it for the first time.  If we only needed to change version in manifest, then we wouldn't change this, but it's not the case, improving the original and adding tests.

project_issue_task/models/project_issue.py
cubells: It's not necessary because these fields no longer exist.

yajo (changes requested):
Definitely this needs tests, some files have a coverage <25%.

project_issue_task/__openerp__.py
dreispt: I took a deeper look at this and you have a point: a significant enough (non trivial) port/translation is a derived work, and can grant copyright to the author. The original author does not lose copyright over the work, and so should be kept as author.
So, in this case it's fair to add to the authors.

dreispt (approved):

pedrobaeza:
@cubells, it's still pending last changes we talked about.

pedrobaeza (approved):

Issue #210, opened 2017-01-01
Project_GTD project.task.copy_data does hard overwrite of defaults instead of checking


ehirt:
Should be: 
        if not default.get('timebox_id', False):    
            default['timebox_id'] = False
        if not default.get('context_id', False):    
            default['context_id'] = False

ehirt:
change was tested on a 6.1 version installation

dreispt:
Can you open a PR with your fix, please?

ehirt:
closed in issue #211

PR #211, opened 2017-01-06
Update project_gtd.py

ehirt:


ehirt:
Pull request for issue #210 
Thanks Daniel for requesting as I did the pull request to my own branch instead of oca branch.

project_gtd/project_gtd.py
```diff
@@ -102,8 +102,10 @@ def copy_data(self, cr, uid, id, default=None, context=None):
             context = {}
         if not default:
             default = {}
-        default['timebox_id'] = False
-        default['context_id'] = False
+        if not default.get('timebox_id', False):
```
dreispt: You can just use `if not default.get('timebox_id'): ` for the same effect.

dreispt (commented):
It's not obvious to me - why is this fix needed?

ehirt:
Hi Daniel,
thanks for the comments. I have streamlined the code accordingly.
The fix is required because the data_copy method does not follow the context and timesheet_id values provided in the default dictionary but overwrites them always with False. This is a reasonable default behaviour such that these data is not copied from the original task, but values for the new task shall be able to be provided through the default dictionary.
Hope the above description is argument enough for the fix

dreispt (approved):
OK, LGTM

ehirt:
Hi Daniel,
How can I make these changes also available in branch 7, 8, 9?
do I have to patch the files per branch and then make individual pull requests?
Thanks
Etienne

dreispt:
@ehirt Yes, we need one PR per branch.

ehirt:
PR is not yet merged -> reopen

ehirt:
Dear Daniel, 
Thanks again for reviewing and approving the PR. Can you also merge it?

dreispt:
@ehirt I'm confused: this PR is for 10.0, but project_gtd hasn't been ported yet...

PR #214, opened 2017-01-18
Ehirt patch project gtd

ehirt:
backport of #211 to branch 9


ehirt:
does not work in this manner

PR #220, opened 2017-02-05
[sale_order_project] Port to Odoo 9.0

dansanti:


dreispt:
This PR does not keep the module's commit history.
[Do that](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0) and then apply your changes on top of it.

PR #224, opened 2017-02-24
[ADD] project_categ_model

martintorresvm:
Add constrain and required field for categ_ids
	modified:   project_categ_model.py

PR #225, opened 2017-02-27
[8.0] [ADD] project_task_category

martintorresvm:
New simple module that have m2o reference to project.category

### DIFF
--- a/project_task_category/README.rst
+++ b/project_task_category/README.rst
@@ -6,8 +6,8 @@
 Project Task Category
 =====================
 
-This module extends the functionality of Project to add an unique mandatory
-category on the task and support filter and group.
+This module extends the functionality of Project to add
+Category field required on the Task with support filter and group by.
 
 Installation
 ============
--- a/project_task_category/README.rst
+++ b/project_task_category/README.rst
@@ -6,7 +6,8 @@
 Project Task Category
 =====================
 
-This module extends the functionality of Project to add an unique mandatory category on the task and support filter and group.
+This module extends the functionality of Project to add an unique mandatory
+category on the task and support filter and group.
 
 Installation
 ============
@@ -29,6 +30,7 @@ Credits
 Contributors
 ------------
 
+* Eric Caudal <eric.caudal@elico-corp.com>
 * Victor Martin <victor.martin@elico-corp.com>
 
 Maintainer
--- a/project_task_category/README.rst
+++ b/project_task_category/README.rst
@@ -1,0 +1,47 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+Project Task Category
+=====================
+
+This module extends the functionality of Project to add an unique mandatory category on the task and support filter and group.
+
+Installation
+============
+
+To install this module, you need to:
+
+ * have basic modules installed (project)
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/project/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Victor Martin <victor.martin@elico-corp.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/project_task_category/README.rst
+++ b/project_task_category/README.rst
@@ -16,6 +16,17 @@ To install this module, you need to:
 
  * have basic modules installed (project)
 
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to project task form view
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot//140/8.0
+
 Bug Tracker
 ===========
 
--- a/project_task_category/models/project_task_category.py
+++ b/project_task_category/models/project_task_category.py
@@ -1,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# © 2016 Elico Corp (www.elico-corp.com).
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openerp import fields, models
+
+
+class ProjectTask(models.Model):
+    _inherit = 'project.task'
+
+    task_categ_id = fields.Many2one(
+        'project.category',
+        'Task Category',
+        required=True
+    )
--- a/project_task_category/models/project_task_category.py
+++ b/project_task_category/models/project_task_category.py
@@ -8,8 +8,8 @@ from openerp import fields, models
 class ProjectTask(models.Model):
     _inherit = 'project.task'
 
-    task_categ_id = fields.Many2one(
+    categ_id = fields.Many2one(
         'project.category',
-        'Task Category',
+        'Category',
         required=True
     )
--- a/project_task_category/models/project_task_category.py
+++ b/project_task_category/models/project_task_category.py
@@ -11,5 +11,4 @@ class ProjectTask(models.Model):
     categ_id = fields.Many2one(
         'project.category',
         'Category',
-        required=True
     )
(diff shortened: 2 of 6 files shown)


project_task_category/README.rst
```diff
@@ -0,0 +1,47 @@
...
+help us smash it by providing detailed and welcomed feedback.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Victor Martin <victor.martin@elico-corp.com>
+
```
elicoidal: Add my name

project_task_category/README.rst
```diff
@@ -0,0 +1,47 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+Project Task Category
+=====================
+
+This module extends the functionality of Project to add an unique mandatory category on the task and support filter and group.
```
elicoidal: no more than 80char.

project_task_category/models/project_task_category.py
```diff
@@ -0,0 +1,15 @@
...
+
+from openerp import fields, models
+
+
+class ProjectTask(models.Model):
+    _inherit = 'project.task'
+
+    task_categ_id = fields.Many2one(
+        'project.category',
+        'Task Category',
```
elicoidal: Category is enough

elicoidal: It should be m2o with the model used in the tags

elicoidal (changes requested):
Model link to be added

project_task_category/models/project_task_category.py
martintorresvm: @elicoidal Maybe I made I mistake but I think actually it is.

martintorresvm:
Please have a check @elicoidal

project_task_category/models/project_task_category.py
elicoidal: LGTM. It is indeed

project_task_category/README.rst
```diff
@@ -0,0 +1,49 @@
...
+category on the task and support filter and group.
+
+Installation
+============
+
+To install this module, you need to:
+
+ * have basic modules installed (project)
+
+Bug Tracker
```
elicoidal: Please add the Runbot

martintorresvm:
@elicoidal Could you approve the changes requested?

martintorresvm:
Message in Runbot is weird to me furthermore Travis is not failing.

2017-03-02 02:18:05,174 186 INFO openerp_test openerp.modules.module: module project_task_category: creating or updating database tables
2017-03-02 02:18:05,202 186 ERROR openerp_test openerp.sql_db: bad query: ALTER TABLE "project_task" ALTER COLUMN "categ_id" SET NOT NULL
File "/root/odoo-8.0/openerp/sql_db.py", line 234, in execute
res = self._obj.execute(query, params)
IntegrityError: column "categ_id" contains null values

2017-03-02 02:18:05,203 186 WARNING openerp_test openerp.models: WARNING: unable to set column categ_id of table project_task not null !
Try to re-run: openerp-server --update=module
If it doesn't work, update records and execute manually:
ALTER TABLE project_task ALTER COLUMN categ_id SET NOT NULL
File "/root/odoo-8.0/openerp/models.py", line 2670, in _auto_init
cr.execute('ALTER TABLE "%s" ALTER COLUMN "%s" SET NOT NULL' % (self._table, k))
File "/root/odoo-8.0/openerp/sql_db.py", line 158, in wrapper
return f(self, *args, **kwargs)
File "/root/odoo-8.0/openerp/sql_db.py", line 234, in execute
res = self._obj.execute(query, params)
IntegrityError: column "categ_id" contains null values

@elicoidal @dreispt @moylop260 @pedrobaeza Could you give a hand to understand why or if I could avoid this.

moylop260:
@victormartinelicocorp 


The error [IntegrityError: column "categ_id" contains null values](https://travis-ci.org/OCA/project/jobs/206843658#L1033-L1038) means:
 - You are changing from not required to required a field, but there are old records with null values for this one

```python
    _inherit = 'project.task'
 
    categ_id = fields.Many2one(
        'project.category',
        'Category',
        required=True   # Add NOT NULL constraint for this field
    )
```

Note: Travis there are errors too but the log is not changing the state (because runbot use a regex to parser the log and travis use hard-coded blacklist... other MQT issue)
https://travis-ci.org/OCA/project/jobs/206843658#L1033-L1038

moylop260:
The database template (with modules dependencies installed before to install `project/` modules) has data demo.

Then if you try create a new field with a NOT NULL constraint you will be the error because the old records don't have this field filled.

screen shot 2017-03-01 at 11 33 21 pm

Using the command `"ALTER TABLE project_task ALTER COLUMN categ_id SET NOT NULL"`
the output is: `ERROR:  column "categ_id" contains null values`

You can fix it:
Option A: Using a `default` value for `categ_id` field.
Option B: Remove `required` parameter for `categ_id` field.

My vote is for option B, because add new required fields you will have errors from other actions/methods that auto-create task without consider `categ_id`

martintorresvm:
> The database template (with modules dependencies installed before to install project/ modules) has data demo.
Then if you try create a new field with a NOT NULL constraint you will be the error because the old records don't have this field filled.

@moylop260 That is a really good point. 

@elicoidal Should we have one default category for Tasks? Which one? 

I would suggest do:
On PR #224 finish improve `project_categ` module and make the `Root Category for Tasks` mandatory

class ProjectProject(orm.Model):
     _columns = {
         'task_categ_id': fields.many2one(
             'project.category',
             required=True,
             default=lambda self: self.env.ref('project.project_category_01'),  
             'Root Category for Tasks'),

Then add default value `project.category` from Odoo Standard demo data.

And here the project.task `categ_id` should be mandatory with default value form project.project `task_categ_id`.

moylop260:
This solution could add a major issue.
If the data is deleted then you will have a ugly traceback...

What about use required=False?

martintorresvm:
> This solution could add a major issue. If the data is deleted then you will have a ugly traceback... What about use required=False?

Understood but anyway we'd like to have an unique category linked to task independent from project and mandatory so I rather add default value just for Task and not modify project. Could be?

elicoidal:
I am  not sure to fully understand the issue. 
Categ_id field should be created by this module and is distinct from the tags. It should be specific for this module so we can do whatever we want with it and have it required.
This means that another non standard module has created this field, right?
@moylop260 @victormartinelicocorp

martintorresvm:
I think the problem is implicit with de demo data as it's used in runbot. 

Travis hasn't this issue.

Runbot first created DB with demo data with the standard Odoo modules installed.

Then over this DB try to install the modules within the PR.

So what it happens if you try to set a field as not null over a non empty table it's normal.

Many options are valid to solve this. One could be add default values 

@moylop260 @elicoidal

project_task_category/models/project_task_category.py
```diff
@@ -0,0 +1,15 @@
...
+from openerp import fields, models
+
+
+class ProjectTask(models.Model):
+    _inherit = 'project.task'
+
+    categ_id = fields.Many2one(
+        'project.category',
+        'Category',
+        required=True
```
dreispt: Why do we need the Category to be always mandatory?
I can see the case for it to be mandatory depending on the Project...
How about having it mandatory only on the Task Form? That way it will be easier to customize more specific use cases (and to have it installed on databases already having Task data).

moylop260:
Did you review my first comment?

If you see the log of travis you will see the same error
https://travis-ci.org/OCA/project/jobs/206843658#L1033-L1038

martintorresvm:
> Model link to be added

@elicoidal Where I should add model link?


@moylop260 @dreispt I think I have done the necessary please could you have a look into my latest commit?

dreispt (dismissed):
OK to merge with squash.
It would be nice to add a note to the README stating that the Category field is mandatory.

martintorresvm:
> OK to merge with squash.
It would be nice to add a note to the README stating that the Category field is mandatory.

@dreispt done

>   Model link to be added
@elicoidal Where I should add model link?

moylop260 (approved):

dreispt:
I just noticed that this is a new module.
But this feature is already implemented here: https://github.com/OCA/project/blob/8.0/project_categ/project_categ_model.py#L78

martintorresvm:
> I just noticed that this is a new module.
But this feature is already implemented here: https://github.com/OCA/project/blob/8.0/project_categ/project_categ_model.py#L78

@dreispt Apparently is the same feature.

But with `project_categ` only is possible to have one `task_categ_id` peer project and for all the task linked.

With this module is possible to have one `categ_id` peer task and not dependent from project.

cc @elicoidal

elicoidal:
@dreispt the idea of this module is to add a specific field Category which is not a tag (m2m).
The purpose of it is to be able to manage statistics or BI report based on this field, which we cannot do with the tags field.
With your module, as far as I can read, we have a related field which means all tasks of a project would have the same content for the category.
This is not the purpose of this module: allow different category for different tasks.

martintorresvm:
@dreispt Could you check my latest comment?

cc @elicoidal

dreispt:
@elicoidal @victormartinelicocorp Yes, you are correct, sorry for my confusion.

dreispt (approved):

dreispt:
@elicoidal You are the blocking element now :wink:

elicoidal (approved):

elicoidal:
hehe not anymore :p

PR #232, opened 2017-03-06
[MIG] project_task_materials: Migrated to 10.0

rigo1985:


project_task_materials/models/project.py
```diff
@@ -1,40 +1,31 @@
...
 
 class Task(models.Model):
     _inherit = "project.task"
-    material_ids = fields.One2many(
-        comodel_name='project.task.materials', inverse_name='task_id',
-        string='Materials used')
+
+    material_ids = fields.One2many(comodel_name='project.task.materials',
+                                   inverse_name='task_id',
+                                   string='Materials used')
```
dreispt, line 15: Please avoid code style only changes.
Both ways are PEP8 compliant, but when so IMO the original author's preference should be respected.

project_task_materials/models/project.py
```diff
@@ -1,40 +1,31 @@
...
-    task_id = fields.Many2one(
-        comodel_name='project.task', string='Task', ondelete='cascade',
-        required=True)
-    product_id = fields.Many2one(
-        comodel_name='product.product', string='Product', required=True)
+
+    task_id = fields.Many2one(comodel_name='project.task', required=True,
+                              string='Task', ondelete='cascade')
+    product_id = fields.Many2one(comodel_name='product.product',
+                                 string='Product', required=True)
```
dreispt, line 25: Same as above-

project_task_materials/views/project_view.xml
```diff
@@ -1,25 +1,24 @@
...
-              <field name="product_id"/>
-              <field name="quantity"/>
-            </tree>
-          </field>
+        <field name="name">project.task.form.materials</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.view_task_form2"/>
+        <field name="arch" type="xml">
+            <notebook position="inside">
+                <page string="Materials used" attrs="{'invisible': [('project_id', '=', False)]}">
```
dreispt, line 10: When completing the task, the user will record all time and materials spent.
IMO it is a better UI to have the Materials spend list in the same place as the work lines.

project_task_materials/models/project.py
```diff
@@ -1,40 +1,31 @@
...
     quantity = fields.Float(string='Quantity')
+    uom_id = fields.Many2one('product.uom', string='Unit of Measure')
+
```
dreispt, line 28: Good addition, but shouldn't it have a default value?

project_task_materials/views/project_view.xml
rigo1985: @dreispt  I think it's easier for the end user. I consider that later the possibility of importing the materials to be used that can be many should be added, so I add a new TAB. 
cc @Tecnativa

project_task_materials/models/project.py
rigo1985: @dreispt This field obtains the value of the unit of measure of the product.

dreispt: OK

project_task_materials/views/project_view.xml
dreispt: This is not my opinion, but I'm open to your view if I see other supporting it.

dreispt: I tried this PR on the OCA Runbot, and the material list was not editable.
I'm not sure if its a problem with Runbot or with the module - can you check?

pedrobaeza:
Superseeded by #251

PR #239, opened 2017-03-17
[8.0] [IMP] project_task_category

martintorresvm:
Add domain specific on categ_id
to let choose only the category with code 'categ_id'
	modified:   project_task_category/models/project_task_category.py


martintorresvm:
@elicoidal @dreispt Could you have a look?

elicoidal (approved):

project_task_category/models/project_task_category.py
```diff
@@ -11,4 +11,5 @@ class ProjectTask(models.Model):
     categ_id = fields.Many2one(
         'project.category',
         'Category',
+        domain=[('code', '=', 'categ_id')]
```
dreispt: Can we have domain listing the child Categories form a predefined root?
You can copy the implementation done for the `categ_ids` Tags, here: 
https://github.com/OCA/project/blob/8.0/project_categ/project_categ_model.py#L83

martintorresvm: @dreispt We only need one set of categories (that independent
if have or not parent_id or are assigned on the Task as root task category) 
are available just for the Task categ_id field.

And we think no need to have different sets for different Projects because this categories are very generic.

Don't you think so?

martintorresvm: @dreispt  Another thing is that I think it's really hard to manage even when start with a clean DB 

I already I tried to do as you suggest before I create this PR and I think we add too much complexity for a simple problem that could add lot of management work for users and later if made a mistake will not work as is expected.

Imagine I have a set of categories without parenting. 

Then I want to have one as a set for `Root Task category` (category available on categ_id  AKA: Category)
I start to add as others categories as child of this root.
 
Finish with that I start with another set for `Root Project category` (category available on categ_ids AKA Tags).
But here is not possible to know if I select one category already present on another root, right? 
and because one2many relation could disappear from another root.

IMO this is a big problem of management and much worst for live environments with multiple tasks and categories created.

dreispt: Maybe for your particular case that's true. You can't be sure that it's always true.
If you use Projects for very different domains, sucha as IT / HR / Fleet / Facility, etc., categories usable on each will be very different.

But I can add that myself in a later PR if you prefer.

martintorresvm: @dreispt I get your point of view but I'd rather have this filter domain and later if you want could be proposed as well.

Just for curiosity how could you categorize or distinguish between project domains? Seems very complex without another "Category level" just for projects

cc @elicoidal

dreispt: OK. Check the implementation on `project_categ`, doing precisely that for m2m Tags.

dreispt: @victormartinelicocorp Focusing on this particular change: I miss the point on the `domain`, and the `'categ_id'` hardcoded text. What is the point?

dreispt: OK, I had a deeper look.
I seems that the point is to have only a subset of Categories available for the m2o field.
Thinking about it, in general it makes sense that the Categories from the m2o and m2m are different, since they will fit different purposes in your implementation.

We could add some type of flag on Categories to make this distinction.
But we should also add the domain to the Tags field, but that can conflict with the `project_categ` module :-/
I suggest that you "copy" `project.categ` into a new model (`project.categ.main`?), using prototype inheritance:
```
class ProjectCategMain(models.Model):
    _name = `project.categ.main`
    _inherit = `project.categ`
```

This also makes easy to have different ACLs on each Category group: main Categories can be read only for regular users, and Tags Categories can be read-write, for example, to support dynamic reusable keywords.

martintorresvm: @dreispt Yes sorry by that I just like to use `code` field on for this purpose then might better just add a new boolean field just to check True/False

dreispt: I think we wrote cross posts :-)

martintorresvm: > I suggest that you "copy" project.categ into a new model (project.categ.main?), using prototype inheritance:
class ProjectCategMain(models.Model):
    _name = `project.categ.main`
    _inherit = `project.categ`
This also makes easy to have different ACLs on each Category group: main Categories can be read only for regular users, and Tags Categories can be read-write, for example, to support dynamic reusable keywords.

@dreispt Yes, is what we want to have from the very beginning when we start this matter with the BR modules (we had before our own model `task.category` instead of use `project.category`).

I'll follow your recommendation for a new model `project.category.main` inherited from `project.category`.

Thanks!

martintorresvm: > I think we wrote cross posts :-)

Yes, weird to reply a comment and find it comes after new one. Anyway thanks for helping :)

martintorresvm:
@elicoidal @dreispt Please let me know if could be merge.

elicoidal (approved):

project_task_category/models/project_task_category.py
dreispt: @victormartinelicocorp I slept on this, and I think that adding a flag to the Category model can do the trick. It does have have a few details to get right that I can share with you. Do you prefer to go this way?

martintorresvm: > victormartinelicocorp I slept on this, and I think that adding a flag to the Category model can do the trick. It does have have a few details to get right that I can share with you. Do you prefer to go this way?

@dreispt Actually I am ok with both solutions my latest commit I follow the way of create new model with inheritance.

If I have to choose I feel a little more right to have a separated DB tables so rather keep my latest commit: https://github.com/OCA/project/pull/239/commits/727c42e301719efbf945083dd51fe158cb466bd4

Please let me know if need to change anything and I'll do the necessary.

cc @elicoidal

dreispt (approved):
Looks great, thanks for bearing with me.

project_task_category/views/project_task_view.xml
```diff
@@ -46,5 +46,14 @@
                 </field>
             </field>
         </record>
+        <act_window id="action_type_category_project_task"
+            name="Project category"
```
seb-elico: @victormartinelicocorp AFAIK OCA guideline is to align attributes with the one from the first line, right?

seb-elico (changes requested):

martintorresvm:
@seb-elico I fix layout

@dreispt @elicoidal could be merged?

PR #241, opened 2017-03-17
[FIX][10.0] project_task_add_very_high: add OCA in authors list

astirpe:
I remember some time ago there was an automatic check for the mandatory OCA author in manifest.
It seems that check was removed...


andhit-r (approved):

pedrobaeza:
The check exists, but only as a warning, not an error, and nobody looks at warnings..

pedrobaeza (approved):

PR #242, opened 2017-03-23
10.0 mig project model to task

mourad-ehm:


project_model_to_task/models/project.py
```diff
@@ -22,7 +22,7 @@ def _authorised_models(self):
         """ Inherit this method to add more models depending of your
             modules dependencies
         """
-        models = self.env['ir.model'].search([('model', '!=', 'project.task')])
+        models = self.env['ir.model'].search([])
```
dreispt: Why are you removing the original domain?

project_model_to_task/models/project.py
```diff
@@ -75,33 +75,26 @@ def goto_document(self):
...
-            context = {}
+    def read(self, fields, load='_classic_read'):
 
         def update_context(action):
             action['context'] = safe_eval(action.get('context', '{}'))
             action['context'].update({
-                'from_model': context.get('active_model'),
-                'from_id': context.get('active_id'),
+                'from_model': self._context.get('active_model'),
+                'from_id': self._context.get('active_id'),
```
dreispt: Why not use the usual `self.env.context`?

project_model_to_task/models/project.py
```diff
@@ -75,33 +75,26 @@ def goto_document(self):
 class IrActionActWindows(models.Model):
     _inherit = 'ir.actions.act_window'
 
-    def read(self, cr, uid, ids, fields=None, context=None,
-             load='_classic_read'):
-        if context is None:
-            context = {}
+    def read(self, fields, load='_classic_read'):
 
         def update_context(action):
```
dreispt: It is good practice to declare all the used variables in the funtion signature.
Can use add `self` to it?:  `def update_context(self, action):`

project_model_to_task/models/project.py
mourad-ehm: I removed the original domain to allow the use of project.task as any other model. Thus we can link a task to an other task.

project_model_to_task/models/project.py
mourad-ehm: self._context return the same result as self.env.context. I do not know if there is a particular advantage of using either one or the other.

dreispt: OK. "_context" is a private attribute, not partt of the official API, and should should not be used directly.
Please prefer using the public API `self.env.context`.

mourad-ehm: OK thanks

bealdav:
Hi daniel could you have a look ?

bealdav:
@dreispt 
- 98.61 % of coverage: the hard way
- previously there were no test, no demo data in this module
- we now are at 74 % of coverage on this module
- 86 % for the whole branch is really acceptable don't you think ?

project_model_to_task/models/project.py
dreispt: ok

project_model_to_task/models/project.py
```diff
@@ -0,0 +1,105 @@
...
+            })
+            if 'params' in self.env.context and 'action':
+                action['context'].update({
+                    'from_action': self.env.context['params'].get('action')})
+            if 'params' in self.env.context and 'action':
+                action['context'].update({
+                    'from_action': self.env.context['params'].get('action')})
+        res = super(IrActionActWindows, self).read(
+            fields=fields, load=load)
+        if not self._context.get('install_mode'):
```
dreispt: s/self._context/self.env.context

project_model_to_task/models/project.py
```diff
@@ -0,0 +1,105 @@
...
+                'from_model': self.env.context.get('active_model'),
+                'from_id': self.env.context.get('active_id'),
+            })
+            if 'params' in self.env.context and 'action':
+                action['context'].update({
+                    'from_action': self.env.context['params'].get('action')})
+            if 'params' in self.env.context and 'action':
+                action['context'].update({
+                    'from_action': self.env.context['params'].get('action')})
+        res = super(IrActionActWindows, self).read(
```
dreispt: It would be nice to add an empty line here, to make it clear that the function declaration has ended

dreispt (approved):
A couple of nitpicks, but LGTM

mourad-ehm:
:+1:

PR #249, opened 2017-04-05
[REF] project_model_to_task

JayVora-SerpentCS:
- Updated Readme with right bug tracker and runbot URL
- Demo files should always be under demo tag of manifest.
- Used API for a functional field.

### DIFF
--- a/project_model_to_task/models/project.py
+++ b/project_model_to_task/models/project.py
@@ -10,14 +10,14 @@ from odoo.tools.safe_eval import safe_eval
 class ProjectTask(models.Model):
     _inherit = 'project.task'
 
+    @api.multi
+    @api.depends('model_reference')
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

project_model_to_task/models/project.py
```diff
@@ -10,14 +10,14 @@
 class ProjectTask(models.Model):
     _inherit = 'project.task'
 
+    @api.multi
```
bealdav, line 13: multi is the default mode in v10 (dixit A Lesuisse). Easier to read with few decorator

bealdav (commented):
Thanks a lot for these fix @JayVora-SerpentCS . 
Happy to see your support :+1:

dreispt (approved):

PR #251, opened 2017-04-11
[10.0][MIG] project_task_materials

chienandalu:
# Project Task Materials

Project Tasks allow to record the time spent on them, but some activities
often require you to keep a record of the material used also.

This module offers the ability to keep track of that material.

Note that only a simple record is made and no accounting or stock moves are
actually performed.

cc @Tecnativa


angelmoya:
I Tested on runbot and reviewed code :+1:

angelmoya:
this work was done on #232

pedrobaeza:
It was a try from one subcontracted guy, but this one is the valid one.

project_task_material/README.rst
```diff
@@ -0,0 +1,61 @@
...
+Usage
+=====
+
+To use this module:
+
+Every task form has a tab where product and their quantities can be added.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/140/9.0
```
dreispt: s/9.0/10.0

project_task_material/__manifest__.py
```diff
@@ -0,0 +1,25 @@
...
+    "version": "10.0.1.0.0",
+    "category": "Project Management",
+    "author": "Daniel Reis,"
+              "Tecnativa,"
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "product",
```
dreispt: `project` is missing

dreispt (commented):
Why I can't see the changes to the original module?
It seems that even unchanged lines are part of this diff.

chienandalu:
@dreispt Seems like I messed it up somehow with a rebase. I will try to fix it :/

chienandalu:
@dreispt The diff is already fixed

project_task_material/models/project.py
```diff
@@ -4,21 +4,21 @@
...
-        string='Materials used')
+        comodel_name='project.task.material', inverse_name='task_id',
+        string='Material used')
 
 
-class ProjectTaskMaterials(models.Model):
-    _name = "project.task.materials"
-    _description = "Task Materials Used"
+class ProjectTaskMaterial(models.Model):
+    _name = "project.task.material"
```
dreispt: If data table names are changed, we need a migration script.

project_task_materials
```diff
@@ -0,0 +1 @@
+project_task_material
```
dreispt: ?

project_task_material/models/project.py
pedrobaeza: Migration script added.

pedrobaeza (approved):
@dreispt anything else?

dreispt (approved):

PR #258, opened 2017-04-18
[FIX] project_issue_task: avoid button breaks form view

cubells:
When module project_issue_task is installed, button breaks form and fields decrease in size. 
Information on those fields can not be read.

cc @Tecnativa

### DIFF
--- a/project_issue_task/__openerp__.py
+++ b/project_issue_task/__openerp__.py
@@ -5,7 +5,7 @@
 {
     'name': 'Project Issue related Tasks',
     'summary': 'Use Tasks to support Issue resolution reports',
-    'version': '9.0.1.0.2',
+    'version': '9.0.1.1.0',
     'category': 'Project Management',
     'author': "Daniel Reis,"
               "Tecnativa, "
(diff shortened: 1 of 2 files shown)


pedrobaeza (changes requested):
Tried in runbot and it's ok, but please change version number

luismontalba (approved):
Tested on runbot

pedrobaeza:
Solve conflicts

project_issue_task/__openerp__.py
```diff
@@ -5,7 +5,7 @@
 {
     'name': 'Project Issue related Tasks',
     'summary': 'Use Tasks to support Issue resolution reports',
-    'version': '9.0.1.0.1',
+    'version': '9.0.1.0.2',
```
dreispt: Use `9.0.1.1.0`, because a server restart is not enough - a module upgrade is needed.

pedrobaeza (approved):

pedrobaeza:
Changed version number and merging with squash.

PR #260, opened 2017-04-19
8.0 project issue task sync

gfcapalbo:
Syncing Tasks and issues


project_categ/project_categ_model.py
```diff
@@ -32,7 +16,7 @@ class ProjectProject(orm.Model):
 class ProjectCategory(orm.Model):
     _inherit = 'project.category'
 
-    def _name_get(self, cr, uid, ids, context=None):
+    def name_get(self, cr, uid, ids, context=None):
         res = []
         rows = self.read(cr, uid, ids, ['name', 'parent_id'], context=context)
         for row in rows:
```
hbrunn: I don't believe this commit should be here

project_issue_task_sync/README.rst
```diff
@@ -0,0 +1,10 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+=======================
+Sync Projects and Tasks
+=======================
+
+This module was written to extend the functionality of ... to support ...
+and allow you to ...
```
hbrunn: use OCA's template and write something about this module

project_issue_task_sync/__openerp__.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Sync issues and tasks",
+    "version": "8.0.1.0.0",
+    "author": "Therp BV",
```
hbrunn: OCA is always an author too

project_issue_task_sync/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+    "license": "AGPL-3",
+    "category": "",
```
hbrunn: fill empty values

project_issue_task_sync/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+    "summary": "",
+    "depends": [
+        'project_issue'
+    ],
+    "data": [
+        'views/project_project.xml',
+        'views/project_task.xml',
+    ],
+    "pre_init_hook": False,
```
hbrunn: remove default values

project_issue_task_sync/models/project_issue.py
```diff
@@ -0,0 +1,51 @@
...
+from openerp import api, fields, models
+
+
+class ProjectIssue(models.Model):
+    _inherit = 'project.issue'
+
+
+    def get_changed_vals(self, task):
+        vals = {}
+        if len(task.issue_ids) ==1:
```
hbrunn: and if there are multiple issues? use falsyness here and deal with multiple issues below. per assumption all stages should be the same anyways, so it's safe to use the first issue

project_issue_task_sync/models/project_issue.py
```diff
@@ -0,0 +1,51 @@
...
+            task = this.sudo().task_id
+            """
+            if the create/write is called by a task creation skip all 
+            syncing functions and do not try to create a task
+            if it is not called vy a task creation 
+            and the project is "sync enabled" create the associated 
+            task.
+            """
+            if (this.project_id.sync_tasks_issues and not 
+                   self.env.context.get('sync_operation', False)):
```
hbrunn: invert this and `continue`

hbrunn: and remove the default that doesn't add anything

hbrunn: but i think this is logically flawed anyways: If I set `sync_operation` to `True`, I'd expect the sync to happen. Now I need to set it to `False`. Probably you want to remove the `not` and default to `True`?

project_issue_task_sync/models/project_issue.py
```diff
@@ -0,0 +1,51 @@
...
+                vals = this.get_changed_vals(task)
+                if vals:
+                   this.task.with_context(
+                       mail_notrack=True, sync_operation=True
+                   ).write(vals)
+
+
+    @api.returns('self', lambda value: value.id)
```
hbrunn: the `returns` decorator is propagated through inheritance. What you miss here is `api.model`

project_issue_task_sync/models/project_project.py
```diff
@@ -0,0 +1,50 @@
...
+from openerp import api, fields, models
+
+
+class project_project(models.Model):
+    _inherit = 'project.project'
+
+
+    sync_tasks_issues = fields.Boolean(
+       string='Sync Issues and tasks',
+       default='feasable_sync'
```
hbrunn: you assign a string as default value for a boolean. Works, but only because any nonempty string is truthy. Why not `True`?

hbrunn: ah, you think this calls the function. No, it doesn't. You want a lambda expression here

project_issue_task_sync/models/project_project.py
```diff
@@ -0,0 +1,50 @@
...
+        tasks and  issues will remain disconected.
+        
+        By triggering an empty write on every task, that will start update of
+        all related issues and possible creation of missing issues.
+
+        """
+        if self.sync_tasks_issues:
+            for task in self.env['project.task'].search(
+                   [('project_id', '=', self.id)]):
+                task.write({})
```
hbrunn: first, this doesn't work (onchange functions are called in an onchange context, which causes all writes to operate on dummy objects, and then what was written on the one record that was changed is actually committed to the database), and second even if it worked, I don't think we should want this. Remove this entirely, and if you feel we need this, provide an action on tasks and issues (or just on projects), that walks through the selected tasks/issues (or the tasks/issues of the selected projects) and does the sync

project_issue_task_sync/models/project_project.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+    #this can run in a wizard to sync all projects
```
hbrunn: see above. I'd rather make this a `multi` function that operates on a set of projects

project_issue_task_sync/models/project_project.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def sync_all_projects(self):
+        for project in self.env.filtered('sync_tasks_issues'):
+            project.sync_issues_for_tasks()
```
hbrunn: calling it here works. so keep the function, but also make it operate on issues

project_issue_task_sync/models/project_project.py
```diff
@@ -0,0 +1,50 @@
...
+
+        if we remove the syncing nothing happens. 
+        tasks and  issues will remain disconected.
+        
+        By triggering an empty write on every task, that will start update of
+        all related issues and possible creation of missing issues.
+
+        """
+        if self.sync_tasks_issues:
+            for task in self.env['project.task'].search(
```
hbrunn: why would you trigger another expensive search operation when you already have the ids in https://github.com/OCA/OCB/blob/8.0/addons/project/project.py#L262 ?

project_issue_task_sync/models/project_task.py
```diff
@@ -0,0 +1,62 @@
...
+
+    issue_ids = fields.One2many('project.issue', 'task_id', 'Issues')
+
+    # CAN EXTEND THIS TO SYNC MORE FILES, if we want to create a model
+    # fields to sync we can do it here.
+
+
+    def get_changed_vals(self):
+        vals = {}
+        if self.issue_ids.stage_id != self.stage_id:
```
hbrunn: protect against multiple issues: `[:1]`, as above, we should be able to assume those to have the same value anyways

project_issue_task_sync/models/project_task.py
```diff
@@ -0,0 +1,62 @@
...
+            vals['stage_id'] = self.stage_id.id
+        if self.issue_ids.user_id != self.user_id:
+            vals['user_id'] = self.user_id.id
+        return vals
+
+    @api.multi
+    def set_issue_vals(self):
+        for this in self:
+            if len(this.issue_ids) > 1:
```
hbrunn: see above

project_issue_task_sync/models/project_task.py
```diff
@@ -0,0 +1,62 @@
...
+                continue
+            # If I see that on write or create my task has no issue , create
+            # it obviously as a sync_operation / no mail
+            if len(this.issue_ids) == 0:
```
hbrunn: use falsyness

project_issue_task_sync/models/project_task.py
```diff
@@ -0,0 +1,62 @@
...
+               self.env['project.issue'].with_context(
```
hbrunn: shouldn't this be under the scope of the conditional below too?

project_issue_task_sync/models/project_task.py
```diff
@@ -0,0 +1,62 @@
...
+                     'project_id':  self.project_id.id,
+                     'name': self.name,
+                     'task_id':  self.id,
+                     'user_id': self.user_id.id,
+                     'stage_id': self.stage_id.id,
+                     'description': self.description
+                   }
+                )
+            if (this.project_id.sync_tasks_issues and not
+                    self.env.context.get('sync_operation', False)):
```
hbrunn: see above

project_issue_task_sync/static/src/js/trp_issue_task_sync.js
```diff
@@ -0,0 +1,7 @@
+//-*- coding: utf-8 -*-
+//© 2017 Therp BV <http://therp.nl>
+//License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+openerp.trp_issue_task_sync = function(instance)
+{
+}
```
hbrunn: remove pointless file

hbrunn (changes requested):
check CI status, especially flake8, and tests are mandatory.

Oh, and no Therp logo in OCA repos

And you should `suspend_security` your create and write calls, because an issue user might not have the permissions to create a task and the other way around.

project_issue_task_sync/models/project_issue.py
gfcapalbo: no, it has been tested. this means: "this write is being done not by the user but is the result of a sync operation already"   there fore  don't do the sync operation again. it's a structure where issue and task call sync eachother. on write they are written and sync the other model with sync_operation =True so that the other model's write won't sync back again. works perfect.

project_issue_task_sync/models/project_task.py
gfcapalbo: no , because the bottom is what to be done if we are NOT in a sync operation. 
it is  possible that we are in a sync operation and therefore we need to create a new issue.
changing the name from sync_operation to is_sync_operation for readability

project_issue_task_sync/models/project_issue.py
hbrunn: then the name is very misleading, better choose something that expresses what you want

gfcapalbo: changed to is_sync_operation

hbrunn (changes requested):
nearly. But you really need to remove commit 1d6cfda61a96f519c64354f3c66d6a2abd424de5 from you branch, now you added a commit that removes the code, but that's not the point here. Then delete the empty css file too, replace the therp logo with the oca one and this should be okay

PR #262, opened 2017-04-20
add module project_issue_task_sync

gfcapalbo:
Syncs (2 way) tasks and issues for selected projects.

### DIFF
--- a/project_issue_task_sync/README.rst
+++ b/project_issue_task_sync/README.rst
@@ -1,0 +1,35 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=====================
+Sync Issues and Tasks
+=====================
+
+This module syncs assigned user and state from task to issue and viceversa.
+If a task is created and does not have a issue a related issue will be
+generated. Projects may or may not have this syncing function, the activation
+of sync between issues and task may be activated on the project form by
+selecting the checkbox "sync issues and tasks". By default projects that use
+tasks and issues will be synced.
+
+
+Contributors
+------------
+
+* Giovanni Francesco Capalbo  <giovanni@therp.nl>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/project_issue_task_sync/models/project_issue.py
+++ b/project_issue_task_sync/models/project_issue.py
@@ -67,8 +67,10 @@ class ProjectIssue(models.Model):
         #project
         sync = self.project_id.sync_tasks_issues
         result = super(ProjectIssue, self).write(vals)
-        if sync:
-            self.set_binded_task_vals()
+        for this in self:
+            sync = this.project_id.sync_tasks_issues
+            if sync:
+                this.set_binded_task_vals()
         return result
 
     @api.multi
--- a/project_issue_task_sync/models/project_issue.py
+++ b/project_issue_task_sync/models/project_issue.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api, models, fields
+
+
+class ProjectIssue(models.Model):
+    _inherit = 'project.issue'
+
+    sync_tasks_issues = fields.Boolean(
+        related=['project_id', 'sync_tasks_issues']
+    )
+
+    def get_changed_vals(self, origin, paired):
+        vals = {}
+        if paired.stage_id != origin.stage_id:
+            vals['stage_id'] = origin.stage_id.id
+        if paired.user_id != origin.user_id:
+            vals['user_id'] = origin.user_id.id
+        if paired.name != origin.name:
+            vals['name'] = origin.name
+        if paired.description != origin.description:
+            vals['description'] = origin.description
+        if paired.project_id != origin.project_id:
+            vals['project_id'] = origin.project_id.id
+        return vals
+
+    def get_or_create_binded_task(self):
+        if self.task_id:
+            return self.task_id
+        task = self.env['project.task'].with_context(
+            mail_notrack=True, is_sync_operation=True).create({
+                'project_id':  self.project_id.id,
+                'name': self.name,
+                'issue_ids': [(4, self.id)],
+                'user_id': self.user_id.id,
+                'stage_id': self.stage_id.id,
+                'description': self.description or 'No Description'
+            })
+        return task
+
+    def set_binded_task_vals(self):
+        for this in self:
+            if not this.env.context.get('is_sync_operation'):
+                task = this.task_id
+                if not len(task) > 0:
+                    task = this.get_or_create_binded_task()
+                vals = this.get_changed_vals(this, task)
+                task.with_context(
+                    mail_notrack=True, is_sync_operation=True
+                ).write(vals)
+
+    @api.model
+    @api.returns('self', lambda value: value.id)
+    def create(self, vals):
+        this = super(ProjectIssue, self).create(vals)
+        if this.project_id.sync_tasks_issues:
+            this.get_or_create_binded_task()
+        return this
+
+    @api.multi
+    def write(self, vals):
+        result = super(ProjectIssue, self).write(vals)
+        if self.project_id.sync_tasks_issues:
+            self.set_binded_task_vals()
+        return result
+
+    @api.multi
+    def unlink(self):
+        result = False
+        for this in self:
+            if this.project_id.sync_tasks_issues:
+                if not self.env.context.get('is_sync_operation'):
+                    # no sudo, we suppose that if you have decided to sync
+                    # issues and tasks and you have delete privledges on one,
+                    # you also have on the other.
+                    this.with_context(is_sync_operation=True).task_id.unlink()
+            result = super(ProjectIssue, this).unlink()
+        return result
--- a/project_issue_task_sync/models/project_issue.py
+++ b/project_issue_task_sync/models/project_issue.py
@@ -53,15 +53,21 @@ class ProjectIssue(models.Model):
     @api.model
     @api.returns('self', lambda value: value.id)
     def create(self, vals):
+        sync = self.project_id.sync_tasks_issues
         this = super(ProjectIssue, self).create(vals)
-        if this.project_id.sync_tasks_issues:
+        if sync:
             this.get_or_create_binded_task()
         return this
 
     @api.multi
     def write(self, vals):
+        #if i change to a project that is not synced it will not trigger the
+        #sync so i need to fetch the project BEFORE the write, in order to have
+        #one last sync (example have the task synced with the new unsynced
+        #project
+        sync = self.project_id.sync_tasks_issues
         result = super(ProjectIssue, self).write(vals)
-        if self.project_id.sync_tasks_issues:
+        if sync:
             self.set_binded_task_vals()
         return result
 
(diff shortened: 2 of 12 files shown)


project_issue_task_sync/README.rst
```diff
@@ -0,0 +1,35 @@
...
+=====================
+Sync Issues and Tasks
+=====================
+
+This module syncs assigned user and state from task to issue and viceversa.
+If a task is created and does not have a issue a related issue will be
+generated. Projects may or may not have this syncing function, the activation
+of sync between issues and task may be activated on the project form by
+selecting the checkbox "sync issues and tasks". By default projects that use
+tasks and issues will be synced.
```
dreispt, line 14: (Functional Q) I'm puzzled why we need to always have Issue/Task pairs.
For example, I have used a workflow like this:
- Create Issue
- Triage Issue: close it or create a Task
- Do the Task, and keep it in sync with the origin issue

In this case you can have Issues without Tasks.
So I think that requiring the Issue - Task pair to always be true is too specific of a particular use case.
What do you think?

project_issue_task_sync/models/project_project.py
```diff
@@ -0,0 +1,35 @@
...
+        update the issue too (and viceversa)
+        if we remove the syncing nothing happens.
+        tasks and  issues will remain disconected.
+        By triggering an empty write on every task, that will start update of
+        all related issues and possible creation of missing issues.
+
+        """
+        if self.sync_tasks_issues:
+            for task in self.task_ids:
+                task.write({})
```
dreispt: This works, but it would be more explicit to say `task.set_issue_vals()`

project_issue_task_sync/models/project_task.py
```diff
@@ -0,0 +1,58 @@
...
+        if self.issue_ids[:1].user_id != self.user_id:
+            vals['user_id'] = self.user_id.id
+        return vals
+
+    @api.multi
+    def set_issue_vals(self):
+        for this in self:
+            # If I see that on write or create my task has no issue , create
+            # it obviously as a sync_operation / no mail
+            if not len(this.issue_ids) > 0:
```
dreispt: You can use the simpler `if not this.issue_ids:`

project_issue_task_sync/models/project_task.py
```diff
@@ -0,0 +1,58 @@
...
+                self.env['project.issue'].with_context(
+                    mail_notrack=True, is_sync_operation=True).create({
```
dreispt: As mentioned before, I don't feel that automatically creating an Issue for a Task is a generic enough requirement.
I would rather not doing it, or having it as an optional behaviour.

project_issue_task_sync/models/project_task.py
```diff
@@ -0,0 +1,58 @@
...
+                        'project_id':  self.project_id.id,
+                        'name': self.name,
+                        'task_id':  self.id,
+                        'user_id': self.user_id.id,
+                        'stage_id': self.stage_id.id,
+                        'description': self.description
+                        })
```
dreispt: You could isolate this dict preparation an a separate method, for extensibility.

project_issue_task_sync/models/project_task.py
```diff
@@ -0,0 +1,58 @@
...
+    # CAN EXTEND THIS TO SYNC MORE Fields, if we want to create a model
+    # fields to sync we can do it here.
+
+    def get_changed_vals(self):
+        vals = {}
+        if self.issue_ids[:1].stage_id != self.stage_id:
+            vals['stage_id'] = self.stage_id.id
+        if self.issue_ids[:1].user_id != self.user_id:
+            vals['user_id'] = self.user_id.id
+        return vals
```
dreispt: You don't really need this method, it is a duplicate of what you have in Project Issues.

project_issue_task_sync/models/project_issue.py
```diff
@@ -0,0 +1,41 @@
...
+        for this in self:
+            task = this.sudo().task_id[:1]
+            # TODO task creation? for now if there is no task just skip.
+            if (this.project_id.sync_tasks_issues and task and not
+                    self.env.context.get('is_sync_operation')):
+                vals = this.get_changed_vals(task)
+                if vals:
+                    this.task.with_context(
+                        mail_notrack=True, is_sync_operation=True
+                    ).write(vals)
```
dreispt: When different, the values on the Issue overwrite the values on the Task.
This is not obvious, and can even be seen as unexpected:
In the worklflow where an Issue is reported, and then a Task is created for it's resolution, we would expect the Task stage and Responsible to change, and have the Issue sync with it.
AFAICS, with this code, if I try to change the Task Stage, the change would be automatically reverted!

dreispt (changes requested):

dreispt:
Hey @gfcapalbo I have concerns about the functional side of this PR. Maybe we're assuming too much? It's best to discuss that first.

project_issue_task_sync/README.rst
gfcapalbo: well your case is well served by full syncing too.  

Create issue (task created automatically)
triage issue: 
          a) close it (task closes too)\
          b) create a task (not needed , task is already there, just do whatever on the task, that is allways synced with the issue.

i think we save also a passage in the workflow by no having to create a task.
True, there is some specificity to this module, it is used internally, but overall it spawned from the idea that  issue and tasks are quite redundant logically, it seems to me like these to models are the result of an accidental duplication of effort. the whole aim of this module is to tidy that out.  there fore the full syncing. there is allways the option to unsync a project and do whatever you want.

also added to views:   
      On a synced project  tasks are readonly on the issue form and issues are read only on the task form.

project_issue_task_sync/models/project_task.py
gfcapalbo: it is optional in the sense that we can choose not to sync the project project_id.sync_tasks_issues == False

gfcapalbo:
@dreispt  thanks for the review  i am now refactoring and fixing according to your notes.  the automatic creation is one point i think should remain , as i explained in my comments above.

Also cleaning up all those commits. I believe the branch was accidentally made off of a 10.0

hbrunn (approved):

project_issue_task_sync/README.rst
dreispt: >  issue and tasks are quite redundant logically, it seems to me like these to models are the result of an accidental duplication of effort

I see your pain, and this is worth discussing.
I've also been having this debate for years now: should Issues and Tasks just be the same Model?

I first started with the Open Issue -> Create Task workflow.
It looked complex to users, so on a second phase I dropped Issues and used only Tasks, and just added triaging as Stages.
Now it seems that it is an oversimplifications and I'm feeling the need to go back to the first form: Issues -> Tasks.

I see Issues as customer/frontend facing, and Tasks as delivery/backend facing.
Issues are for Customers and 1st level Service Desk interaction.
Tasks are for 2nd level Service Desk interaction, with the actual people/teams performing the necessary work.
One Issue may need many Tasks to be performed, possibly by different people.
For example, a "Create New User"  Issue would mean creating Tasks for the Sysadmin to create accounts, for HR to create access card, for IT to provide PC and phone, etc.

Keeping Issues a separate module makes it easier to manage access control, and keep internal messages on Tasks away from the sight of the end Customer.

I'm still thinking if the automatic Task creation improves the process. 
After all you always need to assign the work to someone, and it is about the same effort as the "Create Task" step.
And I'm not sure this behaves properly when you have multiple Tasks for the same Issue.
Perhaps having a "root Task" with "subTasks" can answer that?

What do you think?

hbrunn (changes requested):
CI is still failing

project_issue_task_sync/README.rst
gfcapalbo: Interesting historical perspective! thanks!
if you consider it like that, where an issue has multiple subtasks assigned to different people, this module  as it is does not make sense. The module is useful for projects where the project itself functions as "root task" and the "subtasks" are synced issues/tasks. It should be used only by users who have need for this usecase.
I can explain this better in the readme. 
i think that the root / substask structure could  work, where the root task is synced with the issue, but it is out of the scope for this idea.
In our usecase the issues and it's discussion are client facing. the discussions are not synced (sync_operations are not with messages). the tasks are work for the developers/consultants in the project.

in conclusion, yes "root task/ sub tasks" would accomplish the trick of syncing + having multiple subtasks for different users, but  we would be adding usecases and complexity: for example if i set the issue to "done", the root task should be synced to "done" even if it's subtasks are not completed? All these things should be decided and clear to the user before.
I think the syncing should be kept for what it is, an option for project managers to keep some of their projects with : (stage, assign to, name, description, project_id )   synced two ways. nothing more. 
I agree with how tasks an issues should work: tasks for backend and issues for frontend/helpdesk.

gfcapalbo:
@hbrunn @dreispt  passing tests except for odoo specific false positives of codacy, and "could be function" i would like to keep connected to the model and a generic "except" always from codacy.

project_issue_task_sync/models/project_issue.py
```diff
@@ -0,0 +1,79 @@
...
+        if paired.name != origin.name:
+            vals['name'] = origin.name
+        if paired.description != origin.description:
+            vals['description'] = origin.description
+        if paired.project_id != origin.project_id:
+            vals['project_id'] = origin.project_id.id
+        return vals
+
+    def get_or_create_binded_task(self):
+        if self.task_id:
```
dreispt, line 29: api.ensure_one()

project_issue_task_sync/models/project_issue.py
```diff
@@ -0,0 +1,79 @@
...
+        this = super(ProjectIssue, self).create(vals)
+        if this.project_id.sync_tasks_issues:
+            this.get_or_create_binded_task()
+        return this
+
+    @api.multi
+    def write(self, vals):
+        result = super(ProjectIssue, self).write(vals)
+        if self.project_id.sync_tasks_issues:
+            self.set_binded_task_vals()
```
dreispt: I could be worng, but AFAIK `self` is the state before the write operation.
It's `project_id` value can be outdated if the write operation changes it.
So IMO you should use `result.` instead of `self.` in these two statements.

project_issue_task_sync/models/project_project.py
```diff
@@ -0,0 +1,19 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api, fields, models
+
+
+class ProjectProject(models.Model):
+    _inherit = 'project.project'
+
+    sync_tasks_issues = fields.Boolean(
+        string='Sync Issues and tasks',
+        default=lambda x: x.use_issues and x.use_tasks
```
dreispt, line 12: I think its best to have this feature disabled by default.

project_issue_task_sync/models/project_task.py
```diff
@@ -0,0 +1,77 @@
...
+from openerp import api, fields, models
+
+
+class ProjectTask(models.Model):
+    _inherit = 'project.task'
+
+    sync_tasks_issues = fields.Boolean(
+        related=['project_id', 'sync_tasks_issues']
+    )
+    issue_ids = fields.One2many('project.issue', 'task_id', 'Issues')
```
dreispt, line 13: AFAICT the base use case is 1 Issue -> 1 Task.
If we want to expand it, IMO it makes more sense to have 1 Issue -> n Tasks, rather than 1 Task -> n Issues...
Could you comment on this?

project_issue_task_sync/models/project_task.py
```diff
@@ -0,0 +1,77 @@
...
+                        mail_notrack=True, is_sync_operation=True
+                    ).write(vals)
+
+    @api.model
+    @api.returns('self', lambda value: value.id)
+    def create(self, vals):
+        this = super(ProjectTask, self).create(vals)
+        if this.project_id.sync_tasks_issues:
+            this.get_or_create_binded_issue()
+        return this
```
dreispt, line 49: Does it make sense to have a two way automatic creation?
I can see the point to automatically create a Task for each new Issue, but I will probably prefer to not automatically create an Issue for new Tasks. That should be an additional flag on the Project, or perhaps a button/"More" option on Tasks.
Comments?

dreispt (commented):
I have concerns about the Task -> Issue syncing. 
Check my comments.

project_issue_task_sync/models/project_issue.py
gfcapalbo: write returns a boolean.

project_issue_task_sync/models/project_task.py
hbrunn: this is the inverse of the already existing core field: https://github.com/OCA/OCB/blob/8.0/addons/project_issue/project_issue.py#L252 - in practice, we use this as a 1:1 relation

PR #264, opened 2017-04-28
[9.0][ADD]Project WBS

AaronHForgeFlow:
Project Work Breakdown Structure
================================

* The hierarchy of a project is considered the WBS
  (Work Breakdown Structure)
* The analytic accounts in the project hierarchies are considered WBS
  components
* The analytic account code is shown in the project
* The complete WBS path code  is shown in the analytic account and in the
  project
* The complete WBS path name is shown in the analytic account and in the
  project
* The WBS paths are concatenated with each other
* It is possible to search projects by complete WBS path code & name
* It is possible to search tasks by project complete WBS path code & name
* The WBS components can be classified as project, phase, deliverable, work package.
* The classification is shown in the project and analytic account views

There's an opened issue regarding this: https://github.com/OCA/account-analytic/issues/65


AaronHForgeFlow:
cc @eficent @sysadminmatmoz

andhit-r:
Hi @aheficent , i'm curious how project.task will fit with this module?

sysadminmatmoz:
Hi @andhit-r .
The project task is just the lowest level WBS element (PMBOK, ISO21500, PRINCE2) that cannot be further broke down. The task is meant for an individual while a project level WBS is meant for a team. The lowest project class in the WBS is the work package, which is meant as a collection of tasks. So a project manager (WBS element of the class project) breaks down the project in phases (time based WBS) or deliverables (product based WBS), the deliverable (or phase) manager breaks down his project (class deliverable or phase) in work packages (for teams), the work package manager (or team leader) breaks down its work package to tasks assigned to the single members of its team.
This is the traditional workflow.

andhit-r:
Hi @sysadminmatmoz, thanks for your explanation. I still can create task on project, phase, or deliverable. According to your explanation a task can only be created on work packages level? Is that by design?

sysadminmatmoz:
Yes. Although the normal PMBOK workflow expects tasks only in work packages, we left the possibility to create them on any project class to keep Odoo flexible for any use case.
At the end its not that important to limit tasks only to a single class since the main purpose stays anyway: the project (of any class) is an analytic account, a container for the resources, costs and activities (tasks) - if a project is small and it doesn't require breaking down, we can still keep it just as a single WBS element of class project and hang tasks directly on it - at the end not every project requires low level WBS. That way we can still keep the usual Odoo's project workflow for some projects.

project_wbs/__openerp__.py
```diff
@@ -0,0 +1,31 @@
...
+{
+    'name': 'Work Breakdown Structure',
+    'version': '9.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'Matmoz d.o.o., '
+              'Luxim d.o.o., '
+              'Deneroteam, '
+              'Eficent, '
+              'Odoo Community Association (OCA)',
+    'website': 'https://www.github.com/OCA/project.git',
```
dreispt: Remove the `.git` to get a working Web URL

project_wbs/__openerp__.py
```diff
@@ -0,0 +1,31 @@
...
+    'depends': [
+        'project',
+        'account',
+        'analytic',
+        'account_analytic_parent',
+        'project_issue',
```
dreispt: Can we remove the `account` and `project_issue` dependencies?
Also, `project` already depends on `analytic`, so the later can also be removed.

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,228 @@
...
+                    data.insert(0, acc.code)
+
+                acc = acc.parent_id
+            if data:
+                if len(data) >= 2:
+                    data = '/'.join(data)
+                else:
+                    data = data[0]
+                data = '[' + data + '] '
+            account.complete_wbs_code = data or ''
```
dreispt: I'm sure this can be made simpler:
```
for analytic in self:
    if analytic.parent_id:
        analytic.complete_wbs_code = '/'.join(
            analytic.parent_id.complete_wbs_code,
            analytic.code)
```

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,228 @@
...
+                if acc.name:
+                    data.insert(0, acc.name)
+
+                acc = acc.parent_id
+            if data:
+                if len(data) >= 2:
+                    data = '/'.join(data)
+                else:
+                    data = data[0]
+            account.complete_wbs_name = data or ''
```
dreispt: Same as above.
Use spaces around the ` / `?

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,228 @@
...
+            if data:
+                if len(data) >= 2:
+                    data = ''.join(data)
+                else:
+                    data = data[0]
+            account.wbs_indent = data or ''
+
+    @api.multi
+    @api.depends('account_class')
+    def _get_project_account_id(self):
```
dreispt: It's best to use `analytic`when referring to Analytical Accounts, to avoid confusion with Accounting Accounts.

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,228 @@
...
+    )
+    complete_wbs_name = fields.Char(
+        compute=_complete_wbs_name_calc,
+        string='Full WBS path',
+        help='Full path in the WBS hierarchy',
+        store=True
+    )
+    project_analytic_account_id = fields.Many2one(
+        'account.analytic.account',
+        compute=_get_project_account_id,
```
dreispt: `=_get_project_analytic_id`

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,228 @@
...
+                )] + args, limit=limit
+            )
+            accountbyname = self.search(
+                [('complete_wbs_name', 'ilike', '%%%s%%' % name)] + args,
+                limit=limit
+            )
+            account = accountbycode + accountbyname
+        else:
+            account = self.search(args, limit=limit)
+        return account.name_get()
```
dreispt: This overwrites extensions. 
Must use super().

project_wbs/models/project_project.py
```diff
@@ -0,0 +1,347 @@
...
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+import time
+from datetime import datetime, date
+from openerp import api, fields, models, _
+
+
+class Project(models.Model):
+    _name = "project.project"
```
dreispt: Not necessary - you may remove it.

dreispt (commented):
Partial review.

AaronHForgeFlow:
Hi @dreispt Thank you for your review. I'll attend to any single comment. Regards

project_wbs/models/account_analytic_account.py
AaronHForgeFlow: Tried to modify this as you said but the result was not the same. This is still pending  to be fixed.

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,231 @@
...
+            data = []
+            acc = account
+            while acc:
+                if acc.code:
+                    data.insert(0, acc.code)
+
+                acc = acc.parent_id
+            if data:
+                if len(data) >= 2:
+                    data = ' / '.join(data)
```
dreispt: I know that I suggesting the surrounding spaces, but in the particular case of the Code I think it's best to not have them.

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,231 @@
...
+             'within the project WBS hierarchy',
+        store=True
+    )
+    complete_wbs_name = fields.Char(
+        compute=_complete_wbs_name_calc,
+        string='Full WBS path',
+        help='Full path in the WBS hierarchy',
+        store=True
+    )
+    project_analytic_account_id = fields.Many2one(
```
dreispt: It would be OK to just use `project_analytic_id`, but not a problem.

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,231 @@
...
+        'account.analytic.account',
+        compute=_get_project_analytic_id,
+        string='Analytic Account',
+        store=True
+    )
+    user_id = fields.Many2one(
+        'res.users', 'Project Manager', track_visibility='onchange',
+        default=_default_user)
+    manager_id = fields.Many2one('res.users', 'Account Manager',
```
dreispt: Can we have just "Manager" instead of "Account Manager"?

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,231 @@
...
+         ('work_package', 'Work Package')], 'Class',
+        help='The classification allows you to create a proper project '
+             'Work Breakdown Structure'
+    )
+    parent_id = fields.Many2one(default=_default_parent)
+    partner_id = fields.Many2one(default=_default_partner)
+
+    @api.model
+    def name_search(self, name, args=None, operator='ilike', limit=100):
+        if not args:
```
dreispt: AFAICR the `name_search` already works against whatever `name_get` returns.
So probably this is not needed.
Can you confirm?

project_wbs/models/project_project.py
```diff
@@ -0,0 +1,342 @@
...
+        store=True
+    )
+
+    members = fields.Many2many('res.users', 'project_user_rel', 'project_id',
+                               'uid', 'Project Members', help="""Project's
+                               members are users who can have an access to
+                               the tasks related to this project.""",
+                               default=_get_parent_members,
+                               states={'close': [('readonly', True)],
+                                       'cancelled': [('readonly', True)]})
```
dreispt: I'm +1 for this "members" feature, but I think it shouldn't be bundled here - it should go into it's own module.

project_wbs/models/project_project.py
```diff
@@ -0,0 +1,342 @@
...
+from openerp import api, fields, models, _
+
+
+class Project(models.Model):
+    _inherit = "project.project"
+    _description = "WBS element"
+    _order = "c_wbs_code"
+
+    @api.multi
+    def _get_project_analytic_wbs(self):
```
dreispt, line 20: AFAICS the recursive Name, Code, search, etc are provided by the implicit Analytic Account.
So I feel that the corresponding code should not be in the Project.
Can you confirm this?

AaronHForgeFlow:
Thank you for your review @dreispt I'll attend to everything

project_wbs/tests/test_project_wbs.py
```diff
@@ -0,0 +1,43 @@
...
+             'parent_id': self.parent_account.id})
+        self.son_account = self.project_son.analytic_account_id
+        self.project_grand_son = self.env['project.project'].create(
+            {'name': 'Test project grand son',
+             'code': '02',
+             'parent_id': self.son_account.id})
+        self.grand_son_account = self.project_grand_son.analytic_account_id
+
+    def test_get_child_accounts(self):
+        accounts = self.project._get_project_wbs()
```
MiquelRForgeFlow, line 30: where is defined the method `_get_project_wbs`?

MiquelRForgeFlow (changes requested):

project_wbs/tests/test_project_wbs.py
AaronHForgeFlow: I removed it because it was not needed for functional purposes. As long as it will needed for other modules on top opf this one and also as long as it is useful for reporting purposes I put this method back. :neutral_face:

project_wbs/models/project_project.py
sysadminmatmoz: c_wbs_code is a composed code that reflects the position of the project inside the work breakdown hierarchy and it is used mostly to sort the project list regarding the wbs hierarchy. The project list's sorting order would be random hierarchy wise without it since the hierarchy is not static as with the accounts, but dynamic - a project can be moved to another parent in time and that way moved to another spot in the hierarchy order.

dreispt:
Functional comments:

* I don't see the standard The "Project / Dashboard" standard menu - was it removed?
* When using "Create WBS Child element" button, the "Use tasks" and "Use issues" should default to the parent's, and the "Class" should default to "Work Package"
* When using "Create WBS Child element" button, in the new empty form the WBS child list shows all projects (probably caused by a domain returning no records, so all are shown). I suggest hiding it at this point.
* The "Class" should be mandatory and in general default to "Project".
* In the Project form, the "Project Stages" tab is shown twice.
* The "WBS code" should be without square brackets. Add them on the "name_get" only.
* Its not possible to make an existing "Project" record a child of another existing  "Project" record.

project_wbs/README.rst
```diff
@@ -0,0 +1,75 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+Project Work Breakdown Structure
```
elicoidal: add a line before

project_wbs/README.rst
```diff
@@ -0,0 +1,75 @@
...
+================================
+
+* The hierarchy of a project is considered the WBS
+  (Work Breakdown Structure)
+* The analytic accounts in the project hierarchies are considered WBS
+  components
+* The analytic account code is shown in the project
+* The complete WBS path code  is shown in the analytic account and in the
```
elicoidal: remove extra space

project_wbs/README.rst
```diff
@@ -0,0 +1,75 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+Project Work Breakdown Structure
+================================
+
+* The hierarchy of a project is considered the WBS
```
elicoidal: This list would be easier to read if split in 2 or 3 main topics:
1. Adding WBS information to Odoo Projects
2. Searching and Browsing WBS

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,205 @@
...
+        for account in self:
+            data = []
+            acc = account
+            while acc:
+                if acc.name:
+                    data.insert(0, acc.name)
+                acc = acc.parent_id
+            if data:
+                if len(data) >= 2:
+                    data = ' / '.join(data)
```
elicoidal, line 68: here you have spaces (' / ') but [here](https://github.com/OCA/project/pull/264/files#diff-600b527f4536b12326fea3aed304bc48R181) you don't.
Can you harmonize?

elicoidal (changes requested):

elicoidal:
@aheficent during the tests in runbot, I have this traceback (non-blocking but annoying):
How to reproduce: create a new Project/phase/deliverable and click in the name field:

  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 885, in call_kw
    return self._call_kw(model, method, args, kwargs)
  File "/home/odoo/odoo-10.0/addons/web/controllers/main.py", line 877, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 681, in call_kw
    return call_kw_multi(method, model, args, kwargs)
  File "/.repo_requirements/odoo/odoo/api.py", line 672, in call_kw_multi
    result = method(recs, *args, **kwargs)
  File "/.repo_requirements/odoo/odoo/models.py", line 5458, in onchange
    record = self.new(values)
  File "/.repo_requirements/odoo/odoo/models.py", line 5032, in new
    record._cache.update(record._convert_to_cache(values, update=True))
  File "/.repo_requirements/odoo/odoo/models.py", line 4911, in _convert_to_cache
    for name, value in values.iteritems()
  File "/.repo_requirements/odoo/odoo/models.py", line 4912, in <dictcomp>
    if name in fields
  File "/.repo_requirements/odoo/odoo/fields.py", line 1740, in convert_to_cache
    raise ValueError("Wrong value for %s: %r" % (self, value))
ValueError: Wrong value for project.project.account_class: u'deliverables'

AaronHForgeFlow:
@dreispt Thank you for your comments. I'll attend to everything.

regarding this point:

> Its not possible to make an existing "Project" record a child of another existing "Project" record.

This is because you can assign a parent project for an existing project.

@elicoidal That was fixed. Thanks for the report

Regards

aheficent:
Travis error is unrelated:
`2017-09-04 16:47:57,616 4926 ERROR openerp_test openerp.addons.mail.models.mail_mail: failed sending mail (id: 1) due to type object 'TestTaskDelegation' has no attribute '_mails_args`'

project_wbs/models/account_analytic_account.py
AaronHForgeFlow: Added spaces [here](https://github.com/OCA/project/pull/264/files#diff-600b527f4536b12326fea3aed304bc48R181). Thanks

project_wbs/tests/test_project_wbs.py
AaronHForgeFlow: @mreficent is that ok?

MiquelRForgeFlow (approved):
squash minor commits

rafaelbn:
Hello @dreispt @elicoidal , I think @aheficent attend your comments. It would be very nice to make final review. Thanks!

PR #265, opened 2017-05-04
[10.0][MIG] project_stage_closed

luismontalba:
Project Task Stage Closed
=====================

This module provides a 'closed' flag on project task stages.

Usage
--------

To use this module, you need to:

* Have Manager rights for Project group.
* Go to *Project > Configuration > Stages*.
* You have the "Closed" flag available to be checked on the form view.

cc @Tecnativa

PR #266, opened 2017-05-05
[10.0][MIG] project_timesheet_time_control

luismontalba:
Project Timesheet Time Control
==========================

* This module adds a button at account analytic line level to compute the spent
  time, in minutes, from start date to the current moment.
* It improves the domain applied to task field for accommodating to a project
  flow.
* It adds a datetime field that replaces ``date`` field in tree view, and write
* Finally, it allows to open and close tasks from account analytic lines.
  The selected closed stage is the first one that is found with the mark
  "Closed" checked.

Usage
--------

You can access via timesheets:

* Go to Timesheets > My Timesheet > Detailed Activities.
* Create a new record.
* You will see now that the "Date" field contains also time information.
* If you don't select any "Analytic account/project", you will be able to
   select any task, opened or not.
* Selecting a task, the corresponding analytic account/project is filled.
* Selecting an analytic account/project, tasks are filtered for only allow
   to select opened tasks for that project. Remember that an opened task is
   a task whose stage doesn't have "Closed" mark checked.
* At the end of the line, you will see an icon of a cross inside a circle.
* When you press this button, the difference between "Date" field and the
   current time, writing this in the field "Duration".
* You can modify the "Date" field for altering the computation of the
   duration.

or via tasks:

* Go to Project > Search > Tasks.
* Click on one existing task or create a new one.
* On the "Timesheets" page, you will be able to handle records the same way
   as you do in the above explanation (except the task selection part, which
   in this case doesn't appear as it's the current one).

cc @Tecnativa


luismontalba:
This PR depends on #265

pedrobaeza:
Please rebase now that I have merged the dependency

dreispt (approved):

project_timesheet_time_control/views/account_analytic_line_view.xml
```diff
@@ -1,18 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
 
-    <record id="account_analytic_line_tree" model="ir.ui.view">
+    <record id="hr_timesheet_line_tree_inherited" model="ir.ui.view">
```
pedrobaeza: Don't rename the view. Put the same name as the original one.

pedrobaeza (changes requested):
When I don't have any analytic account selected, I only see tasks that don't have a project assigned. The idea is to see all the tasks in this case. Please look for the domain returned on the account_id onchange

pedrobaeza (approved):

pedrobaeza:
Merging for continuing with the last dependent module. @luismontalba you can go on with project_issue_timesheet_time_control

PR #268, opened 2017-05-09
[10.0][MIG] project_issue_timesheet_time_control

luismontalba:
Project Issue Timesheet Time Control
==============================
This module adds a button at project issue timesheets to compute the spent
time, in minutes, from start date to the current moment.

It also allows to open and close issues from account analytic lines.
The selected closed stage is the first one that is found with the mark "Closed"
checked.

Installation
-------------
This module is auto-installed when you install
project_timesheet_time_control and project_issue_sheet.

Usage
--------
You can access via timesheets:

*    Go to Timesheets > My Timesheets > Detailed Activities.
*    Create a new record.
*    If you don't select any "Project", you will be able to select any "Issue", opened or not.
*    Selecting an "Issue", the corresponding "Project" is filled.
*    Selecting a "Project", "Issues" are filtered for only allowing to select opened issues for that project.       Remember that an opened issue is an issue whose stage doesn't have "Closed" mark checked.
*    At the end of the line, you will see an stop icon. When you press this button, the difference between "Date" field and the current time, writing this in the field "Duration".
*    You can modify the "Date" field for altering the computation of the duration.

or via issues:

*    Go to Project > Search > Issues.
*    Click on one existing issue or create a new one.
*    On the "Timesheets" page, you will be able to handle records the same way as you do in the above explanation (except the issue selection part, which in this case doesn't appear as it's the current one).

cc @Tecnativa


pedrobaeza:
Why are you not following migration guide?

luismontalba:
I do it!  (or I think I do it ....)
My steps:
git clone https://github.com/OCA/project -b 9.0
cd project
git checkout -b 10.0-mig-project_issue_timesheet_time_control origin/10.0

And then my local repo is updated and the module disappear. So I put the upgraded module in my local repo and commit changes, but by this way we can see changes ...

Can you help me?
Thanks

pedrobaeza:
You are forgetting the command:
```
git format-patch --stdout origin/10.0..origin/9.0 -- $MODULE | git am -3
```
that brings back the module.

You mustn't copy the module.

luismontalba:
I see now, thank You !

PR #271, opened 2017-05-09
[10.0][MIG] project_closing: Migration to Odoo 10

Jerther:
- Migration of module project_closing. The project.state field has been replaced by project.active so the module must follow this modification.
- Addition of unit tests.


Jerther:
I read the OCA Guidelines, the maintainer-tools' Migration to version wiki, and also the [quick guide on the web site](https://odoo-community.org/page/code). I really did my best to get this straight. I started with multiple commits and ended up squashing them into a single one, but the unit tests coverage would not run again.

I also don't know what to do with the problems found by the code review bot...

P.S.: Sorry for opening a new PR for this. I understand I can just update this one.

dreispt:
Thank you for contributing.
You have only [two minor lint issues](https://travis-ci.org/OCA/project/jobs/230407835#L529):

`project_closing/models/project.py:13:1: W391 blank line at end of file` - add a line break at the EoF
`project_closing/tests/test_project.py:8:80: E501 line too long (89 > 79 characters)` - break the like in two

project_closing/README.rst
```diff
@@ -4,8 +4,8 @@
 Project closing
 ===============
 
-Automatic account analytic closing when related project is closed,
-and if a project is opened, the related analytic account will be re-opened.
+Automatic account analytic archiving when related project is archived,
+and if a project is dearchived, the related analytic account will be dearchived as well.
```
dreispt: I know it's annoying, but can you please recreate the README using the latest template https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

dreispt (commented):
LGTM, waiting for the minor fixes

dreispt (approved):

project_closing/models/project.py
```diff
@@ -0,0 +1,12 @@
...
+from odoo import models, api
+
+
+class ProjectProject(models.Model):
+    _inherit = 'project.project'
+
+    @api.multi
+    def toggle_active(self):
+        for record in self:
+            record.analytic_account_id.active = not record.active
```
espo-tony: It's better to invoke _toggle_active_ method on `record.analytic_account_id` in order to keep consistency with the already existing logic (and eventually with other modules editing that functionality).

espo-tony (changes requested):
The module seems to work very well and does what it is supposed to do. Still, I'd like to suggest a couple of enhancements:

- To alter the Confirmation Message on the button. It currently says: `"(Un)archiving a project automatically (un)archives its tasks and issues. Do you want to proceed?"`; It would be better to add, in these message, that also the related analytic account will be (un)archived.

- As remarked in the code, I think it would be much preferable to (un)archive the _account.analytic.account_ object invoking its own _toggle_active_ method rather than just changing the _active_ flag state.

Moreover, on a functional side, I strongly believe we should build this module in order to work both ways: if an analytic account is archived, its related project should be archived too, in order to preserve consistency between analytic account and project (this is definitely necessary if you all agree with me on invoking the _toggle_active_ method on the _account.analytic.account_ object).

project_closing/models/project.py
dreispt: @espo-tony I checked and `toggle_active` is provided by the Models API.
So AFAICS its pointless to redefine it here.
IMO this should be completely removed.

AaronHForgeFlow (approved):
LGTM. However, wouldn't be correct to archive the project if the analytic account is archived?

dreispt:
@aheficent AFAIK project and analytic_account share the same `active` field.

dreispt:
Do you agree on removing the `project_closing/models/project.py` file?

LoisRForgeFlow:
@dreispt thats the only logic in this module, are you saying this module is not needed at all?

project_closing/models/project.py
espo-tony: @dreispt I disagree: toggle_active is defined in the base model, therefore it's available for all the inherited models (which basically means all the models), but it's still possible (and in cases such as this covered by this module, needed) to extend its behavior for single models.
In our case we need to add further logic, which applies only when the method toggle_active is invoked by the project.project model. This is correctly addressed by the committed code, as far as I can see. I just think it should make use of the method which is deputed to apply the archive/unarchive logic (toggle_active) also for the account.analytic.account model.

espo-tony:
As far as I've seen, project.project and account.analytic.account don't share the same active field. Removing this file, the module won't work at all: the base behavior when a project is archived, is to not archive the related analytic account.

AaronHForgeFlow:
So if this is the standard behavior this modules would do nothing then. I'll check if they share the active field.

LoisRForgeFlow:
The only question remaining is, whether the project should be archived or not when archiving a analytic account. What do you think?

espo-tony:
I think it should.

project_closing/models/project.py
dreispt: @espo-tony I took a better look at this, and I agree with your comment. I can do it directly in this PR.

dreispt:
@espo-tony Please check my fix.

espo-tony:
@dreispt I think it's cleaner like this, thanks! I think we should just decide if we want to make the functionality two-ways: when an analytic account is archived, also the related project should be.

Currently, if we are in the situation of an archived analytic account related to an unarchived project, and we archive the project, it will automatically unarchive the analytic account.

dreispt:
@espo-tony Yes it should. That should be added then.

Jerther:
@espo-tony @dreispt I'm on it :)

dreispt:
@espo-tony I'm also already on it: its not trivial, in order not to get cycles

dreispt:
The action should be done from the Analytic Account method...

Jerther:
@dreispt  Ohh, true. Go ahead then.

espo-tony:
@dreispt @Jerther  Yes, avoiding the loop is the issue here. I think the easiest way to do it could be to pass a parameter in the context of the first invoked method and to check for it in the other. Something like:

In project.project:
```
if not self.env.context.get('avoid_loop', False): 
    project.analytic_account_id.with_context(avoid_loop=True).toggle_active()
```

And in account.analytic.account:
```
if not self.env.context.get('avoid_loop', False): 
    self.env['project.project'].search([('analytic_account_id','=',analytic.id)]).with_context(avoid_loop=True).toggle_active()
```

Not the cleanest solution, maybe, but it should be effective.

dreispt:
My implementation is here: https://github.com/metalsartigan/project/pull/1/files
Waiting for the result of the build here: https://travis-ci.org/dreispt/project/builds/237170926

dreispt:
My changes are now passing tests: https://travis-ci.org/dreispt/project/builds/237174982

LoisRForgeFlow (approved):
Now seems good. Code review :+1:

espo-tony (approved):
:+1: 
It works like a charm!

dreispt:
Congratulations @Jerther , I think this is your first merged PR! 🎂

Jerther:
I had some minor ones before, but this is the first with such collaboration. Thanks :)

PR #273, opened 2017-05-19
[ADD][9.0] Project task delegation

SodexisTeam:


project_task_delegate/models/project.py
```diff
@@ -0,0 +1,78 @@
...
+    _inherit = 'project.task'
+
+    delegated_user_id = fields.Many2one(
+        'res.users', related='child_ids.user_id', string='Delegated To')
+
+    @api.model
+    def _delegate_task_attachments(self, delegated_task_id):
+        attachment = self.env['ir.attachment']
+        attachment_ids = attachment.search(
+            [('res_model', '=', self._name), ('res_id', '=', self.id)])
```
dreispt: This is a Model method, so `self` should be irrelevant and we shouldn't get values from it...
Should this rather be an `@api.multi` method?

project_task_delegate/models/project.py
```diff
@@ -0,0 +1,78 @@
...
+        new_attachment_ids = []
+        for attachment_id in attachment_ids:
+            new_attachment_ids.append(attachment_id.copy(
+                default={'res_id': delegated_task_id}))
```
dreispt: AFAICS this duplicated the attachments.
Would it be better to instead just link to an existing attachment_id?

project_task_delegate/models/project.py
```diff
@@ -0,0 +1,78 @@
...
+            delegated_task_id = self.copy({
+                'name': vals[task.id]['name'],
+                'project_id': vals[task.id]['project_id'],
+                'stage_id': vals[task.id]['stage_id'],
+                'user_id': vals[task.id]['user_id'],
+                'planned_hours': vals[task.id]['planned_hours'],
+                'parent_ids': vals[task.id]['parent_ids'],
+                'description': vals[task.id]['description'],
+                'child_ids': vals[task.id]['child_ids'],
+            })
```
dreispt: Pass the entire `vals` dict to `copy()` instead of picking the actual keys to use.
For this you probably want to pop `task_name` from the dict, and leave the `remaining_hours` and `planned_hours`out of the prepare method.
This allows for ` _prepare_delegate_values` to be more flexible and support different lists of keys.

project_task_delegate/models/project.py
```diff
@@ -0,0 +1,78 @@
...
+            task._delegate_task_attachments(delegated_task_id)
+
+            task.write({
+                'remaining_hours': vals[task.id]['remaining_hours'],
+                'planned_hours': vals[task.id]['task_planned_hours'],
```
dreispt: I think it's OK to extract these two values directly from `delegate_data`, to avoid dealing with them in `_prepare_delegate_values`.

project_task_delegate/wizards/project_task_delegate.py
```diff
@@ -0,0 +1,125 @@
...
+        task = self.env['project.task'].browse(record_id)
+        task_name = tools.ustr(task.name)
+
+        if 'project_id' in fields:
+            res['project_id'] = int(
+                task.project_id.id) if task.project_id else False
+
+        if 'name' in fields:
+            if task_name.startswith(_('CHECK: ')):
+                newname = tools.ustr(task_name).replace(_('CHECK: '), '')
```
dreispt, line 62: I wonder where this `CHECK:` constant comes from?

project_task_delegate/wizards/project_task_delegate.py
```diff
@@ -0,0 +1,125 @@
...
+                _check_rec(child)
+            return True
+
+        _check_rec(eview)
+        res['arch'] = etree.tostring(eview)
+        for field in res['fields']:
+            if 'Hours' in res['fields'][field]['string']:
+                res['fields'][field]['string'] = res['fields'][
+                    field]['string'].replace('Hours', tm)
+        return res
```
dreispt, line 106: AFAICS the point for this `fields_view_get()` is to make the "Hours" field visible, depending  on the value`user:id.company_id.project_time_mode_id`.
Wouldn't it be much simpler to have that value available as a related field, and then use it in an `attrs` attribute in the form view field?

project_task_delegate/wizards/project_task_delegate.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @api.multi
+    def delegate(self):
+        self.ensure_one()
+        context = dict(self._context or {})
```
dreispt: Not used - delete.

dreispt (commented):
Great effort.
Made a few comments on things that probably came from the original implementation, but we probably could improve now.

project_task_delegate/models/project.py
atchuthan: if we link it to an existing attachment and the delegated task is deleted, how do we handle this special case? so, we used the Odoo's default functionality

dreispt: Ideally you would just unlink the attachments.
But I understand that doing so is not so straightforward.
We can leave it as is, but could you add a note on the Roadmap as a possible improvement?

project_task_delegate/models/project.py
dreispt: Are you planning doing something on these two suggestions?

project_task_delegate/wizards/project_task_delegate.py
dreispt: Ping!

project_task_delegate/models/project.py
dvdhinesh: @dreispt yes, I am refactoring the original implementation including the above-mentioned suggestion.

project_task_delegate/models/project.py
dvdhinesh: I will add it to the Roadmap for now.

project_task_delegate/models/project.py
dreispt: Great! Let us know when its ready for review.

project_task_delegate/wizards/project_task_delegate.py
dvdhinesh: [R74](07e1bb2823e91a4b658a5bdda87b682c609a5119#diff-df61f31c4ab01ba08a21ebf274c6dec5R74)

```
        if 'prefix' in fields:
            if task_name.startswith(_('CHECK: ')):
                newname = tools.ustr(task_name).replace(_('CHECK: '), '')
            else:
                newname = tools.ustr(task_name or '')
            prefix = _('CHECK: %s') % newname
            res['prefix'] = prefix
```

Here we are adding the constant `CHECK: ` to the delegation wizard's `default_get`. 
That constant will be written to the parent task if unchanged by the user while doing delegation.

project_task_delegate/models/project.py
dvdhinesh: It's done :+1:

project_task_delegate/wizards/project_task_delegate.py
dvdhinesh: @dreispt i am not sure i understand the above comment. 
The idea is to change the string of `planned_hours` and `planned_hours_me` fields based on the value from the config settings. 
So the `Planned Hours` string on the wizard will be displayed as `Planned Hour(s)` or `Planned Day(s)`.
Please let me know if i missed something.

dreispt: I see -maybe we can do that with XML, using `<label for="..." attrs="{...}">`

project_task_delegate/wizards/project_task_delegate.py
dreispt: ok

dreispt:
Let me know when its ready for review.

project_task_delegate/wizards/project_task_delegate.py
dvdhinesh: so we had to define the `label` for all the fields that has the string `Hours` in it?
i am thinking what would happen if someone inherits and add their own fields.

dvdhinesh:
@dreispt yes. please apply the needs review label and tell us your feedback for improvements.

atchuthan (approved):
code review :+1:

dreispt (approved):

PR #274, opened 2017-05-19
[10.0][FIX] project_timesheet_time_control: fix onchange

luismontalba:


pedrobaeza (approved):
Fast-tracking this when CIs finish as this is the same code as in previous version, but it was lost in the migration.

PR #252, opened 2017-04-13
10.0 mig project issue code

thinkwelltwd:
I explicitly [followed the directions](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0#technical-method-to-migrate-a-module-from-90-to-100-branch); Copied & Pasted. 

The exact same problem as https://github.com/OCA/project/pull/244 still remains. I don't know why. Perhaps the directions don't apply because `project_issue_code` module never got migrated to the v10 branch.

In any case, it'd be nice if this module could be upgraded to v10. If my small snippet of migration work is helpful and you can merge this, wonderful. If not, could somebody else just make it happen?

PR #276, opened 2017-05-29
[FIX] project_issue_code : Fixed v10.0 migration

Garamotte:
Attend comments done a bit too late in the migration PR : #252 
The copy method prevented custom values to be defined while duplicating records.

cc @pedrobaeza @dreispt @lreficent @aheficent

### DIFF
--- a/project_issue_code/data/project_issue_sequence.xml
+++ b/project_issue_code/data/project_issue_sequence.xml
@@ -1,11 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<odoo>
-    <data noupdate="1">
-        <record id="project_issue_sequence" model="ir.sequence">
-            <field name="name">Issue code</field>
-            <field name="code">project.issue</field>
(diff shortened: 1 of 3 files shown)


dreispt (approved):

espo-tony (changes requested):
What happens when a record is duplicated without passing a default value for _issue_code_?
As far as I can see, the _issue_code_ is not duplicated from the original record, but as nothing is passed as parameter, it will set its value to _None_. That's not allowed by the ORM definition of the field itself, which says: `required=True` .

I'd like to see a test for this case.

Garamotte:
There's already a test for this case : https://github.com/OCA/project/blob/10.0/project_issue_code/tests/test_project_issue_code.py#L37-L43
The test checks that the value is not the same as the copied record's value, and it also checks the value with the next to be generated from the sequence.

espo-tony (approved):

AaronHForgeFlow (approved):
:+1: Thanks

PR #277, opened 2017-05-29
9.0 project propose end date

serpentcs-dev1:
## Project Propose End Date

Proposes the end date based on the start date of the project

Please Review @aheficent


AaronHForgeFlow (changes requested):
Thank you for your contribution but I cannot install the module :collision::

image

project_propose_end_date/__openerp__.py
```diff
@@ -0,0 +1,20 @@
...
+    "author": "Eficent",
+    "website": "http://www.eficent.com",
+    "category": "Generic",
+    "depends": ["project"],
+    "license": "AGPL-3",
+    "data": [
+        "views/project_view.xml",
+    ],
+    'installable': True,
+    'active': False,
```
dreispt: Deprecated, you may remove it.

project_propose_end_date/__openerp__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# Copyright 2017 Serpent Consulting Services Pvt. Ltd.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Project Propose End Date",
+    "version": "9.0.1.0.0",
+    "summary": "Proposes the end date based on the start date",
```
dreispt, line 9: We need a README file with a longer description

project_propose_end_date/models/project_project.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Project(models.Model):
+    _inherit = "project.project"
+
+    @api.multi
+    @api.onchange('date_start')
+    def onchange_date_start(self):
+        for project in self:
+            if project.date_start and not project.date:
+                project.date = project.date_start
```
dreispt, line 17: The point is to copy the start date to the end date?

dreispt:
TravisCI confirms the error: https://travis-ci.org/OCA/project/jobs/237177460#L1020

elicoidal (commented):
Not sure to understand the finality of this module

project_propose_end_date/models/project_project.py
serpentcs-dev1: Yes, it populates the end date if empty from the start date.

AaronHForgeFlow:
Basically for me it does not make much sense a project without and end date and this field should always be informed (with the estimated date). If this Pull Request is not the proper way maybe we can discuss in an issue and close this one. 

Regards

dreispt:
@darshan-serpent can you describe a typical use case for this feature?

AaronHForgeFlow:
Hi @dreispt The typical case is that the users create many projects that were never taken so if the manager needs to filter the projects that are currently undertaken they get a lot not needed information. 

But thinking it twice maybe this is not OCA module and maybe it's just a customization.

AaronHForgeFlow (commented):
Maybe would it be better to add a new field called 'estimated end date' that is required. Then the end_date is filled using that value. What do you think? @dreispt

AaronHForgeFlow:
Besides, I'd add the expiration date to the project at the time it's closed or done @dreispt

AaronHForgeFlow:
HI @darshan-serpent The module is working fine now. However, what it's not a generic module but  a customization. IMHO this can be closed.

serpentcs-dev1:
@aheficent Yes right agree, it is not enough to include it as a whole new module. I'll close it for now, thanks!

PR #278, opened 2017-05-29
[9.0][MIG] project_closing to v9

LoisRForgeFlow:
Backport of https://github.com/OCA/project/pull/271


dreispt (approved):
LGTM

project_closing/README.rst
```diff
@@ -1,21 +1,36 @@
...
-If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
-`here <https://github.com/OCA/project-service/issues/new?body=module:%20project_closing%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
-
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/project/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smash it by providing detailed and welcomed feedback.
 
 Credits
 =======
```
cubells: Add Images section after credits section in order to add oca icon as is in https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst file

project_closing/__openerp__.py
```diff
@@ -1,29 +1,15 @@
...
     "name": "Project closing",
-    "version": "8.0.0.1.1",
-    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "version": "9.0.1.0.0",
+    "author": "Camptocamp, Métal Sartigan, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
     "website": "http://www.camptocamp.com",
     "category": "project Management",
     "depends": ["project"],
     "data": [],
```
cubells: Remove keys with empty lists

project_closing/models/analytic.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+# Copyright 2013 Camptocamp Author Vincent Renaville
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, api
```
cubells: Reorder import alphabetically

cubells (changes requested):
Fast code review. Minor changes.

project_closing/models/analytic.py
dreispt: @cubells That's not required

project_closing/README.rst
dreispt: Isn't that only relevant when we include images from other authors?

project_closing/models/project.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class ProjectProject(models.Model):
+    _inherit = 'project.project'
+
+    @api.multi
+    def toggle_active(self):
+        if not self.env.context.get('doing_project_toggle_active'):
+            # When called directly from Project, delegate to Analytic Account
+            res = self.analytic_account_id.toggle_active()
```
cubells: Analytic_account_analytic has no attribute active

This line doesn't raise an error??

cubells (changes requested):

project_closing/models/project.py
dreispt: No, `toggle_active` is a [Model API method](https://github.com/odoo/odoo/blob/9.0/openerp/models.py#L5314).

cubells: So I've got this error on archiving a project 

captura de 2017-05-29 18-31-28

project_closing/models/analytic.py
elicoidal: Actually it is (pay attention to the comment in the code-block here:https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports)
I think it is nearly never enforced: I tried in the past and all developers complaining that it was not necessary since there is tool for it (isort).

dreispt: I see, you are right, the rule is there.
Personally, when I import several modules in the same line, the order has some meaning or logic - usually the order they should later be used. For me this usually translates to `models, fields, api`.
I'm for coding standards, but there is a point past which rules do more harm than good.

project_closing/models/project.py
dreispt: It looks like Analytic Accounts don't have an `active` field [on 9.0](https://github.com/odoo/odoo/blob/9.0/addons/analytic/models/analytic.py#L14).

LoisRForgeFlow: yes, I was added in v10, I'll addapt the module.

Thanks @cubells

LoisRForgeFlow:
In v9 there is no status in `account.analytic.account`. I have used `account_type` instead. Do you agree with my approach? @cubells @dreispt

AaronHForgeFlow (approved):
After the last change using account_type it works fine :+1:

dreispt (approved):

MiquelRForgeFlow:
Please, can someone explain me the meaning of this backport? The module still has `'installable': False`.

PR #282, opened 2017-05-30
8.0 [ADD] project issue timesheet matching

thomaspaulb:
When timesheet lines are read in from an external source, eg. through the [Toggl connector](https://github.com/OCA/hr-timesheet/pull/102) are not connected to a project issue. This module allows to match timesheet lines to issues.


pedrobaeza:
Isn't better to integrate this on the connector itself?

thomaspaulb:
Personally, at least for 8.0, I had the same problem even before I was working with the connector. If an employee fills in their timesheet, or imports their hours from their Excel/CSV file, there is the same problem, the issue is not filled, and in standard Odoo I did not find a convenient way to add it.

pedrobaeza:
OK then

feketemihai:
@thomaspaulb For me the same, i think it's too specific, so try to incorporate it in the other PR.

PR #284, opened 2017-05-31
[IMP] Add default locations to consume materials

cubells:
These changes allows user to define default locations from and to which materials must be consumed by project.

cc @Tecnativa

### DIFF
--- a/project_task_materials_stock/models/project.py
+++ b/project_task_materials_stock/models/project.py
@@ -6,6 +6,23 @@
 from openerp import models, fields, api, exceptions, _
 
 
+class ProjectProject(models.Model):
+    _inherit = 'project.project'
+
+    location_source_id = fields.Many2one(
+        comodel_name='stock.location',
+        string='Source Location',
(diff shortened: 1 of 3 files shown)


project_task_materials_stock/models/project.py
```diff
@@ -6,6 +6,23 @@
 from openerp import models, fields, api, exceptions, _
 
 
+class ProjectProject(models.Model):
+    _inherit = 'project.project'
+
+    location_id = fields.Many2one(
```
dreispt: Can we name this `location_source_id`?
I've been using `location_id` in Tasks for the location where the work is to be done.
There is no actual collision, but using the same name for different things can be confusing.

cubells: @dreispt sure!

dreispt (approved):

rafaelbn (approved):
Tested in runbot 👍

pedrobaeza (approved):

PR #286, opened 2017-06-02
[FIX][10.0]fix auto install error

leemannd:
When launching test I had an auto-install error linked to this module


dreispt:
@leemannd The current CI build is green, and I see nothing wrong with the current code. 
Can you provide details on your error?

dreispt (commented):
I believe we want the `auto_install` flag here. Do you agree closing the PR?

leemannd:
@dreispt We can close it, but still I

PR #289, opened 2017-06-12
fix amount_currency computation

damdam-s:
use the date in context to be able to use currency rates

### DIFF
--- a/project_timesheet_currency/__manifest__.py
+++ b/project_timesheet_currency/__manifest__.py
@@ -13,10 +13,12 @@
         'analytic',
         'sale_timesheet',
         'project',
+        'account',
     ],
     'data': [
         'security/record_rules.yml',
         'security/analytic_security.xml',
         'security/project_security.xml',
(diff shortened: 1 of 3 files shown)


leemannd (approved):

jgrandguillaume (commented):
Hi,


Did we really want to call the field "timesheet_currency_id" and name it "TS original currency" ?

Wouldn't it be better to call it project_currency_id ?

damdam-s:
@jgrandguillaume the project's currency is the one of the analytic line.
There, I though it was useful to be able to see the `amount_currency` with the currency of the user who timesheeted.

jgrandguillaume:
@damdam-s Got it, yes it is a good idea. I was disturb because in my vision we would have:

 * Company currency
 * Project currency

This was how I designed it on our v7.. But I figured out that now, project and company currency are always the same, so I guess what you did is alright.

:+1:

gurneyalex (approved):

PR #283, opened 2017-05-30
Port project_stage_state to 9.0

dreispt:

PR #294, opened 2017-06-26
[ADD]project_change_state

AaronHForgeFlow:
Project Change State
================
This module allows the user to change the status of multiple projects simultaneously.

## Roadmap/Issues
- Change the stage of the task using the module project_stage_state (https://github.com/OCA/project/pull/283)
- Propose the expiration_date when closing a project

### DIFF
--- a/project_change_state/wizard/project_change_state.py
+++ b/project_change_state/wizard/project_change_state.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from openerp import api, models
+
+
+class ProjectChangeState(models.TransientModel):
+    _name = "project.change.state"
+    _description = "Change the status of the selected projects"
+
+    @api.multi
+    def set_pending(self):
+        project_ids = self.env.context.get('active_ids')
+        self.env['project.project'].browse(project_ids).set_pending()
+        return {'type': 'ir.actions.act_window_close'}
+
+    @api.multi
+    def set_open(self):
+        project_ids = self.env.context.get('active_ids')
+        self.env['project.project'].browse(project_ids).set_open()
+        return {'type': 'ir.actions.act_window_close'}
+
+    @api.multi
+    def set_close(self):
+        project_ids = self.env.context.get('active_ids')
+        self.env['project.project'].browse(project_ids).set_close()
+        return {'type': 'ir.actions.act_window_close'}
+
+    @api.multi
+    def set_cancel(self):
+        project_ids = self.env.context.get('active_ids')
+        self.env['project.project'].browse(project_ids).set_cancel()
+        return {'type': 'ir.actions.act_window_close'}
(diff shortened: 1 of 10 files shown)


AaronHForgeFlow:
@Eficent

mreficent:
Please rebase ☺️

project_change_state/models/project.py
```diff
@@ -0,0 +1,48 @@
...
+            project.state = 'close'
+
+    @api.multi
+    def set_cancel(self):
+        for project in self:
+            if project.state not in ('open', 'pending', 'close'):
+                raise ValidationError(_("Selected project(s) cannot be "
+                                        "changed to 'Cancelled' state as they "
+                                        "are not all in 'Pending', 'Open' "
+                                        "or 'Close' states."))
```
dreispt: Do we really need all these validations? What if they were removed?

project_change_state/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_product_file_import_user,project.change.state,model_project_change_state,project.group_project_user,1,1,1,1
```
dreispt: Wizards don't need ACLs, you should remove this file.

project_change_state/wizard/project_change_state.py
```diff
@@ -0,0 +1,37 @@
...
+
+class ProjectChangeState(models.TransientModel):
+    _name = "project.change.state"
+    _description = "Change the status of the selected projects"
+
+    @api.multi
+    def set_pending(self):
+        project_ids = self.env.context.get('active_ids')
+        for project in self.env['project.project'].browse(project_ids):
+            project.set_pending()
```
dreispt: No need for a for-loop. Just use: 

    self.env['project.project'].browse(project_ids).set_pending()

Or:

    projects = self.env['project.project'].browse(project_ids)
    projects.set_pending()

dreispt (commented):
Overall LGTM, just made a few comments and suggestions

project_change_state/models/project.py
AaronHForgeFlow: Actually, nothing would happen. I'll remove this because it's a customization

project_change_state/wizard/project_change_state_view.xml
```diff
@@ -0,0 +1,30 @@
...
+                    <p class="oe_grey">
+                        You can change the status of multiple
+                        projects simultaneously.
+                    </p>
+                    <footer>
+                        <button name="set_pending" string="Pending" class="oe_highlight" type="object" groups="base.group_user"/>
+                        <button name="set_open" string="Re-open Project" type="object" groups="base.group_user"/>
+                        <button name="set_close" string="Close Project" type="object" groups="base.group_user"/>
+                        <button name="set_cancel" string="Cancel Project" type="object" groups="base.group_user"/>
+                        <button string="Cancel" class="oe_link" special="cancel" groups="base.group_user"/>
```
dreispt: Di we need the ` groups="base.group_user"`?

dreispt (approved):
It would be great to have tests, but let's move on with this.

project_change_state/wizard/project_change_state_view.xml
AaronHForgeFlow: I agree. Better project.group_project_user

dreispt: You're right, but IMO either way it is redundant. Why not remove it?

MiquelRForgeFlow (approved):
LGTM :+1:

PR #299, opened 2017-07-18
[10.0][ADD] Module to check progress and worked hours for a project

benwillig:
This module simply computes planned and spent hours for a project.


project_timesheet_hours/README.rst
```diff
@@ -0,0 +1,56 @@
...
+Installation
+============
+
+Just install it.
+
+Configuration
+=============
+
+No configuration needed.
+
```
rafaelbn: Please add "usage" section and explain where we have to go for checking this new information. Thanks!

pedrobaeza:
Please don't include plurals in the name

dreispt (approved):
LGTM

sbidoul:
@pedrobaeza don't you think the plural `hours` is justified in this case?

I mean, singular is good for classes (in a broad sense -- eg project, timesheet), but here `hours` refer to class variables which are naturally plurals in the code (eg `planned_hours`).

pedrobaeza:
The rule is for module names, without exceptions. See for example `project_task_material` and `project_task_material_stock` that was renamed in v10 because the `s` in the middle makes a lot of times that the second module name was mistyped.

benwillig:
@rafaelbn Usage section added.
@pedrobaeza Removed plurals in module name.

rafaelbn (approved):
Tested in runbot

rafaelbn:
@pedrobaeza if agree please merge, thanks

project_timesheet_hour/models/project_project.py
```diff
@@ -0,0 +1,45 @@
...
+
+from odoo import api, fields, models
+
+
+class ProjectProject(models.Model):
+
+    _inherit = 'project.project'
+
+    planned_hours = fields.Float(
+        string="Planned hours", compute='_compute_hours', store=True)
```
pedrobaeza: Don't use the same compute method for the 3 fields, but instead call directly the other methods for the inheritance (although renaming to `_compute_*`).

pedrobaeza (changes requested):

project_timesheet_hour/models/project_project.py
benwillig: I created a method for each field.

project_timesheet_hour/models/project_project.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @api.multi
+    @api.depends(
+        'total_hours_spent', 'planned_hours')
+    def _compute_progress(self):
+        for rec in self:
+            rec.progress = rec._get_progress()
+
+    @api.multi
+    def _get_planned_hours(self):
```
pedrobaeza: You don't need now this intermediate methods (that's why I said about renaming directly these methods to `_compute`). Any way, please include the code directly in `_compute` methods.

rafaelbn: HI @benwillig could you please check this?

project_timesheet_hour/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV, Odoo Community Association (OCA)',
+    'website': 'https://www.acsone.eu',
+    'depends': [
+        'hr_timesheet',
+    ],
+    'data': [
+        'views/project_project.xml',
+    ],
+    'demo': [
```
Jerther: Avoid empty keys.

Jerther (changes requested):

dreispt:
@pedrobaeza I fixed your comments; can you check?

benwillig:
Sorry, I forgot to check back this PR. About the intermediate methods in the compute methods, I understand it might look 'overengineering' but, if you want to change the compute method only in some case, you must loop on all records and make a super on a single record if you want to keep base method functionning. That's why I added theses intermediate methods. I'm aware that if you inherit and add an other field dependency, you must inherit from the compute method, add the field and then make a single super, but you won't have to make several super while looping. I have a use case where I have to change the computation of those fields, and with the last commit, I have to loop twice on the recordset or make several super.

pedrobaeza:
@benwillig, not really. You can always use recordset operations and call super only with the remaining records.

Check for example:
```python
def _compute_whatever(self):
    group_of_interest = self.filtered(lambda x: <criteria>)
    for record in group_of_interest:
        <compute the new way>
    super(Class, self - group_of_interest)._compute_whatever()
```

I think you shouldn't create so much method overhead when there are simple solutions through recordsets.

benwillig:
Yes, I'm agree, but then you loop more than one time on the recordset. In your example, We have one loop on the filtered, an other one to process records, and an other one to process the super.
And what if when inheriting, you need the result of the super method ? (then the result of the intermediates methods)

project_timesheet_hour/models/project_project.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @api.multi
+    @api.depends(
+        'total_hours_spent', 'planned_hours')
+    def _compute_progress(self):
+        for rec in self:
+            progress = 0.0
+            if rec.planned_hours:
+                progress = rec.total_hours_spent / rec.planned_hours * 100.0
+            rec.progress = progress
```
benwillig: @dreispt Removing the round method is not a good idea.

pedrobaeza:
@benwillig, compute methods mustn't return values, and a iteration over a recordset is very cheap. Please trust me on this that I have passed the "create hooks phase".

benwillig:
About the hypothetic performances issues, ok I can trust you.
My intermediate methods were not 'compute' method, so there is no problem for them to return values.

We can merge this, but as I said in a code comment, @dreispt should readd the round method on the progress field before merging.

And as I said, I understand that some people dislike thoses intermediates methods, but I still think it's the better way to do when there could be some inheritance. But that's only my opinion.

project_timesheet_hour/models/project_project.py
dreispt: Why?

benwillig: Because you don't want to display something like 6.6666666666666666667.

acsonefho (approved):

dreispt:
Hi @benwillig do you plan to follow up on this PR?

rafaelbn:
@dreispt in the @benwillig last comment said

> We can merge this, but as I said in a code comment, @dreispt should readd the round method on the progress field before merging.

Are you going to readd something?

In this PR the problem is how to make the code, I think we could unblock and merge this module as it works.

PR #302, opened 2017-08-01
Added default stage to tree view

PCatinean:
Boolean field was missing from tree view making it harder to identify default stages.

### DIFF
--- a/project_task_default_stage/__manifest__.py
+++ b/project_task_default_stage/__manifest__.py
@@ -6,7 +6,7 @@
 {
     'name': 'Project Task Default Stage',
     'summary': 'Recovery default task stage projects from v8',
-    'version': '10.0.1.0.0',
+    'version': '10.0.1.1.0',
     'category': 'Project',
     'author': 'Tecnativa, '
               'Odoo Community Association (OCA)',
--- a/project_task_default_stage/views/project_view.xml
+++ b/project_task_default_stage/views/project_view.xml
@@ -12,6 +12,17 @@
         </field>
     </record>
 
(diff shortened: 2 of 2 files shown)


pedrobaeza (changes requested):
Please change module version according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

PCatinean:
Done

project_task_default_stage/__manifest__.py
```diff
@@ -6,7 +6,7 @@
 {
     'name': 'Project Task Default Stage',
     'summary': 'Recovery default task stage projects from v8',
-    'version': '10.0.1.0.0',
+    'version': '10.0.2.0.0',
```
pedrobaeza: It should be 10.0.1.1.0

PCatinean: Thought the same since it's a small change though it matched better the middle version that said it should require module update. Also if someone should do a position after the same element it would change the design slightly.

But if you think it should increment the last number I can make the change, your call.

pedrobaeza: Yeah, but I think you have read incorrectly the guide. The change I propose is for changes that require a module update.

PCatinean: Ah, I read it corectly but the 0 from 10.0 got me confused.

pedrobaeza (approved):
Thanks

rafaelbn (approved):
Tested in runbot 👍

dreispt (approved):

PR #311, opened 2017-08-23
[FIX] project_double_alias: Get proper reply to address

pedrobaeza:
In v10, the reply to address is not properly caught when having several aliases for the same parent model. We inject a condition in the alias search for avoiding the problem.

cc  @Tecnativa


pedrobaeza:
@rafaelbn @Yajo please review

yajo (approved):
Code review

gurneyalex (approved):

PR #319, opened 2017-09-25
8.0 project definition

gfcapalbo:
Add a new Tab Definitions to projects. A link to the definition model introduced by document_definition module


hbrunn (changes requested):
did you consider just using a combination of `document_page_partner_id` and possibly `document_page_tags` for this?

gfcapalbo:
No , because that links a partner to a page, not per project
THe idea of making definitions just pages popped in my mind , but having them distinguished by tags would force us to do more work to keep them out of document_page.
We do not want the definitions to be mixed up in the document_page views. I was also able to impose a UNIQUE sql constraint for the definition term.
Plus with document_definition this i could add the with a lot less work on domains we could offer a set of defintions to add in the project.

hbrunn:
I think you most decidedly should not want to keep definitions out of document_page. It's a piece of knowledge, and that's what we use document pages for.

In any case, a many2many has the advantage that you can recycle stuff, and the big disadvantage that you recycle stuff. This will cause users to change things in the context of one project, just to find out that they've changed the definition globally. A more flexible version of this would be (either on document.page as I think is right, or on document.definition what you have) two optional fields `partner_id` and `project_id` and a function field that collects all global definitions, partner specific ones from the project's partner and project specific definitions if any. The form view then can warn when you're editing a partner global or database global definition.

gfcapalbo:
closing this. will repost with better datamodel.

PR #320, opened 2017-09-26
8.0 project definition

gfcapalbo:
Adds a new document page type "Term" to create dictionary-style word archive and
definition.  Adds Searchable view  and menu for "terms and definitions" in the Knowledge menu.
Has a search view per term and per definition and a dictionary Style Alphabetical grouping view.

Shows term definitions on every project.  just adds a relation between
document_page of type "term"  and project models and adds a tab to the project view.

### DIFF
--- a/project_definition/README.rst
+++ b/project_definition/README.rst
@@ -1,0 +1,54 @@
+. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+============================
+Term definitions in projects
+============================
+
+Adds a new document page type "Term" to create dictionary-style word archive and
+definition. 
+Adds Searchable view for terms and definitions in the Knowledge menu.
+Has a search view per term and per definition and a dictionary Style Alphabetical grouping view.
+Shows term definitions on every project. just adds a relation between
+document_definition  and project models and adds a tab to the project view.
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
+<https://github.com/OCA/project/issues>`_. In case of trouble, please
+check there if your issue has already been reported. If you spotted it first,
+help us smashing it by providing a detailed and welcomed feedback.
+
+Credits
+=======
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Giovanni Francesco Capalbo <giovanni@therp.nl>
+* Holger Brunn <hbrunn@therp.nl>
+
+Do not contact contributors directly about help with questions or problems concerning this addon, but use the `community mailing list <mailto:community@mail.odoo.com>`_ or the `appropriate specialized mailinglist <https://odoo-community.org/groups>`_ for help, and the bug tracker linked in `Bug Tracker`_ above for technical issues.
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
--- a/project_definition/__openerp__.py
+++ b/project_definition/__openerp__.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Term Definitions in projects",
+    "version": "8.0.1.0.0",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "",
+    "summary": "",
+    "depends": [
+        'project',
+        'document_page',
+    ],
+    "data": [
+        'views/templates.xml',
+    ],
+    "installable": True,
+}
--- a/project_definition/models/document_page.py
+++ b/project_definition/models/document_page.py
@@ -17,6 +17,11 @@ class DocumentPage(models.Model):
     # need to be stored in order to use grouping
     first_letter = fields.Char(compute='_get_first_letter', store=True)
 
+    project_id = fields.Many2one(
+        'project.project',
+        'Project'
+    )
+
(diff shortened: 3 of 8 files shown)


project_definition/__openerp__.py
```diff
@@ -0,0 +1,20 @@
...
+    "name": "Term Definitions in projects",
+    "version": "8.0.1.0.0",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "",
+    "summary": "",
+    "depends": [
+        'project',
+        'knowledge',
+        'document_page_partner_id',
```
gfcapalbo: currently useless dependency, will need it if we want specific partner_id and project_ids associations in future. Module lives on same repo.

hbrunn: change to document_page anyways, many people are dependency purists

project_definition/views/templates.xml
```diff
@@ -0,0 +1,92 @@
...
+        <record id="view_definition_form" model="ir.ui.view">
+            <field name="name">document.definition.form</field>
+            <field name="model">document.page</field>
+            <field name="priority">100</field>
+            <field name="arch" type="xml">
+                <form string="Definitions">
+                    <sheet>
+                        <h1>
+                            <field name="name"/>
+                        </h1>
```
hbrunn: I think it would be nice to have a div here that warns about editing a global term, with `invisible="context.get('active_model') != 'project.project'` which causes the div to show up when you edit this inline from a project, but not otherwise

project_definition/__openerp__.py
```diff
@@ -0,0 +1,20 @@
...
+{
+    "name": "Term Definitions in projects",
+    "version": "8.0.1.0.0",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "",
+    "summary": "",
+    "depends": [
+        'project',
+        'knowledge',
```
hbrunn: what's this dependency good for?

project_definition/models/document_page.py
```diff
@@ -0,0 +1,23 @@
...
+
+    # a relation to partner_id is given by our dependency on
+    # document_page_partner_id if there is no partner_id and no project_id
+    # specified the term is global
+
+    type = fields.Selection(selection_add=[('term', 'Term')])
+
+    # stored compute field used for making the letter grouping
+    # need to be stored in order to use grouping
+    first_letter = fields.Char(compute='_get_first_letter', store=True)
```
hbrunn, line 18: that's fine for now, but in the future consider other more elegant options:

- generate filters A-Z + rest in the search view, either manually or in `fields_view_get`
- use a custom group operator that does this
- have a module that adds A-Z + rest as buttons to the search view which will do a name_search('$letter%', operator='=ilike')
- have a module that overrides [`_read_group_process_groupby`](https://github.com/OCA/OCB/blob/8.0/openerp/models.py#L1945) to support some syntax like `$char_field:first`

project_definition/views/templates.xml
```diff
@@ -0,0 +1,92 @@
...
+                </form>
+            </field>
+        </record>
+
+        <!-- filters to group by alphabetical order, and search by term or in
+         definition -->
+
+        <record id="view_definition_filter" model="ir.ui.view">
+            <field name="name">document.definition.search</field>
+            <field name="model">document.page</field>
```
hbrunn, line 71: also add a high priority here

project_definition/views/templates.xml
```diff
@@ -0,0 +1,92 @@
...
+            <field name="arch" type="xml">
+                <search string="Definitions">
+                    <field name="name" string="Search Term" filter_domain="[('name','ilike',self)]"/>
```
hbrunn: any reason to add a filter domain which does the same as the [default](https://github.com/OCA/OCB/blob/8.0/addons/web/static/src/js/search.js#L1424)?

hbrunn (changes requested):
to fix CI, you'll have to add the knowledge repository to the [`oca_dependencies.txt`](https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/oca_dependencies.txt) file

project_definition/views/templates.xml
gfcapalbo: nice, but it was conceived to be also managed from the knowledge interface. We can create terms and only after add them to multiple projects if we want.

project_definition/views/templates.xml
gfcapalbo: to give it another label. "Search Term"

hbrunn: I talk about `filter_domain`

project_definition/views/templates.xml
hbrunn: yes, and that's nice. My proposal would just reassert to the user what she should know in the first place

hbrunn: so this warning will be shown when we're *not* on a project...

gfcapalbo: fixed. also added the form_ref to the context when clicking on the many2many

mFlayyan:
:+1:

mFlayyan (approved):

project_definition/README.rst
```diff
@@ -0,0 +1,54 @@
...
+============================
+Term definitions in projects
+============================
+
+Adds a new document page type "Term" to create dictionary-style word archive and
+definition. 
+Adds Searchable view for terms and definitions in the Knowledge menu.
+Has a search view per term and per definition and a dictionary Style Alphabetical grouping view.
+Shows term definitions on every project. just adds a relation between
+document_definition  and project models and adds a tab to the project view.
```
dreispt, line 14: Can we have a usage example, showcasing a situation where this feature is needed?

project_definition/views/templates.xml
```diff
@@ -0,0 +1,106 @@
...
+            <field name="search_view_id" ref="view_definition_filter"/>
+            <field name="domain">[('type', '=', 'term')]</field>
+            <field name="context">{'default_type':'term'}</field>
+        </record>
+
+        <menuitem 
+            id="menu_definition_top"
+            parent="knowledge.menu_document"                 
+            name="Definitions"
+            sequence="10" />
```
hbrunn: increase sequence so that this comes after https://github.com/OCA/knowledge/blob/8.0/document_page/document_page_view.xml#L7

project_definition/README.rst
gfcapalbo: In our local usage we needed a different place to have terms and definitions to attach to a project.  We wanted to create a "terms and definitions" type with it's own menu, search views and  filters without fractioning the KB with a new model.
So we extended page model with a new type "term"
The advantage of this is that such "terms" are visible in their own section, and when attaching terms to a project our selection will show only available terms , not all pages of the KB.

PR #321, opened 2017-09-27
[FIX] project_timesheet_time_control: Condition always gets evaluated as False

fuentes73:
Hello,

We've notice that if a new analytic line is created with a date_time different from today. The field date is never equals to field date_time, because date has a [default value](https://github.com/odoo/odoo/blob/10.0/addons/analytic/models/analytic_account.py#L97) defined at analytic module. So the condition in eval_date function is always evaluated as False

### DIFF
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -32,7 +32,7 @@ class AccountAnalyticLine(models.Model):
             self.project_id = self.task_id.project_id.id
 
     def eval_date(self, vals):
-        if vals.get('date_time', False):
+        if 'date_time' in vals and vals.get('date_time'):
             vals['date'] = fields.Date.from_string(vals['date_time'])
         return vals
 
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -32,7 +32,7 @@ class AccountAnalyticLine(models.Model):
             self.project_id = self.task_id.project_id.id
 
     def eval_date(self, vals):
-        if 'date_time' in vals and 'date' not in vals:
+        if vals.get('date_time', False):
             vals['date'] = fields.Date.from_string(vals['date_time'])
         return vals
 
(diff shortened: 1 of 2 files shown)


project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -32,7 +32,7 @@ def onchange_task_id(self):
             self.project_id = self.task_id.project_id.id
 
     def eval_date(self, vals):
-        if 'date_time' in vals and 'date' not in vals:
+        if vals.get('date_time', False):
```
pedrobaeza: Please change it then to `if 'date_time' in vals and vals.get('date_time'):`

pedrobaeza (changes requested):
Please change module version number

fuentes73:
Done

pedrobaeza (approved):
@luismontalba @cubells please review

luismontalba (approved):
Tested on runbot

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -32,7 +32,7 @@ def onchange_task_id(self):
             self.project_id = self.task_id.project_id.id
 
     def eval_date(self, vals):
-        if 'date_time' in vals and 'date' not in vals:
+        if 'date_time' in vals and vals.get('date_time'):
```
carlosdauden: I think that `get` is not needed in this case.
`if 'date_time' in vals and vals['date_time']:`

carlosdauden (approved):
Minor comments

project_timesheet_time_control/models/account_analytic_line.py
lasley: Actually removing the first part and leaving the `get` would be functionally equivalent and cleaner:
```
if vals.get('date_time'):
```

pedrobaeza: Well, I wanted to avoid the `None` case, but you're right. It was correct the first time.

fuentes73: Ok, I'm going to change it as @lasley says.

PR #334, opened 2017-10-31
Update README.md

Prernak17:


pedrobaeza:
This change is incorrect as the correct length is the current one.

PR #337, opened 2017-11-09
IMP Create related project from template

rubencabrera:
Added the possibility of setting a project template to allow salesperson to directly create related project from it.

### DIFF
--- a/sale_order_project/models/sale.py
+++ b/sale_order_project/models/sale.py
@@ -86,13 +86,10 @@ class SaleOrder(models.Model):
                         'date': new_date_end,
                         'parent_id': order.project_template_id.parent_id.id
                     })
-                new_id = order.project_template_id.copy(default=vals)
-                order.write({
-                    'project_id': new_id.analytic_account_id.id
-                })
+                project = order.project_template_id.copy(default=vals)
             else:
                 project = project_obj.create(vals)
-                order.write({
-                    'project_id': project.analytic_account_id.id
-                })
-                return True
+            order.write({
+                'project_id': project.analytic_account_id.id
+            })
+            return True
--- a/sale_order_project/models/sale.py
+++ b/sale_order_project/models/sale.py
@@ -23,7 +23,8 @@
 ###############################################################################
 
 from openerp import api, fields, models
-from datetime import date
+from datetime import date, datetime
+import time
 
 
 class SaleOrder(models.Model):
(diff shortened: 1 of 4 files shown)


rubencabrera:
@carlosap92

sale_order_project/models/sale.py
```diff
@@ -59,8 +65,34 @@ def action_create_project(self):
...
+                        'date': new_date_end,
+                        'parent_id': order.project_template_id.parent_id.id
+                    })
+                new_id = order.project_template_id.copy(default=vals)
+                order.write({
+                    'project_id': new_id.analytic_account_id.id
+                })
+            else:
+                project = project_obj.create(vals)
+                order.write({
```
dreispt: The `order.write` is repeated in the if and else. You can have it only once, outside the if-else.

sale_order_project/models/sale.py
```diff
@@ -59,8 +65,34 @@ def action_create_project(self):
         project_obj = self.env['project.project']
         for order in self:
             vals = self._prepare_project_vals(order)
-            project = project_obj.create(vals)
-            order.write({
-                'project_id': project.analytic_account_id.id
-            })
-        return True
+            if order.project_template_id:
+                new_date_start = time.strftime('%Y-%m-%d')
```
dreispt: It is more correct to use `fields.Date. context_today()`. See http://www.odoo.com/documentation/10.0/reference/orm.html

sale_order_project/models/sale.py
```diff
@@ -59,8 +65,34 @@ def action_create_project(self):
...
+                    start_date = date(*time.strptime(
+                            order.project_template_id.date_start, '%Y-%m-%d'
+                        )[:3])
+                    end_date = date(*time.strptime(
+                            order.project_template_id.date,
+                            '%Y-%m-%d')[:3])
+                    new_date_end = (datetime(*time.strptime(
+                            new_date_start,
+                            '%Y-%m-%d'
+                        )[:3])+(end_date-start_date)).strftime('%Y-%m-%d')
```
dreispt, line 92: AFAICT if the template has start and end dates, you take the duration from it and apply it to the new project.
This feature should be documented, at least in the README.
You get bonus points if you can improve the readability to make this evident 😸

dreispt (commented):
LGMT, just a few improvement suggestions

sale_order_project/models/sale.py
rubencabrera: I added it to the README, with the text to datetime conversions I find it quite difficult to make the code readable. But if you have any idea of how to tackle it, I'd like to try.

sale_order_project/models/sale.py
rubencabrera: Done, seems github doesn't take the change as done but it's out of the if/else now. Thanks for pointing it out!

dreispt (approved):

sale_order_project/models/sale.py
```diff
@@ -30,6 +30,10 @@
 class SaleOrder(models.Model):
     _inherit = "sale.order"
 
+    description = fields.Char(
+        string = 'Descripción del proyecto'
```
dreispt, line 34: This should be in english

PR #338, opened 2017-11-09
[9.0][FIX] project_closing: instalable true

LoisRForgeFlow:
Fixes https://github.com/OCA/project/pull/278#issuecomment-343189937


MiquelRForgeFlow (approved):

LoisRForgeFlow:
cc @dreispt @aheficent

pedrobaeza (approved):
We can fast-track this as obvious

PR #347, opened 2017-12-01
[11.0][ADD] project_parent_task_filter

emagdalenaC2i:
This module adds a filter to show only the parent tasks in a project


elicoidal (approved):
LGTM (not yet tested)

emagdalenaC2i:
Any clue about the failures of Runbot and Travis is appreciated?
In Travis: 
_Couldn't load module project_parent_task_filter
cannot import name 'project'_

pedrobaeza:
You have missed to add project.py file.

project_parent_task_filter/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
carlosdauden: Is needed?

carlosdauden (changes requested):

project_parent_task_filter/__init__.py
emagdalenaC2i: No. I was looking that right now, I could delete the models folder.
And the project_parent_task_filter/__init__.py should be empty or could be deleted?

carlosdauden: Remove model folder and empty project_parent_task_filter/init.py but don't remove it.

carlosdauden (approved):

dreispt:
This seems too small to be a module.
Can we add this to an existing module instead?

emagdalenaC2i:
You are right, but I don´t know which module fits better with it. Maybe none.
Also I am working in an improvement that will add more functionality.

PR #348, opened 2017-12-05
[10.0][MIG] sale_order_project

Abranes:
Migration of the module `sale_order_project` to `10.0`.

We remove the field `related_project_id` because now it's present in `sale` base module.


pedrobaeza:
Is this really needed with current v10 possibilities about products that creates tasks?

Abranes:
That's true it's not really needed with v10. But if you want to create a related empty project and after create your tasks by hand this module is useful. (That is the use case of one of our customers)

pedrobaeza:
OK, a requirement a bit weird, but I have seen worse, hehe.

Check Travis.

Abranes:
Travis is complaining about project_closing tests. sale_order_project tests are OK.

pedrobaeza:
But main branch is green, so it should be your module that is breaking the other one.

astirpe:
Rebase the branch. The Travis failure should be fixed now.

acsonefho (changes requested):
Hello, I think that you should check into the action_create_project (in sale.order) that a project_id is not already set for the sale order (and raise an exception if it's the case).
We can't fully trust a button. If there is an issue with a button (or if someone add it somewhere else without the invisible attribute).

PR #349, opened 2017-12-23
11.0 mig project task default stage

novawish:


project_task_default_stage/views/project_view.xml
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
```
pedrobaeza: This shouldn't be removed

pedrobaeza:
This PR makes `project_closing` tests to fail. Please check.

novawish:
I checked module `project_closing` and its tests failed itself. Therefore i don't think the failure was caused by this PR

pedrobaeza:
It should be a change upstream, because it was green on merge time. @Jerther can you please check the module you migrate?

ghost:
Same on https://github.com/OCA/project/pull/351#issuecomment-354138259

project_task_default_stage/data/project_data.xml
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
```
ghost: I'd like to leave encoding headers in XML files :)

project_task_default_stage/models/project.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def _get_type_common(self):
+        ids = self.env['project.task.type'].search([
+            ('case_default', '=', True)])
+        return ids
+
+    type_ids = fields.Many2many(
+        comodel_name='project.task.type', relation='project_task_type_rel',
+        column1='project_id', column2='type_id', string='Tasks Stages',
+        default=_get_type_common
```
ghost: remove everything except for `default=_get_type_common` attribute, since everything else [was already defined](https://github.com/odoo/odoo/blob/11.0/addons/project/models/project.py#L189) in core `project` module, you have no need to define it again

project_task_default_stage/data/project_data.xml
```diff
@@ -0,0 +1,49 @@
+<odoo noupdate="1">
```
arkostyuk: I'd like to leave coding headers in XML files :)

ghost (changes requested):

melroy89:
can't review until fixed.

ghost (approved):

novawish:
@pedrobaeza @danger89 I re-run travis and it's ok now, please check

melroy89 (approved):

project_task_default_stage/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    'license': 'AGPL-3',
+    'depends': [
+        'project',
+    ],
+    'data': [
+        'views/project_view.xml',
+        'data/project_data.xml',
+    ],
+    'installable': True,
+}
```
benwillig, line 21: Key 'website' is missing there. it should be repository URL.

project_task_default_stage/models/project.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ProjectProject(models.Model):
+    _inherit = 'project.project'
+
+    def _get_type_common(self):
+        ids = self.env['project.task.type'].search([
+            ('case_default', '=', True)])
+        return ids
+
+    type_ids = fields.Many2many(default=_get_type_common)
```
benwillig: Could you please use a lambda there? Then we can inherit from this method.
Also I think that '_get_type_common' could be renamed to '_get_default_type_common' or something similar. but at least prefixed by '_get_default'.

benwillig (changes requested):

project_task_default_stage/models/project.py
novawish: I'm not sure what you mean. Did you expect something like this: `default=lambda self: self._get_type_common()`? If yes, then i suggest we just simply put the function name inside quotes: `default='_get_type_common'`. By that, we can inherit from the method also.

pedrobaeza: No, @novawish, default is the only Odoo method that you can't inherit enclosing the method name into quotes. This is because you can pass an string also as default, making Odoo to not know what do you want to make. You have to user the `lambda` trick.

That was for <=v10, but I haven't tried in v11. I suppose it happens the same that in previous versions.

novawish: @pedrobaeza You are absolutely right. I'm mistaken it with `compute` attribute, fixed with lambda function.

project_task_default_stage/__manifest__.py
novawish: Fixed, thanks. Can you please check again?

benwillig (approved):

PR #357, opened 2018-01-11
[10.0][REM] project_closing removal (became obsolete)

astirpe:
This is a proposal to remove module project_closing from 10.0 branch.
Backport of https://github.com/OCA/project/pull/356
Motivation and discussion is here: #353
Fixes #353

PR #358, opened 2018-01-11
[10.0][REM] project_closing removal (became obsolete)

mpanarin:
same as https://github.com/OCA/project/pull/356 but for v10.0


pedrobaeza:
There's already #357.

astirpe:
@mpanarin you may want to approve https://github.com/OCA/project/pull/357, it's exactly the same as this PR.

PR #364, opened 2018-02-23
[ADD][11.0] Adding module project_portal

pnajman-modoolar:
Hello,

Module ``project_portal`` overrides methods of the project portal controller to make them more usable for other modules.

Kind Regards,
Petar


pedrobaeza:
I don't like this too much, because this makes any derivative module from this one incompatible with the rest. You should propose to Odoo that hooks instead.

pnajman-modoolar:
Due to the comment of @pedrobaeza I'm closing this pull request

LoisRForgeFlow:
@pnajman-modoolar Did you propose any PR to odoo? could you point them?

pnajman-modoolar:
Not yet but i'll definitely do that ...

PR #307, opened 2017-08-18
[MIG][10.0]project_wbs

AaronHForgeFlow:
Tasks done so far:
- rename __openerp__.py to __manifest__.py
- change readme and version in the manifest

PR #366, opened 2018-02-23
[ADD][11.0] Adding module project_key

pnajman-modoolar:
Hello,

Module ``project_key`` provides functionality to uniquely identify projects and tasks by ``key`` field.
Project key is either auto generated out of project name or manually entered.
Task key is always auto generated from its project specific ``ir.sequence``

The general idea is to have JIRA like task key sequencing: _PA-1, PA-2 for Project A_,  _PB-1, PB-2 for Project B_ and so on.

Module supports tasks re-indexing in case you decide to update key of an running project.

With this module you can also browse projects and tasks by their keys:
- https://<<your-domain>>/browse/PROJECT-KEY
- https://<<your-domain>>/browse/TASK-KEY

Anyhow, I've seen that there is a module ``project_task_code`` which brings something similar but not quite the same. So, I'm contributing these functionalities as a separate module hopping we can all decide whether to keep them as a separate modules or to squash them into one module. 
   
Kind Regards,
Petar


pedrobaeza:
We already have project_task_code that I figure it's the same. There's no project_code though, because it usually takes the analytic account code.

pnajman-modoolar:
Hi @pedrobaeza,

Thanks for you input, I've looked at ``project_task_code`` module to be sure of it's implementation and as far as I can see they are similar but not quite the same. The code on task is created from a single ``ir.sequence`` which will generate global task code sequence, so there is no option to have independent task sequences per project.

For example: project_task_code module will generate task codes from a single sequence and thus you'll have first task of the project_a with code 1 and then you could have a first task of the project_y with code 3899 and so on.

What this module introduces are unique project code and task code sequence per project.
For example if we have project_a with code PA the code of the first task would be PA-1, for project_b with code PB the code of the first task would be PB-1 and so one. Also, what's good in this approach are smaller task codes and ability to put task in context of the project just by looking at the task code.   

Good point about analytic account code. As that field is not populated when project is created, maybe I could adapt module to use analytic account code instead of adding additional field on project?

project_key/README.rst
```diff
@@ -0,0 +1,73 @@
...
+#. Go to Project > Configuration > Projects
+#. Create
+#. Enter project name and use auto generated key or simply override value by entering your own key value.
+
+In form View:
+
+#. Go to Project > Dashboard
+#. Open the projects settings
+#. Modify the "key" value
+#. After modifying project key any existing task will be reindexed automatically.
```
StefanRijnhart: I find the terminology confusing as this does not concern the sql column index. How about `the code of any existing tasks will be updated automatically` (rename terminology in the code as well).

StefanRijnhart:
I like reusing the analytic code field.

project_key/README.rst
```diff
@@ -0,0 +1,73 @@
...
+
+On ``project.task`` level:
+
+#. Actually there is nothing to be done here
+#. Task keys are auto generated based on project key value with per project auto incremented number (i.e. PA-1, PA-2, etc)
+
+In browser address bar:
+
+#. Navigate to your project by entering following url: http://<<your-domain>>/browse/PROJECT-KEY
+#. Navigate to your task by entering following url: http://<<your-domain>>/browse/TASK-KEY
```
StefanRijnhart: I see you mimicked Jira here that uses `browse` but is this not too generic for Odoo? How about using `tasks` or `tickets`?

project_key/models/project_project.py
```diff
@@ -0,0 +1,183 @@
...
+        for rec in self:
+            if rec.name:
+                rec.key = self.generate_project_key(rec.name)
+            else:
+                rec.key = ''
+
+    @api.model
+    def create(self, vals):
+        if 'key' not in vals:
+            if vals['name']:
```
StefanRijnhart: `name` is a required field. If it can be falsey, maybe it can also be the case that it is not in `vals` so you might want to use `vals.get('name')`. If it cannot, do away with the condition entirely.

project_key/models/project_project.py
```diff
@@ -0,0 +1,183 @@
...
+    @api.model
+    def name_search(self, name, args=None, operator='ilike', limit=100):
+        args = args or []
+        domain = []
+        if name:
+            domain = [
+                '|', ('key', 'ilike', name + '%'), ('name', operator, name)
+            ]
+            if operator in expression.NEGATIVE_TERM_OPERATORS:
+                domain = ['&'] + domain
```
StefanRijnhart: Overriding the domain in `name_search` is always tricky. For instance, if args is not defined, it looks like you end up with an unbalanced domain. The same can happen if args is an unbalanced domain itself. And to be compatible with any other module that might be installed on the same database, it would be better to call the super method. But my main question is, what is the goal of this method override?

project_key/models/project_project.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        values = {
+            'name': "%s %s" % (
+                _("Project task sequence for project "), self.name
+            ),
+            'implementation': 'standard',
+            'code': 'project.task.key.%s' % (self.id,),
+            'prefix': "%s-" % (self.key,),
+            'use_date_range': False,
+            'active': False,
```
StefanRijnhart: Why is it set to inactive?

StefanRijnhart:
Nice work! There are no tests in this module. You could easily add a test to check what happens if the project key gets changed (effect on project sequence and task codes).

project_key/controllers/portal.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2018 Modoolar <info@modoolar.com>
+# License LGPLv3.0 or later (https://www.gnu.org/licenses/lgpl-3.0.en.html).
+
+from odoo import _
+from odoo.osv.expression import OR
+from odoo.addons.project_portal.controllers.portal import CustomerPortal
```
StefanRijnhart: We do not have this in Odoo 11CE I'm afraid.

project_key/README.rst
pnajman-modoolar: Hi @StefanRijnhart: I've considered your comment and I've actually decided to rename the module to "project_code" and also I've changed the terminology as per your comment

project_key/README.rst
pnajman-modoolar: I've changed it to ``projects`` and ``tasks``

project_key/models/project_project.py
pnajman-modoolar: project ``name`` will always be present in the ``vals`` ...

project_key/models/project_project.py
pnajman-modoolar: The idea behind this method was to include task ``code`` into name search.

project_key/models/project_project.py
pnajman-modoolar: At first I didn't want this sequence to be seen as it's maintained by the module itself but now when you mentioning that I've removed that :)

project_key/controllers/portal.py
pnajman-modoolar: I've added that module via another pull request, but as I got reviews overt here that it's best to give that to Odoo, I've removed that dependency for now.

project_key/README.rst
pnajman-modoolar: As it turns the new code was in collision with module ``project_task_code`` so I switched it back.

AaronHForgeFlow:
I agree with @pedrobaeza regarding the sequence for projects In v10 we used https://github.com/OCA/account-analytic/tree/10.0/account_analytic_sequence and take the sequence from the analytic account. Is the sequence for the analytic account back in v11? If not we can migrate that module. 

We also use that sequence here to calculate the complete_code for the complete hierarchy here: https://github.com/OCA/project/pull/307
For tasks i think it's ok to add the new feature.

pnajman-modoolar:
@aheficent, @StefanRijnhart, @pedrobaeza, I understand your points here. It make sense to use ``code`` field from account.analytic.account instead of adding new field ``key`` but regarding project and task sequencing this module mimics JIRA like sequencing. It let's you manually choose project code/key (it event propose project code based on project name)  and even let's you to redefine it at any point. So, imho i don't see how we would integrate  ``account_analytic_sequence`` as the logic is quite different.

pedrobaeza:
OK, let's continue this way, but please put CIs in green.

pnajman-modoolar:
Can somebody please take a look the ci/runbot [log file](https://runbot.odoo-community.org/runbot/static/build/3328421-366-5c6252/logs/job_10_test_base.txt)  it seems that there are some docker related issues ...

pnajman-modoolar:
As far as I can see the only errors are those docker related.

project_key/models/project_project.py
StefanRijnhart: ~~Am I missing something? I don't see any references to task `code` in this method.~~
Ah, you call it `key`.

project_key/models/project_project.py
StefanRijnhart: But can it be falsey?

project_key/models/project_project.py
StefanRijnhart: To restore the compatibility with other modules (e.g. core project and analytic modules which have their own name_get logic), I would reap a list of ids from super and do a new search on [('|', ('key', 'ilike', name + '%'), ('id', 'in', ids)] and return name_get on the result of that.

StefanRijnhart:
About runbot, in the cogwheel dropdown click on Full all logs. You can see the following warnings:
```
2018-06-12 17:07:23,416 186 WARNING openerp_test odoo.models: ir.sequence.create() includes unknown fields: key
2018-06-12 17:07:23,541 186 WARNING openerp_test odoo.models: ir.sequence.create() includes unknown fields: key
2018-06-12 17:07:23,565 186 WARNING openerp_test odoo.models: ir.sequence.create() includes unknown fields: key
2018-06-12 17:07:23,586 186 WARNING openerp_test odoo.models: ir.sequence.create() includes unknown fields: key
2018-06-12 17:07:23,627 186 WARNING openerp_test odoo.models: ir.sequence.create() includes unknown fields: key
2018-06-12 17:07:23,652 186 WARNING openerp_test odoo.models: ir.sequence.create() includes unknown fields: key
```
These make runbot 'fail'.

pnajman-modoolar:
Thank you very much. I'll get right on it.

project_key/models/project_project.py
pnajman-modoolar: I see :D

project_key/models/project_project.py
pnajman-modoolar: Fixed.

project_key/models/project_project.py
pnajman-modoolar: Fixed.

StefanRijnhart (approved):
Thanks for the updates!

pnajman-modoolar:
@pedrobaeza @aheficent we need one more positive review :)

pnajman-modoolar:
I've added some unit tests as well ...

pnajman-modoolar:
Hey guys, can someone help me with codecov? What can I do to make it green? Thanks ...

AaronHForgeFlow:
Hi @pnajman-modoolar you can click on Details here in the Pull Request. Then in the new window there is a "diff" tab. Click on that and you will see the lines that are not tested highlighted in red.

pnajman-modoolar:
Me again, module is fully covered with unit tests :)

AaronHForgeFlow:
100% codecov, good job! :medal_military:

pnajman-modoolar:
Thanks man. Can we get one more positive review so we can merge this :)

AaronHForgeFlow:
I'll post my review soon :)

AaronHForgeFlow (approved):
:+1: It works fine. Here I'm adding the code field to the project #307 but that module it's adding many other features and it's in v10. If both modules are merged in OCA then, I guess I'll rename that field to key also. Thanks

pnajman-modoolar:
We have two positive reviews. Let's merge this PR :)

pnajman-modoolar:
Thank you guys for all your help!

pedrobaeza:
Thanks to you for your contributions

StefanRijnhart:
What's a good module from the scrum project to review next?

pnajman-modoolar:
imho, before we can get to project-agile repo, we should plow through dependency modules and their satellites. [project_agile](https://github.com/OCA/project-agile/pull/4) module has dependency to [project_workflow](https://github.com/OCA/project/pull/371) which has dep on [web_diagram_position](https://github.com/OCA/web/pull/879).

So I would say that the following prioritized list should be a good start:

1. [web_diagram_position](https://github.com/OCA/web/pull/879)
2. [project_workflow](https://github.com/OCA/project/pull/371)  
    Then, while we are here it would be good to cover satellite modules
    - [project_workflow_security](https://github.com/OCA/project/pull/372)
    - [project_workflow_default_state_per_group](https://github.com/OCA/project/pull/373)
3. [project_agile](https://github.com/OCA/project-agile/pull/4)
4. Other pr's from  [project agile](https://github.com/OCA/project-agile) repo.

Also, as part of the project we have these modules as well ...

1. [project_git](https://github.com/OCA/project/pull/367)
2. [project_git_github](https://github.com/OCA/project/pull/368)
3. [project_git_bitbucket](https://github.com/OCA/project/pull/369)
4. [project_git_gitlab](https://github.com/OCA/project/pull/370)

Does this sounds like a plan?

StefanRijnhart:
Yes it does!

pnajman-modoolar:
I'm currently writing unit tests for ``project_workflow`` and colleague of mine is working on ``web_diagram_position`` and issues found by travis which we are struggling to reproduce ...

PR #371, opened 2018-02-23
[ADD][11.0] Adding module project_workflow

pnajman-modoolar:
Hi,

Module ``project_workflow`` provides functionality to create fully configurable workflow around ``project.task`` by wrapping project.task.type (i.e. task stage) with workflow states.

This module depends on following custom modules:

- [project_portal](https://github.com/OCA/project/pull/364)
- [web_diagram_position](https://github.com/OCA/web/pull/879)
- [web_ir_actions_act_multi](https://github.com/OCA/web/pull/877)
- [web_ir_actions_act_view_reload](https://github.com/OCA/web/pull/878)

Kind Regards,
Petar

PR #372, opened 2018-02-23
[ADD][11.0] Adding module project_workflow_security

pnajman-modoolar:
Hello,

Module ``project_workflow_security`` extends [project_workflow](https://github.com/OCA/project/pull/371) in order to provide configuration of allowed security groups for workflow transitions. In this way we can control which security role can apply specific task workflow transitions.

Kind Regards,
Petar


project_workflow_security/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2017 - 2018 Modoolar <info@modoolar.com>
+# License LGPLv3.0 or later (https://www.gnu.org/licenses/lgpl-3.0.en.html).
+{
+    "name": "Project Workflow Security",
+    "summary": "Module extends workflow transitions with security groups.",
+    "category": "Project",
+    "version": "11.0.1.0.0",
+    "license": "LGPL-3",
+    "author": "Odoo Community Association (OCA), Modoolar",
+    "website": "https://www.modoolar.com/",
```
nikul-serpentcs, line 10: @pnajman-modoolar IMHO Change URL as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L10)

PR #374, opened 2018-03-28
[ADD][10.0] analytic_account_sequence

AaronHForgeFlow:
Generate an automatic and unique WBS code for the WBS structure


pedrobaeza:
I think this should be in OCA/account-analytic

AaronHForgeFlow:
@pedrobaeza I added it here because it strongly depends on project_wbs but I don't mind to move it to OCA/account-analytic

pedrobaeza:
Then change the module name, and you have conflicts in the code.

AaronHForgeFlow:
Actually this does not depends on project_wbs but the other way around, project_wbs needs a unique sequence. Moved to account-analytic here: https://github.com/OCA/account-analytic/pull/164

PR #378, opened 2018-04-16
[IMP] Implements name_get() in 'project.task'

rodrigets:
Relational fields and the task notification email inform the task code.

email

m2o

### DIFF
--- a/project_task_code/models/project_task.py
+++ b/project_task_code/models/project_task.py
@@ -27,3 +27,14 @@ class ProjectTask(models.Model):
             default = {}
         default['code'] = self.env['ir.sequence'].next_by_code('project.task')
         return super(ProjectTask, self).copy(default)
+
+    @api.multi
+    def name_get(self):
+        result = super(ProjectTask, self).name_get()
+        new_result = []
+
+        for task in result:
+            rec = self.browse(task[0])
+            name = '[%s] %s' % (rec.code, task[1])
+            new_result.append((rec.id, name))
+        return new_result
--- a/project_task_code/__manifest__.py
+++ b/project_task_code/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "Sequential Code for Tasks",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.1.0",
     "category": "Project Management",
     "author": "OdooMRP team, "
               "AvanzOSC, "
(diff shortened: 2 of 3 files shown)


aldo774 (approved):
I think that is a valid implementation :+1:

pedrobaeza (changes requested):
Don't overwrite the method without calling super, as you are breaking inheritance. Call super, and on the result, append at the beginning the code.

mstuttgart (approved):

project_task_code/models/project_task.py
```diff
@@ -27,3 +27,15 @@ def copy(self, default=None):
...
+
+    @api.multi
+    @api.depends('name', 'code')
+    def name_get(self):
+        result = super(ProjectTask, self).name_get()
+        new_result = []
+
+        for task in result:
+            rec = self.browse(task[0])
+            name = '%s - %s' % (rec.code, rec.name)
```
tarteo: Can you change this to:

`name = '%s - %s' % (rec.code, task[1])`

So other modules can still append other strings to it.

tarteo (changes requested):

rodrigets:
Thank you @tarteo , this review is really relevant.
I made the change.

tarteo (approved):
@rodrigets Thanks!

project_task_code/models/project_task.py
```diff
@@ -27,3 +27,15 @@ def copy(self, default=None):
             default = {}
         default['code'] = self.env['ir.sequence'].next_by_code('project.task')
         return super(ProjectTask, self).copy(default)
+
+    @api.multi
+    @api.depends('name', 'code')
```
dreispt: The api.depends is not relevant for name_get, you should remove it.

project_task_code/models/project_task.py
```diff
@@ -27,3 +27,15 @@ def copy(self, default=None):
...
+    def name_get(self):
+        result = super(ProjectTask, self).name_get()
+        new_result = []
+
+        for task in result:
+            rec = self.browse(task[0])
+            name = '%s - %s' % (rec.code, task[1])
```
dreispt: Do you have a strong opinion about this format?
In other modules I used the `[%s] %s` format instead.

pedrobaeza: +1 to this

rodrigets:
@dreispt , I have not strong a opinion for format, i go utilize your suggestion.
Thank you.

dreispt (changes requested):
Waiting for fixes.

rodrigets:
@dreispt , which fixes?

dreispt:
Sorry - they were already there.

dreispt (approved):

pedrobaeza (approved):

PR #384, opened 2018-05-07
[MIG] scripts migration for upgrade safe from v8.0

martintorresvm:


martintorresvm:
Close as duplicated with #386

PR #388, opened 2018-05-16
[FIX] Fix dependency recursion error due to upstream changes in odoo

PCatinean:
Due to a recent commit in upstream Odoo https://github.com/odoo/odoo/commit/e5fb990db1d6216a156617e387d85d13c1a0b93c the name of the method used in this module to check recursion overrides the original one and generates an error when setting subtasks. To avoid overlapping I renamed the method and changed the description slightly.

### DIFF
--- a/project_task_dependency/__manifest__.py
+++ b/project_task_dependency/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'Project Task Dependencies',
-    'version': '11.0.1.0.0',
+    'version': '11.0.1.0.1',
     'category': 'Project',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Please change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

PCatinean:
Done

pedrobaeza (approved):

RoelAdriaans (approved):

PCatinean:
@pedrobaeza thanks!

Issue #391, opened 2018-06-11
[RFC] Base HelpDesk


DarioLodeiros:
Helpdesk will provide the base to manage support tickets. It can be improved with new related modules to manage tickets in website, timesheet, etc.

I’m not sure about the best way to design the model structure, I think that we can choose between 3 main options:

A) New Model ticket with all logic to manage a “special task”… (it’s mean use a o2m or m2m field that allows generate tasks from ticket, etc.)
B) Use delegated inheritance from task… We should hide the linked tasks in project menu.
C) Use the project.task model itself through normal inheritance using a new boolean field in the project model, for example ‘Is helpdesk project’.

I think proposal A option is more generic… we can have different tasks and tickets linked to same project without interfere with the normal project module logic. Set the helpdesk goal on customer issues resolution and service quality… If we need use timesheets or other tasks features in tickets, we can propose use linked tasks to them, and though these control the time, sales, etc.

Proposed Base Features:

- WorkFlow Ticket
- Team Support Manager (notifications, assigned...)
- Project related to the ticket as an option
- Ticket division into tasks  as an option
- Analytic Tickets
- Tickets Types
- Mail integration

Future Features:
- Website Integration with Portal HelpDesk

hbrunn:
a) is what project_issue in earlier versions did, and I'm currently in the process of migrating our helpdesk away from this. We decided to do all helpdesk stuff with tasks, because in the end, you don't want to have to switch between your tasks and your tickets, a ticket *is* a task, so I'd favor c) a lot for a new implementation.

DarioLodeiros:
Hi! Yes! Finally, the idea that we are developing is based in 2 modules, the first is the HelpDesk base, like a new area in Odoo (Its valid to a shop, service company, etc...) the second module allow to link project module with help desk... Its mean create tickets from project alias and access from the project, or Help Desk Team. I think that will be a good job! :)

aitorbouzas:
[This repository has a Helpdesk system](https://github.com/SythilTech/Odoo/tree/11.0/website_support). Maybe it's worth a look even if it differs from your initial implementation idea.
It also has an interesting automatic mail per ticket state system that maybe'll help you with mail integration.

JordiBForgeFlow:
[...]

PR #394, opened 2018-06-13
[ADD][WIP] Base HelpDesk APP 

DarioLodeiros:
#391 
Work In Progress, Base module structure.


helpdesk/data/helpdesk_data.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
```
nikul-serpentcs: @DarioLodeiros IMHO Directly use `<odoo noupdate="1">`

DarioLodeiros: :+1:

PR #396, opened 2018-06-22
[11.0] Add project_task_type_active

grindtildeath:


tschanzt (changes requested):
Although its a very small module i think it still should have a readme

grindtildeath:
@tschanzt There were some breaking changes on the OCA regarding readme during your holidays. You can check this out :
https://github.com/OCA/maintainer-tools/pull/339
https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

tschanzt:
:scream: yeah definitely missed those.

tschanzt (approved):

project_task_type_active/models/project_task_type.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ProjectTaskType(models.Model):
+
+    _inherit = 'project.task.type'
+
+    active = fields.Boolean(default=True)
```
simahawk, line 10: no need for a archive/unarchive button (using toggle_active)?

simahawk (changes requested):

project_task_type_active/models/project_task_type.py
grindtildeath: Not needed, because there is no form view and Archive/Unarchive actions are available automatically.
image

simahawk (approved):

TDu (approved):

PR #403, opened 2018-07-13
[11.0][IMP] project_task_dependency: Auto arranging

tarteo:
Adds an setting which enables arranging start and end dates of tasks based on their dependencies. This feature is optional and can be turned on in the project settings.

auto arrange


project_task_dependency/__manifest__.py
```diff
@@ -1,9 +1,9 @@
-# © 2016 Onestein (<http://www.onestein.eu>)
+# Copyright 2016 Onestein (<http://www.onestein.eu>)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 {
     'name': 'Project Task Dependencies',
-    'version': '11.0.1.0.1',
+    'version': '11.0.1.1.1',
```
astirpe: Should be `'11.0.1.1.0'`

project_task_dependency/models/__init__.py
```diff
@@ -1,4 +1,5 @@
-# © 2016 Onestein (<http://www.onestein.eu>)
+# Copyright 2016 Onestein (<http://www.onestein.eu>)
```
astirpe: No need of Copyright in the `__init__.py` files.

astirpe (changes requested):

astirpe (approved):

astirpe:
@tarteo it would be nice if you can add some screenshots that show the effect of this feature in a timeline view.
This module provides a feature that is useful but also very cool if you see it in action on the timeline. :wink:

tarteo:
I've updated the description with a little video of how it works.

pedrobaeza:
I'm not sure about this, as this is not taking into account weekends and days off. I think we should move this to the schedule planning module we are doing. What do you think?

tarteo:
@pedrobaeza If the arranging can be done on the fly on tasks, then I agree. I'll keep my branch so you can use the code of it if you want.

pedrobaeza:
Not yet, but it can be added using current module methods for skipping days off.

pedrobaeza:
Are you going to continue with this? The branch has conflicts.

tarteo:
@pedrobaeza I have no time for this at the moment, I'll close this one and try to put it project_schedule another time.

tarteo:
And with the addition of https://github.com/OCA/web/pull/1024 this is not really adding much anymore.

pedrobaeza:
Certainly...

PR #409, opened 2018-07-25
[10.0][BACKPORT] project_task_add_very_high

gaelTorrecillas:
Use 4th priority level on tasks (v10.0)


tarteo (changes requested):
High is already in Odoo project module: https://github.com/odoo/odoo/blob/f05b10ce4984c10be9a9664f17f093ba2ec5ee85/addons/project/models/project.py#L330

gaelTorrecillas:
@tarteo I understand your message and suggestion, but we need 3 stars, so add one priority.
I backported from 11.0 to 10.0
https://github.com/OCA/project/blob/11.0/project_task_add_very_high/models/project_task.py#L11

I don't know how to rename 2nd and 3rd priorities, I used the same name of 11.0

tarteo:
@gaelTorrecillas Very High and Critical?

AaronHForgeFlow (commented):
Why changing this? This will require a change in the version number, a migration script and also be ported to v11. Is it really necessary?

flotho:
Hi @aheficent , @tarteo ,
@gaelTorrecillas and I are probably misunderstanding your comment.
The main reason of this change is that we need those 3 level for a customer working in v10.0. And as we love to share our work, we made this PR.
@aheficent , of course you're right regarding the version number and @gaelTorrecillas changed this. But we don't really understood the part of the migration script. Datas are actually stored and expressed in filters with the first part of the tuple for selection field so that all the filter will still correct. 
If somebody want to use this module, he should be aware of the fact that the priority range will change. Our last proposal would be to implement a full rewrite of the values of this selectio  so that it could match to the v11 release.
@aheficent , what's your point of view for this.

regards

project_task_add_very_high/models/project_task.py
```diff
@@ -11,4 +11,4 @@ class ProjectTask(models.Model):
     priority = fields.Selection(selection_add=[
         ('2', 'Very High'),
         ('3', 'critical')
-        ])
+        ])
```
astirpe: Do not remove the newline at the end of the file

project_task_add_very_high/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 {
     "name": "Project Task Add Very High",
     "summary": "Adds an extra option 'Very High' on tasks",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
```
astirpe: Version should be 10.0.1.1.0 at least

astirpe (changes requested):
`def uninstall_hook()` must be updated accordingly, since a new value `3` was added and it must be correctly handled while uninstalling the module.

Migration script to V11 is not necessary IMO

astirpe:
Oh and I forgot to say ... also the readme and the screenshots must be updated accordingly

gaelTorrecillas:
Hello, 

@astirpe I changed readme.

astirpe:
@gaelTorrecillas thanks! Could you also update the `def uninstall_hook()` ?

project_task_add_very_high/models/project_task.py
```diff
@@ -8,4 +8,8 @@
 class ProjectTask(models.Model):
     _inherit = 'project.task'
 
-    priority = fields.Selection(selection_add=[('2', 'Very High')])
+    priority = fields.Selection(selection_add=[
+        ('2', 'Very High'),
+        ('3', 'critical')
```
tafaRU: It would be better to start with a capital letter: `s/critical/Critical`

project_task_add_very_high/models/project_task.py
```diff
@@ -8,4 +8,8 @@
...
+        ])
+
```
tafaRU: no newline at the end of file

tafaRU (changes requested):
Also please check Travis and squash a bit the commits.
Thanks!

gaelTorrecillas:
Hello,

 @astirpe, I updated this function.
@tafaRU I changed the code to be accepted by travis.

tafaRU:
@gaelTorrecillas, it seems https://github.com/OCA/project/pull/409/files#r216627182 hasn't been attended, isn'it?

gaelTorrecillas:
@tafaRU I'm sorry, with commit revert, I forgot to add this change.

I'm doing it now.

tafaRU:
@gaelTorrecillas, ok thanks, could you also please squash your commits in order to keep only the meaningfull ones?

astirpe:
@gaelTorrecillas You also may want to add yourself in the Contributors list.

gaelTorrecillas:
@tafaRU, @astirpe I deleted all commits to create one commit with only useful changes.
but when I verified Travis was not successful.( build #2145).
but I have check one other Travis Build #2146 was successful (for this PR).
I wait your opinion to know how to solve this problem.

tafaRU:
@gaelTorrecillas, I just restarted all the Build Jobs at https://travis-ci.org/OCA/project/builds/430537214

gaelTorrecillas:
@tafaRU thank you.
All checks have passed.

astirpe (approved):

flotho (approved):
LGTM

tafaRU (approved):

tarteo (approved):

AaronHForgeFlow (commented):
@flotho I see your point here, it is not reducing the range but making it bigger. But in v11 we have shorter range and we need to parse then the critical priority to something else. Either that or introduce the same change in v11.
Anyway I tested and it worked. I will not block this.

AaronHForgeFlow (approved):
:+1:  Merging, as approved by the original author of the module. Tested on runbot.

PR #410, opened 2018-07-27
[11.0][IMP] project_timeline: Task timeline template

tarteo:
This needs:

- [x] https://github.com/OCA/web/pull/996
- [x] https://github.com/OCA/web/pull/1003

The project task timeline will look like:

image


pedrobaeza:
It would be great to have planned hours also on the view, and maybe the progress bar

pedrobaeza:
You can rebuild now for having the other PR included

tarteo:
@pedrobaeza I've added the progress bar but to format the planned hours like a float_time widget I need to  make some changes to web_timeline to include like web.field_utils in the rendering of the qweb template or do it before the render.

tarteo:
@pedrobaeza Removed the progress bar because otherwise the module needs a dependency to hr_timesheet.

pedrobaeza:
Ok, we can put a new module with that dependency. About the planned hours, I still consider very important to have it on this view, so the modification on web_timeline is welcome

caiuka:
I agree, it would be a great improvement

tarteo:
@pedrobaeza Here you go.

tarteo:
The progress bar https://github.com/OCA/project/pull/411 (like in https://github.com/OCA/web/pull/996)

pedrobaeza (approved):

pedrobaeza:
Great work!

sergio-teruel (approved):

PR #415, opened 2018-08-08
[MIG] project_issue_timesheet_time_control: Migration script

pedrobaeza:
Fetch the date_time field from old issues and put it on new created tasks.


pedrobaeza:
Wrong model for the field, so closing.

PR #393, opened 2018-06-13
[MIG] project_task_category: Migration to 11.0

igallart:

PR #418, opened 2018-08-13
11.0 mig project task category

devang-dreambits:


pedrobaeza:
There's already another migration in #393. Please review that one instead.

PR #424, opened 2018-09-05
[10.0][ADD] Add sql for migrate the category_id for tasks

JoJoJoJoJoJoJo:
@elicoidal @seb-elico @lonelysun @victormartinelicocorp 

- Add sql to migrate the task category 
- The script should be executed before the module loaded, otherwise there will errors of foreign key constraint for `categ_id` on `project.task`

### DIFF
--- a/project_task_category/migrations/10.0.1.0.0/pre-migration.py
+++ b/project_task_category/migrations/10.0.1.0.0/pre-migration.py
@@ -29,4 +29,12 @@ def migrate(env, version):
                  );
             """
     openupgrade.logged_query(cr, query)
+
+    update_categ_query = """
+        UPDATE project_task task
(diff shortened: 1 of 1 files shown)


project_task_category/migrations/10.0.1.0.0/pre-migration.py
```diff
@@ -29,4 +29,12 @@ def migrate(env, version):
                  );
             """
     openupgrade.logged_query(cr, query)
+
+    update_categ_query = """
+        UPDATE project_task task
+        SET categ_id = categ.id
+        FROM project_category categ, project_category_main categ_main
```
seb-elico, line 36: Use `INNER JOIN`

seb-elico (changes requested):

Issue #422, opened 2018-08-24
[BUG] project_task_dependency: wrong dependencies when duplicating a project


astirpe:
Steps to reproduce:

1. create a project **Project1**
2. Create tasks **Task1A**, **Task1B**, **Task1C** of *Project1*
3. *Task1C* depends on *Task1B* which depends on *Task1A*
4. duplicate project *Project1* and rename the copy as **Project2**

Now *Project2* has 3 tasks: **Task2A**, **Task2B**, **Task2C**.

The problem is that *Task2C* depends on *Task1B*. This is wrong, it should depend on *Task2B* instead.
Also *Task2B* depends on *Task1A*. This is wrong, it should depend on *Task2A*.

Impacted versions: all.

astirpe:
cc @tarteo

PR #425, opened 2018-09-05
[11.0][FIX] Copying dependencies in project_task_dependency

tarteo:
Fixes https://github.com/OCA/project/issues/422

### DIFF
--- a/project_task_dependency/__manifest__.py
+++ b/project_task_dependency/__manifest__.py
@@ -5,6 +5,7 @@
     'name': 'Project Task Dependencies',
     'version': '11.0.1.0.2',
     'category': 'Project',
+    'website': 'https://github.com/OCA/project',
     'summary': 'Enables to define dependencies (other tasks) of a task',
     'author': "Onestein,Odoo Community Association (OCA)",
     'license': 'AGPL-3',
--- a/project_task_dependency/__manifest__.py
+++ b/project_task_dependency/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'Project Task Dependencies',
-    'version': '11.0.1.0.1',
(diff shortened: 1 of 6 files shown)


project_task_dependency/models/project_task.py
```diff
@@ -72,3 +72,16 @@ def _check_dependency_recursion(self):
             raise ValidationError(
                 _('You cannot create recursive dependencies between tasks.')
             )
+
+    @api.multi
+    def copy(self, default=None):
+        self.ensure_one()
+        if self.env.context.get('project_copy'):
+            res = super(ProjectTask, self).copy(default)
```
astirpe: This line could be moved before the `if`. This way you can avoid the `else`.

astirpe (changes requested):

project_task_dependency/models/project_task.py
astirpe: And this way the coverage will be not decreased also :)

elicoidal (approved):
LGTM

astirpe (approved):

project_task_dependency/__manifest__.py
```diff
@@ -3,7 +3,7 @@
```
nikul-serpentcs, line 3: @tarteo IMHO Add website attribute, for reference  [here](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L10)

tarteo: Thanks, didn't notice it.

tarteo:
Can this be merged? :)

nikul-serpentcs (approved):
Code Review LGTM

PR #426, opened 2018-09-06
[11.0][FIX] project_key: Copying / duplicating projects

tarteo:
Currently you can't copy a project because the key is not regenerated. The key must be unique so it throws an error. This PR will prevent the key of being copied to the new project and generate a new key instead.


AaronHForgeFlow (approved):
:+1:

pedrobaeza (approved):

elicoidal (approved):
LGTM

PR #433, opened 2018-09-25
[11.0][IMP] project_key

agajic-modoolar:
This pull request brings two improvements to the project_key module:

1. Enables creating task with existing key, while keeping sequence in valid state.
2. When installing module, tasks in folded stages didn't get key assigned.


project_key/models/project_task.py
```diff
@@ -46,7 +47,13 @@ def create(self, vals):
 
         if project_id:
             project = self.env['project.project'].browse(project_id)
-            vals['key'] = project.get_next_task_key()
+            key = vals.get('key', False)
+            if key:
+                seq = int(key.split("-")[1])
```
alexey-pelykh: Keeping `"` and `'` consistent would be great!

project_key/models/project_task.py
```diff
@@ -46,7 +47,13 @@ def create(self, vals):
...
+                if seq >= project.get_task_key_sequence():
+                    project.set_task_key_sequence(seq + 1)
```
alexey-pelykh: Wouldn't this create holes in numbering sequence?

project_key/models/project_project.py
```diff
@@ -130,6 +130,12 @@ def _prepare_sequence_data(self, init=True):
     def get_next_task_key(self):
         return self.sudo().task_key_sequence_id.next_by_id()
```
alexey-pelykh: Shouldn't this instead check if returned key is valid and generate new ones until proper is found?

alexey-pelykh (changes requested):

rafaelbn:
Please @agajic-modoolar could you review comment of the @alexey-pelykh 's review? Thanks!

project_key/models/project_task.py
agajic-modoolar: Yes, that is good advice. I am going to update it.

project_key/models/project_task.py
agajic-modoolar: It would create gap, but imagine that you have to import tasks that already have a key, since key is what people remember, and refer to, they should not be changed (There is a Project Git module which parses commit message looking for key, and links commit to a task, and changing keys would introduce confusion). By increasing task sequence I am making sure that non of the previously assigned task keys is reused. Those holes are made only when creating tasks with keys, which could happen when importing, so in regular use case this would not happen.

alexey-pelykh: So it's for import-only? Can we somehow detect that? Since there may be the case when I'm manually creating A-1 A-2 and then mistype and create A-20, as a result next in sequence would be A-21

I don't feel that's expected behaviour

project_key/models/project_project.py
alexey-pelykh: @agajic-modoolar I still see no reason if I've created a task A-200 to enforce next in sequence to be A-201 instead of "next unused number"

rafaelbn:
@alexey-pelykh I have found problems in v12 with this modules... This Improvements are needed in v12 too. Regards. It lools like @agajic-modoolar is quite busy.

project_key/models/project_task.py
agajic-modoolar: Why would you ever type a key? It should be generated, and in default view, it's readonly. I only see need for specifying a key when importing or with RPC. But think it this way. If you always generate keys in sequence, and you have A-20, there must have been A-1...A-19 at some point of time. I am thinking of keys as unique ID, just like in database.

agajic-modoolar: I don't see this as a problem really. 

> @alexey-pelykh I have found problems in v12 with this modules... This Improvements are needed in v12 too. Regards. It lools like @agajic-modoolar is quite busy.

I am, indeed :) I like to spend time with my baby :)

agajic-modoolar:
> @alexey-pelykh I have found problems in v12 with this modules... This Improvements are needed in v12 too. Regards. It lools like @agajic-modoolar is quite busy.

I am, indeed :) I like to spend time with my baby :)

project_key/models/project_task.py
alexey-pelykh: Do you think it would be good to make that optional?

project_key/models/project_project.py
```diff
@@ -130,6 +130,12 @@ def _prepare_sequence_data(self, init=True):
...
 
+    def get_task_key_sequence(self):
+        return self.sudo().task_key_sequence_id.number_next_actual
+
+    def set_task_key_sequence(self, next):
```
alexey-pelykh: Do we need these oneliners as separate functions?

project_key/models/project_project.py
pnajman-modoolar: > Shouldn't this instead check if returned key is valid and generate new ones until proper is found?

Interesting approach but the idea here was to have a "straight line" of task numbers as "jumping" numbers would only confuse people. Imagine you have tasks from "A001" to "A010" and someone deletes task "A003" as that task was not needed anymore. After some time a new task is created and instead of assigning key "A011" you get "A003" again and then the person who deleted "A003" see it and deletes it again thinking it's a bug in the system or something ....  I don't think that a good approach, so that's why we are always using the next sequence which is only natural to me.

project_key/models/project_project.py
pnajman-modoolar: @alexey-pelykh Actually, yes we do because this oneliner method enables you to override just that part of the code/logic in your extension module in case you want to have different functionality. For example, functionality you've mentioned above where you'd like to have task keys without "blanks" or implement some completely different logic.

project_key/models/project_task.py
pnajman-modoolar: @alexey-pelykh I agree with your suggestion to make this optional. What do you think about adding some context flag for this?

agajic-modoolar: Can you provide a use case where problem could emerge with this implementation?

project_key/models/project_project.py
alexey-pelykh: That's not how it would work. Assume that:
 1) sequence is at 001,
 2) you create 5 more (when task key is not set)
 3) sequence becomes at 005, 
 4) then you import 5 more tasks, with manual names set to "A007-A011", sequence generator was not used since names were set
 5) next time when you create task as usual, it would propose to use A006 and it's still "unused" 
 6) next time when you create task as usual, it would propose to use A007 and yet it's used, so next one 008...011 until 012 would be generated, and sequence would get updated.

I think these are just two approaches that would be good to have as configurable settings - as this is a behaviour-breaking change.

project_key/models/project_project.py
alexey-pelykh: `set_next_task_key`/`get_next_task_key`?

project_key/models/project_task.py
alexey-pelykh: 1) Not all keys have XXX-1234 format, I think it would be better to consume all digits from end of string, with regexp maybe?
2) This changes current behaviour, without any warning in readme or option to turn it off (maybe make it optional and described?)

project_key/models/project_project.py
pnajman-modoolar: Um, if I understand correctly you are afraid that the sequence "A006" will be generated two times?

If, so, i can assure you that it won't happen because, after you import "A007-A011" the next sequence number will be set to "12" (please check line 53) and then when you next time normally create task it will get key "A012".

alexey-pelykh: We're not on the same page right now, unfortunately. I understand that it would be set to "A012" in the end. I'm just not sure if this is the desired default approach, and if it should be done as a wizard with import options like "regenerate keys" vs "import keys" vs "import keys + shift sequence"

alexey-pelykh: Maybe I may even address that myself in v12

tbaden:
travis warning:
project_key/tests/test_controller.py:6: [W7950(odoo-addons-relative-import), ] Same Odoo module absolute import. You should use relative import with "." instead of "openerp.addons.project_key"

`from ..controllers.main import ProjectBrowser`

instead of 

`from odoo.addons.project_key.controllers.main import ProjectBrowser`

should do the trick

PR #434, opened 2018-09-25
project_stage_closed: Add closed field to project.task class

ernestotejeda:
Cc @Tecnativa

### DIFF
--- a/project_stage_closed/models/__init__.py
+++ b/project_stage_closed/models/__init__.py
@@ -1,2 +1,4 @@
+# Copyright 2018 Tecnativa - Ernesto Tejeda
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from . import task_stage
+from . import project_task
--- a/project_stage_closed/models/__init__.py
+++ b/project_stage_closed/models/__init__.py
@@ -1,4 +1,3 @@
-# Copyright 2018 Tecnativa - Ernesto Tejeda
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from . import task_stage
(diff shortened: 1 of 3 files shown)


project_stage_closed/models/__init__.py
```diff
@@ -1,2 +1,4 @@
+# Copyright 2018 Tecnativa - Ernesto Tejeda
```
astirpe: Please avoid to add the copyright line in the `__init__.py`, as stated in the guidelines

astirpe (changes requested):
Thanks for this PR!

You should update the version in the manifest to `11.0.1.1.0`.

You may also want to add yourself in the contributors list.

ernestotejeda:
Thank you @astirpe

astirpe (approved):

AaronHForgeFlow:
Thank you for the contribution. Why is this field needed?

ernestotejeda:
This field allows us to know directly from another module if a task is closed without having to first ask if it has an associated stage and then ask if the stage is closed. For example, I can get all the tasks that are closed in this way: self.env ['project.task']. Search ([('closed', '=', True)])

pedrobaeza:
@ernestotejeda This is not needed for searches. You can add this domain to the task search:

```python
self.env['project.task'].search([
    ...
    ('stage_id.closed', '=', False),
    ...
]
```

The only real need for this is to want to show something conditionally on task view, so we need a field in that model.

pedrobaeza:
@ernestotejeda I guess this can be closed, isn't it?

ernestotejeda:
Yes, I agree

PR #438, opened 2018-10-05
[12.0][MIG]project_description

Change2improve:

PR #439, opened 2018-10-05
[12.0] [MIG] project_description

Change2improve:


emagdalenaC2i:
As I mentioned in https://github.com/OCA/project/pull/438
Should I install OCA/maintainer-tools and run oca-gen-addon-readme or can I keep the old README.rst file as is mentioned in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#tasks-to-do-in-the-migration ???

emagdalenaC2i:
@nikul-serpentcs I have removed xpath but there are some errors. Am I doing it right?
Here Directly use field, remove xpath.
Ex. <field name="name" position="after">

tafaRU:
> Should I install OCA/maintainer-tools and run oca-gen-addon-readme or can I keep the old README.rst file as is mentioned in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-12.0#tasks-to-do-in-the-migration ???

AFAIK, both methods are valid, but the manual generation is preferable. See https://github.com/OCA/web/pull/963#issuecomment-400344328 as further reference.

tafaRU:
> I have removed xpath but there are some errors.

Try instead the following piece of code:

```
            <field name="name" position="after">
                <field name="description"/>	
            </field>
```

emagdalenaC2i:
@tafaRU @nikul-serpentcs @elicoidal @astirpe @pedrobaeza @Yajo @carlosdauden could you review this PR?

tafaRU (approved):
Code review and test on runbot.

nikul-serpentcs (approved):
Code Review LGTM

project_description/readme/USAGE.rst
```diff
@@ -0,0 +1,14 @@
+To set a description for a project, you need to:
+
+#. Go to *Project > Project > Projects*.
+#. Edit or create one.
+#. You will see a new *Description* tab, with an HTML area to set the project's description.
+
+.. figure:: static/description/image.png
```
astirpe: I think that the path for the image should be absolute (and not relative) because the `gen-addon-readme` will automatically generate the `static/description/index.html`. See https://github.com/OCA/maintainer-tools/pull/370.

So IMO this line should be:

```
.. figure:: https://raw.githubusercontent.com/OCA/project/12.0/project_description/static/description/image.png
```

astirpe (changes requested):

yajo (approved):

nikul-serpentcs:
@tafaRU @Yajo If Done then Move Forward !

project_description/readme/USAGE.rst
emagdalenaC2i: Done

astirpe (approved):

yajo (approved):
:shipit:

PR #442, opened 2018-10-10
[10.0][IMP]project_wbs. Propagate active from the hierarchy

AaronHForgeFlow:
As this module uses the hierarchy proposed by the analytic account it is likely that users set the analytic account to archived instead of the project. In this case the project remains active and can be confusing for project users. It makes sense to close all the projects related when the analytic account is closed.


AaronHForgeFlow:
wrong

PR #444, opened 2018-11-12
12.0 mig project_stage_closed

ernestotejeda:
Cc @Tecnativa


chienandalu (approved):
:+1: Tested on runbot. Thanks!

LoisRForgeFlow (approved):
Tested on runbot :+1:

project_stage_closed/tests/test_project_stage_closed.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class TestProjectStageClosed(common.SavepointCase):
+    @classmethod
+    def setUpClass(cls):
+        super(TestProjectStageClosed, cls).setUpClass()
+        cls.stage = cls.env['project.stage.type'].create({
+            'name': 'Test stage',
+            'closed': True,
+        })
```
nikul-serpentcs: @ernestotejeda here Test Some condition, True Or False

nikul-serpentcs (commented):
Minor Change

AaronHForgeFlow (approved):
:+1: Functional tests.

BTW why is it not called project_task_stage_closed? I know this was like that in previous versions, but it is a bit confusing, it is suggesting we have stages in the project.

pedrobaeza:
@aheficent. Stages are at project level, not task. Previously, when there were issues, the same stages were used. That's why it doesn't have the `_task` surname. Now that they have gone, indeed they are the only place right now where they are used, but any way, we can leave it as is as the stages can be reused in third party modules.

project_stage_closed/tests/test_project_stage_closed.py
pedrobaeza: This is the setup, so no assert should be done here.

pedrobaeza: Ah, OK, I got it. This is a test with no test! @ernestotejeda don't add a test if not needed, like this case.

pedrobaeza (changes requested):
Remove the useless test.

AaronHForgeFlow:
@pedrobaeza Yes, stages may be used in other places but the object is called 'project.task.type'. Anyway it is not blocking IMO.

pedrobaeza:
For that blame Odoo that didn't rename that model in the good moment :wink:

project_stage_closed/README.rst
```diff
@@ -0,0 +1,88 @@
...
+.. contents::
+   :local:
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Have Manager rights for Project group.
+#. Go to *Project > Configuration > Stages*.
```
AaronHForgeFlow, line 41: Is this option still available? I seems to me this menu was removed.

chienandalu: Enable debug mode :wink:

AaronHForgeFlow: I wonder Odoo put all those settings under debug. Changing those settings has nothing to do with debugging...

pedrobaeza: That's why we always install `base_technical_features` and enable it for all users.

chienandalu: That's why OCA's `base_technical_features` exists :slightly_smiling_face:

AaronHForgeFlow: hehe I would put `'auto install' = True` to that module

pedrobaeza (approved):

PR #447, opened 2018-11-12
12.0 mig project_timesheet_time_control

ernestotejeda:
Depend On #444 

Cc @Tecnativa


alexey-pelykh:
There's an alternative implementation in https://github.com/OCA/hr-timesheet/pull/139, it would be great to get reviews there as well

pedrobaeza:
Sorry, Alexei, but @ernestotejeda works for us, and as stated, we are betting on this one, not yours.

alexey-pelykh:
As stated in other thread, modules are not replacements for one another, and since Tecnativa is the author company, fact that you're protecting your version is kinda obvious

pedrobaeza:
Yes, they are replacements, and I don't understand why don't you want to add the missing features. This one has more features. Well, perfect, you don't have to use them, the same as the features you are wanting to add might not be used by the previous users of this module.

alexey-pelykh:
They are not replacement, modules do the same in different manner, apart for having different features - e.g. tracking into multiple tasks at the same time. And the way modules implement same feature is different, thus they are alternative to each other, not a replacement.

pedrobaeza:
Well, I don't agree, as we can always set that behavior through configuration options, not through different modules, but I think this is going to be an endless discussion.

alexey-pelykh:
I guess we can safely agree to disagree 👍

pedrobaeza:
Please rebase

pedrobaeza:
@ernestotejeda please review the alternative proposed and summarizes here pros and cons.

ernestotejeda:
What I have observed about the alternative proposal (hr_timesheet_timetracker):

pros:
* If you start to perform an activity on a date after the date it was inserted in the system, it's convenient to just click on a button (Start) to update this start date instead of editing the date field to insert manually the date field.
* If an activity named A1 already has time registered, and you click on the Start button at a time T2 a new activity is created already started with date T2 and with the same name A1. In this way you can record several times spent in the same activity in a more comfortable way.

cons:
* If this module is installed we will have a behavior for time tracking similar to the one that exists in project_timesheet_time_control (which allows time tracking although in a different way) but nevertheless, since we can not install this last one, we won't be able to use the following features:
  * It improves the domain applied to task field for accommodating to a project
  flow.
  * It adds a datetime field that replaces ``date`` field in tree view, and write
  * Finally, it allows to open and close tasks from account analytic lines.
  The selected closed stage is the first one that is found with the mark
  "Closed" checked.

    taken from (https://github.com/OCA/project/blob/8640848bd00c030e47e729ab51946a7242ea7370/project_timesheet_time_control/README.rst)

* The system calculates the time from start to stop, but not from the start date established by the user to stop, so if the user doesn't click Start button on time, he will not be able to calculate the time used automatically, since that the system does not do this automatic calculation only giving stop. The user will have to estimate the time according to his perception to enter it into the system.
* The project_timesheet_time_control module already existed for a similar purpose of time tracking since Odoo v9, so there are users who use it and recommend it.

pedrobaeza:
Thanks for the analysis, Ernesto. I think we will continue with this one for having all integrated. What I think it's good from the other PR is the pause button, which seems to be very appreciated by timesheet users, but that one goes against the way this is designed. We would need an extra table and for sure add the possibility to set manually the starting datetime. Do you think it's worth? Any other idea to implement this feature? Maybe we can create a new record from the existing one when we "restart" the job.

alexey-pelykh:
@ernestotejeda thanks for detailed review. Please let me address cons with following comments:
Since `project_timesheet_time_control` and `hr_timesheet_timetracker` should not be used simultaneously, yet there's a userbase of `project_timesheet_time_control` that would need specific outstanding features, these features have to be addressed somehow.

> It improves the domain applied to task field for accommodating to a project
flow.

This is definitely an important usability enhancement, which should be kept, though since it's not directly related either to time-tracking or task-lifecycle-control, it's an open question where/how to bundle it.

> It adds a datetime field that replaces `date` field in tree view, and write
date field with `datetime` field value.

I'm not sure how important it is to actually see hours, minutes, and seconds when task tracking has been started especially in the treeview w/o over-loading the UI, though sorting order is important. Yet the sorting order is taken care of in `hr_timesheet_timetracker` as well.

> Finally, it allows to open and close tasks from account analytic lines.

This is the most important feature I think, but it's 100% not related to actual time tracking. Yet I think it would be great to extract/leave this feature in a module that will extend `hr_timesheet_timetracker` to do the same.

> The system calculates the time from start to stop, but not from the start date established by the user to stop, so if the user doesn't click Start button on time

Not sure I got you correctly, `hr_timesheet_timetracker` allows setting/editing the start and stop timestamps exactly in order to allow user to adjust mis-started or mis-stopped entries. Please elaborate on this con.

> The project_timesheet_time_control module already existed for a similar purpose of time tracking since Odoo v9, so there are users who use it and recommend it.

Well, there are users who liked approach of `hr_timesheet_timetracker` as well, maybe not that many yet that's not a good point to compare in pros/cons IMHO.

@pedrobaeza from the review above I'd conclude that there are:
 * 1 outstanding feature - which should've been separate from time-tracking,
 * 1 important UX enhancement - semi-related to time-tracking since takes closed/not-closed filter on task
 * 1 (IMHO) opinionated enhancement, that's related on date or datetime is shown to user.

I'm open to walk an extra mile and implement first two missing points in a separate module, that would live in `project` since is task-related.

rafaelbn:
Dear @alexey-pelykh @ernestotejeda ,

I like this debate. This is my view of the situation.

- We develop as Open Source this module "project_timesheet_control" from OpenERP 7. This is the 6th year with this module in enough Odoo instance so we cannot discard it of course.
- @alexey-pelykh from scratch made a new nice work! https://github.com/OCA/hr-timesheet/pull/139 . Desired of course it would be to have this debate before de work :smile: 
- Finally all this stuff if done for customer which nice this functionalities like ourselves! :thinking: so the scope and objetive is making user easy and fast when making input of times on Task, Timesheets, Leads, etc... because the better solution is Enterprise with plugins like https://chrome.google.com/webstore/detail/awesome-timesheet/jepmfegpglmmpcldolgpnfocekggingc

My questions is easy to ask difficult to make :smiley: 

- How to take the best of both?

Let's talk and make this possible :+1:  for all OCA!

alexey-pelykh:
@rafaelbn from review by @ernestotejeda above I can conclude that there are 3 things that `hr_timesheet_timetracker` lacks in order to be used w/o loosing benefits of `project_timesheet_time_control` while having more advanced _time-tracking_ features. Of course, in order to compete with Toggl/Harvest/Enterprise-Odoo, an app or browser plugin would be great, yet it's far later in our roadmap. Regarding the 3 features above - they are not timetracking-related, but task/project-related, and as proposed above - may be reimplemented as complimentary module in order to provide similar user experience for `project_timesheet_time_control` users.

alexey-pelykh:
As promised, I'd like to present following bundle of pull-requests that are intended to provide smooth substitution:
 * `hr_timesheet_timetracker` (https://github.com/OCA/hr-timesheet/pull/139): actual timetracking module
 * `hr_timesheet_task_stage` (https://github.com/OCA/hr-timesheet/pull/179): complimentary module with specific feature from original `project_timesheet_time_control`
 * `hr_timesheet_task_domain` (https://github.com/OCA/hr-timesheet/pull/180): complimentary module with specific feature from original `project_timesheet_time_control`
 * `project_timesheet_time_control` (#474): data migration module

emagdalenaC2i:
Finally, what option will be followed?
Should I evaluate this PR, should I evaluate https://github.com/OCA/hr-timesheet/pull/139 or both?

alexey-pelykh:
@emagdalenaC2i right now plan is: @pedrobaeza to take a look on those 4 modules and decide from there. `project_timesheet_time_control` is more used in v11 and before, `hr_timesheet_timetracker` is more advanced for time tracking and is for v12 and above.

pedrobaeza:
@emagdalenaC2i please evaluate both and give your opinion. The question for deciding includes:

* Migration path
* Code maintenance.

apart from pure features.

rafaelbn:
Hi @emagdalenaC2i so what do you think?

We have been using this module from 5 years in previous version so we need opinions

rafaelbn (approved):
Functionally :+1:

emagdalenaC2i (approved):

emagdalenaC2i:
@rafaelbn as you said, this module has been used for a long time. So I think this should be merged.
I will check OCA/timesheet#139 but this one should go ahead

alexey-pelykh:
@emagdalenaC2i what's the point of merging one prior to reviewing another if idea was to evaluate both first?

yajo (approved):

rafaelbn:
Please @ernestotejeda set module as MATURE in readme. Thanks

alexey-pelykh:
Not sure if this was taken care of: some features of this module was extracted into separate https://github.com/OCA/timesheet/pull/180 and https://github.com/OCA/timesheet/pull/179 to avoid forcing users to have time control when all they need is embedded feature.

pedrobaeza:
@ernestotejeda for putting that development status, you need to declare a maintainer. As you have been the migrator of last 2 versions, you can put yourself:  `'maintainers': ['ernestotejeda'],`.

But before merging, please rip off the features present in hr_timesheet_task_domain and hr_timesheet_task_stage and depend on them.

pedrobaeza:
Add installation section on README for talking about dependencies on OCA/timesheet (This module depends on ... that are in OCA/timesheet...), and regenerate README.

ernestotejeda:
Changes done

pedrobaeza (approved):

PR #448, opened 2018-11-13
project_timesheet_time_control: Change 'account analytic line' default order

ernestotejeda:
Cc @Tecnativa

### DIFF
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -9,6 +9,7 @@ from datetime import datetime
 
 class AccountAnalyticLine(models.Model):
     _inherit = 'account.analytic.line'
+    _order = 'date_time desc'
 
     date_time = fields.Datetime(default=fields.Datetime.now, string='Date')
     closed = fields.Boolean(related='task_id.stage_id.closed', readonly=True)
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -9,7 +9,7 @@ from datetime import datetime
 
 class AccountAnalyticLine(models.Model):
     _inherit = 'account.analytic.line'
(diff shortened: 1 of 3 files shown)


project_timesheet_time_control/views/project_task_view.xml
```diff
@@ -93,6 +93,9 @@
                     </group>
                 </form>
             </xpath>
+            <xpath expr="//field[@name='timesheet_ids']/tree" position="attributes">
+                <attribute name="default_order">date_time</attribute>
```
pedrobaeza, line 97: Why adding this?

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -9,6 +9,7 @@
 
 class AccountAnalyticLine(models.Model):
     _inherit = 'account.analytic.line'
+    _order = 'date_time desc'
```
pedrobaeza: Only this field? Shouldn't be more?

project_timesheet_time_control/views/project_task_view.xml
ernestotejeda: The "timesheet_ids" field in the Tasks view form has default_order = "date" specified, so i write that to order timesheets by date_time field insted.

project_timesheet_time_control/models/account_analytic_line.py
ernestotejeda: The only field that I have seen that the account.analytic.line is ordered for is 'date' field field, therefore, I have only set it.

pedrobaeza: But what about `id`? I think we should add it

pedrobaeza (approved):

carlosdauden (approved):

chienandalu (approved):

cristinamartinrod (approved):
Tested on Runbot 👍🏻

PR #449, opened 2018-11-17
[ADD] project_hr: Link HR with project

pedrobaeza:
Project HR
==========

This is a technical module for adding some HR information to projects and
provide technical fields with HR information:

- Employee categories at project and task level.
- Employee linked to the user assigned to a task.
- Employee categories at user level.

These fields can be used later for extra features like task scheduling or
security restrictions.

It also provides some facilities on project management:

* Limit selection of assigned users according selected employee category at
  task level.
* Limit selection of employee categories at task level depending on the
  selected categories at project level. All are shown if no categories at
  project level.

Configuration
=============

For setting employee categories at project level:

1. Go to *Project > Dashboard*.
2. Click on the 3 vertical dots for unfolding options and select "Settings".
3. Put the wanted employee categories on the field "Employee Categories".

For setting employee categories:

1. Go to *Project > Search > Tasks*.
2. Select or create a new task.
3. Put the wanted employee category on the field "Employee category".
4. If there's already some employee categories selected at project level, those
   will be the only selectable ones in the task.

Usage
=====

1. Go to *Project > Search > Tasks*.
2. If there's an employee category selected in the task, you will only be able
   to select those users whose employee belongs to that category.

Known issues / Roadmap
======================

* Limit project and task visibility according employee categories through
  overriding security methods (as through record rules is not possible without
  modifying existing ones).

PR #419, opened 2018-08-16
project_task_scheduling: Module for automatic task planning

ernestotejeda:
Cc @Tecnativa

PR #452, opened 2018-11-20
[12.0][ADD] project_role: role-based project roster

alexey-pelykh:
image


alexey-pelykh:
@pedrobaeza once your wonderful `project_hr` (#449) will get into 12.0, I'll add a `project_hr_role` to connect tags with roles in order to limit who can be assigned to a role. Meanwhile `sale_timesheet_role` is something to be done, yet please advise to which repository it should go (`project`? `sale-workflow`?)?

alexey-pelykh:
@nikul-serpentcs @pnajman-modoolar @aheficent it would be great to get your opinion on the matter, if you don't mind

nikul-serpentcs (approved):
Code Review LGTM

alexey-pelykh:
Added extra capabilities to review assignments from various retrospectives:
image

alexey-pelykh:
Inviting more good people: @tarteo @astirpe - please review, if you don't mind

alexey-pelykh:
Updated PR to assign employees (not users) to roles

project_role/models/project_assignment.py
```diff
@@ -0,0 +1,69 @@
...
+    project_id = fields.Many2one(
+        comodel_name='project.project',
+        string='Project',
+        required=True,
+    )
+    role_id = fields.Many2one(
+        comodel_name='project.role',
+        string='Role',
+        required=True,
+        track_visibility='onchange',
```
bodedra, line 35: IMO we need to `_inherit = ['mail.thread']` so we can use `track_visibility` attributes.

alexey-pelykh: @bodedra as far as I know, it's not required to use email for that, since this changes are tracked in the history that is not emailed by default. Did I get you right?

bodedra: @alexey-pelykh I meant with chatter section to track field.

alexey-pelykh: @bodedra got it! I'd appreciate a reference to another module that behaves like you expect 👍

bodedra: @alexey-pelykh Here is reference from _sale_ module. https://github.com/odoo/odoo/blob/12.0/addons/sale/models/sale.py#L22 and https://github.com/odoo/odoo/blob/12.0/addons/sale/views/sale_views.xml#L575

In our case, we need things in .py:

    _inherit = ['mail.thread']

and .xml

    <div class="oe_chatter">
        <field name="message_follower_ids" widget="mail_followers"/>
        <field name="message_ids" widget="mail_thread"/>
    </div>

So whenever _Role_ field update, Odoo will update chatter section with old value -> new value.

bodedra: Or You may use https://github.com/bodedra/account-financial-tools/commit/b62c6cacfc4f115888a0cee7ed899258160da021 as a reference which I have recently migrated module.

alexey-pelykh: @bodedra thanks for a hint, implemented!

alexey-pelykh:
@bjeficent if you don't mind, I'd like to request your review as well

bjeficent (approved):
LGTM :+1:

bodedra:
@alexey-pelykh minor one: would you please check run bot id ? I have tried to find module on run bot but couldn’t find it. I have use following as reference Runbot:

https://github.com/OCA/project/pull/452/files#diff-53be047a2aeb8a118f8091df0bb2e1aaR23

bodedra (approved):
Good job @alexey-pelykh  :+1:

xiaoyuan0017:
what

Bhavesh Odedra <notifications@github.com> 于 2018年11月28日 周三 19:47写道：

alexey-pelykh:
@bodedra re: runbot - frankly speaking, I have no idea how it works with PRs. That particular line was generated by `oca-gen-addon-readme --repo-name project --branch 12.0 --addon-dir project_role` so maybe that's due to branch specification?

bodedra:
@pedrobaeza would you please help us here?

https://github.com/OCA/project/pull/452#issuecomment-442419172

project_role/README.rst
```diff
@@ -0,0 +1,74 @@
+=============
```
tarteo, line 1: Can you explain how to configure and use the module in readme, please.

tarteo (changes requested):

project_role/README.rst
alexey-pelykh: @tarteo thanks for feedback, added _Usage_ section,

tarteo (approved):
Tested it, looks good!

pedrobaeza:
What I don't like of this module is that you are mixing employees and projects seamlessly. Employees are not users, and the reverse is also true. You can have an external contractor that you give a user (portal or backend, it doesn't matter), and you want to give roles, assign to a task, etc, and that's why Odoo uses all time `res.users`, and not `hr.employee`.

This module should respect this for having users belonging to roles, not employees.

alexey-pelykh:
@pedrobaeza I got your point, it was implemented like that originally, yet using `res.users` assumes that person assigned to a role has a corresponding user - which is not necessarily a case. Frankly speaking, I understand pros & cons of my approach, I understand your position - and I don't see how to mix those :|

alexey-pelykh:
@tarteo @bodedra @nikul-serpentcs @bjeficent what are your thoughts on the matter?

bodedra:
@alexey-pelykh I have just review your code previously. IMO you can better explanation/purpose of these module. You may add some points or test case on end user side, like what happened with and without these module.

alexey-pelykh:
This module appeared to provide ability to solve following:
 * possibility to limit who can enter timesheets time towards the project (https://github.com/OCA/hr-timesheet/pull/140)
 * build reports who is assigned to which project (not necessarily having tasks assigned within the project)
 * ability to add pre-/post-assignment and unassignment reactions (e.g. modify LDAP group membership, etc.)

The topic in question is: who should be an assignee - a user (`res.users`), an employee (`hr.employee`), or a partner (`res.partner`).

Non-employees not necessarily have user accounts, or should have user accounts.
On other hand, even when this person is not Employee, it can be either:
* customer representative (has Public user probably),
* third-party outstaffed employee (has Employee probably, and has user),
* representative of third-party company (who also works on the project yet is not obliged to use ERP, thus has only Partner entry probably).

If we would use a `res.user` in this module, it would cover only 2 of 3 cases (with a requirement for every employee to have a user set via Related User, which I don't really like).
If we would use a `hr.employee`, it would cover only 1 of 3 cases.
Using `res.partner` would allow to cover all 3 cases, by computing `user_id` from the partner and computing `hr.employee` from `user_id`.

@pedrobaeza ^

alexey-pelykh:
There's also another candidate: `resource.resource` - it has a direct relation to user and employee, yet will not fit for case 3 described above

alexey-pelykh:
After thinking a lot about this I'm leaning towards or `res.users` or `res.partner`, preferably latter. @pedrobaeza your thoughts?

bodedra:
@alexey-pelykh Thanks for your details information and elaborate cases. IMO `res.users` seems good to me. So later on, we can use existing Odoo access rights.

Regarding your third case, just create user and don't provide login credentials. That's it.

alexey-pelykh:
@bodedra it sounds a little bit as a workaround, especially if to take into account that every user has a relevant `res.partner`. Yet I'm open to this, just want to get as much feedback as possible

pedrobaeza:
IMO you should go to `res.users` and not `res.partner`. Why? Because:

* You can tweak permissions for that element as said by @bodedra.
* Any partner can easily be a portal user whenever you want from portal wizard.
* Not all partners are automatically selectable from the beginning, so you keep your "possible users" list under control.
* You can use all the portal benefits.

tarteo:
If I just have to choose between res.users, res.partner, or hr.employee. I would choose res.users it's consistent with the rest of the system and compatible with portal as @pedrobaeza mentioned.
Using employee won't affect changing permissions. as you can simply do `employee_ids.user_id` in the domain.

Here employee link is created to use the work schedule https://github.com/OCA/project/pull/419/files For this module employee information is not needed.

alexey-pelykh:
Thanks all for the feedback, I'll update the module today

pedrobaeza:
Note that #419 will change to be adapted to #449 soon (but it's basically the same, but less intrusive).

tarteo:
@pedrobaeza Nice.

alexey-pelykh:
@tarteo @pedrobaeza @bodedra thanks again for valuable feedback, module implementation changed to `res.users` + adjusted to have hooks for further extensions.

alexey-pelykh:
Should we proceed to the magic of merging?

alexey-pelykh:
@pedrobaeza ?

project_role/README.rst
```diff
@@ -0,0 +1,84 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+To configure the list of roles available go to
```
pedrobaeza, line 39: This should be in configuration section

project_role/models/project_assignment.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @api.multi
+    def unlink(self):
+        self._assignment_preremove()
+        values = self.read()
+        result = super().unlink()
+        self._assignment_postremove(values)
+        return result
+
+    def _assignment_preprocess(self, values):
```
pedrobaeza: I don't think this is needed. You can inherit `write` and `create` as well in your extension module.

project_role/models/project_assignment.py
```diff
@@ -0,0 +1,117 @@
...
+        """ Extension point to execute additional actions before assignment
+            has been changed.
+        """
+        return values
+
+    def _assignment_postprocess(self, values):
```
pedrobaeza: The same for this

project_role/models/project_assignment.py
```diff
@@ -0,0 +1,117 @@
...
+        """ Extension point to execute additional actions after assignment
+            has been changed.
+        """
+        pass
+
+    def _assignment_preremove(self):
```
pedrobaeza: Same

project_role/models/project_assignment.py
```diff
@@ -0,0 +1,117 @@
...
+        """ Extension point to execute additional actions before assignment
+            has been removed.
+        """
+        pass
+
+    def _assignment_postremove(self, values):
```
pedrobaeza: Same

project_role/models/project_assignment.py
```diff
@@ -0,0 +1,117 @@
...
+    _sql_constraints = [
+        (
+            'project_role_user_uniq',
+            'unique (project_id, role_id, user_id)',
+            'User may be assigned per role only once within a project!'
+        ),
+    ]
+
+    @api.depends('project_id.name', 'role_id.name', 'user_id.name')
+    def _compute_name(self):
```
pedrobaeza, line 53: This is usually done with `display_name` instead of `name`.

project_role/models/project_assignment.py
```diff
@@ -0,0 +1,117 @@
+# Copyright 2018 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models, _
+from odoo.exceptions import ValidationError
+
+
+class ProjectAssignment(models.Model):
```
pedrobaeza, line 8: As this is a one2many that is accessed inside project, the guideline is to put this code in `project_project.py` file instead of an isolated file.

project_role/models/project_role.py
```diff
@@ -0,0 +1,39 @@
...
+        required=True,
+    )
+    description = fields.Html(
+        string='Description',
+        translate=html_translate,
+    )
+    company_id = fields.Many2one(
+        comodel_name='res.company',
+        string='Company',
+        required=True,
```
pedrobaeza: This shouldn't be required if you want the same role across several companies, and it has all the sense for multi-company users.

pedrobaeza: You have to add as well a record rule for multi-company restrictions.

project_role/models/res_company.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2018 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = 'res.company'
+
+    project_role_ids = fields.One2many(
```
pedrobaeza: This doesn't serve for multi-company roles, so solutions are:

1. Do this a compute field for only information purposes, and put a reference to where to remove this.
2. Remove the field at all.
3. Put a compute + inverse, feeding the other table property. My only fear with this is that you are unaware of the multi-company feature from this view, and then duplicate roles across companies.

project_role/models/res_config_settings.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2018 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    project_role_ids = fields.One2many(
```
pedrobaeza: I think this shouldn't be here. You already have roles in other menu entry.

pedrobaeza (changes requested):

project_role/models/project_assignment.py
alexey-pelykh: Did some research on this, it seems that `display_name` is only used when model has it's own assigned `name`, which is not the case here, hence there's no difference between `name` and `display_name`

project_role/models/project_assignment.py
alexey-pelykh: Ok! If possible, please refer to the relevant guideline via link

project_role/models/project_assignment.py
alexey-pelykh: It's not _needed_, yet similar approach has been used in https://github.com/odoo/odoo/blob/97aec9f836bf4154e95fa9857cef5e31cbad5d00/addons/hr_timesheet/models/hr_timesheet.py#L134 and I think that it's great to explicitly point to extension points that are by-design.

project_role/models/project_assignment.py
pedrobaeza: OK, we can keep it as is.

project_role/models/project_assignment.py
pedrobaeza: It's not the same, Alexey. There you already have code, and it has been unplugged from the method to have the opportunity to be disabled or overrided. Here, you don't have any code.

project_role/models/project_assignment.py
pedrobaeza: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#15file-naming

AaronHForgeFlow (commented):
I'd like the user_id field in the project.role to be not required. I'd like the idea of having different external roles. Those roles are not linked to a user or an employee, but in the project can define all the roles that take part of the project, the stakeholders of the project.

project_role/models/project_assignment.py
alexey-pelykh: To be exact, this is not required, right?
> An exception *can be* made when the model is a model intended to be used only as a one2many model nested on the main model. In this case, you can include the model definition inside it. Example sale.order.line model *can be* together with sale.order in the file models/sale_order.py.

project_role/models/project_assignment.py
alexey-pelykh: Got your point, have to think about it a bit more. Also, @aheficent since you have some extension in mind, how this impacts your idea?

project_role/models/project_assignment.py
pedrobaeza: Well, any of the guideline can be, but it's convenient to be homogeneous.

project_role/models/project_role.py
alexey-pelykh: Re: shouldn't be required - you mean to unlink it from company? What are the common usecases for multi-company setups?

Re: record rule - I'm not 100% sure what exactly you mean, a link to something similar would be much appreciated.

project_role/models/res_company.py
alexey-pelykh: The original idea was to keep roles "owned" by companies, so this should be addressed if that would change

project_role/models/res_config_settings.py
alexey-pelykh: This is a purely usability enhancement, implemented with warning that roles are configured per-company

alexey-pelykh:
@aheficent you mean in `project.assignment`? There's no point in having assignment without someone being actually assigned. Please elaborate on your usecase

AaronHForgeFlow:
There is a point, because some workers may not be registered as users in the system, they just do their job, that happens frequently in construction companies. Even maybe there's some externals working on a project. I will not block this PR, it is just a proposal.

Edit: However I understand that in they are not registered as a user we just don't create the assignment and that's it.

pedrobaeza:
@aheficent we have already discussed about employee vs user and there's no problem in creating users without effective access to the system (for example, being portal users).

project_role/models/res_config_settings.py
pedrobaeza: The problem is this makes even slower the load of the settings, which is now very bloated loading everything for all the settings in any area. If you want to have this usability improvement, just put a link to the action as you can see for example for the pricelists when you select the option of multiple prices:

seleccion_060

project_role/models/res_company.py
pedrobaeza: Yes, that's why I'm proposing that solutions. I still fee that this is not important to be put here on the company form. At the end, normal project managers don't have access to this form, and this is not useful for system admins.

project_role/models/project_assignment.py
alexey-pelykh: In this particular case I'll surely give it a lot of consideration for some time, yet as of right now I feel it's better to opt in "not using that exception".

project_role/models/project_role.py
pedrobaeza: Yes, if a role doesn't have any company filled, it's available for all companies. It's the normal use for a lot of objects in Odoo. See any of the multi-company rules for the core objects:

seleccion_061

project_role/models/project_assignment.py
pedrobaeza: OK, I won't block for this.

project_role/models/res_config_settings.py
alexey-pelykh: Understood, will adjust this. In UI it actually is like that:
image

project_role/models/project_role.py
alexey-pelykh: Oh dear, you force me to learn some dark voodoo SQL magic for that :)))

alexey-pelykh:
@pedrobaeza please re-review

pedrobaeza (approved):

rafaelbn:
Hello @alexey-pelykh after talking deeply with @pedrobaeza here (sorry) you were right in https://github.com/OCA/project/pull/452#issue-232212007

- This module should be name project_employee_role
- And yes, in Odoo core you can make timesheets without having a res.user because the mandatory is having an employee with its Timesheet Cost...

Are we on time to correct this?

This happens after our Google Meet reviewing sale_timesheet_enhance and reviewing all more deeper to go ahead and help you.

Note: if we don't change this the we are forcing to customer to have always a res.user per employee to use this modules which is not an optimal situation if the customer use an Odoo Enterprise (cost per user).

Note: if we don't change this at least the module must be rename as project_user_role

Important: all my comment is obviously not mandatory is my PoV after 4 hours reviewing several modules of @alexey-pelykh

rafaelbn:
Hello @alexey-pelykh after talking deeply with @pedrobaeza here (sorry) you were right in https://github.com/OCA/project/pull/452#issue-232212007

- This module should be name project_employee_role
- And yes, in Odoo core you can make timesheets without having a res.user because the mandatory is having an employee with its Timesheet Cost...

Are we on time to correct this?

This happens after our Google Meet reviewing sale_timesheet_enhance and reviewing all more deeper to go ahead and help you.

Note: if we don't change this the we are forcing to customer to have always a res.user per employee to use this modules which is not an optimal situation if the customer use an Odoo Enterprise (cost per user).

Note: if we don't change this at least the module must be rename as project_user_role

Important: all my comment is obviously not mandatory is my PoV after 4 hours reviewing several modules of @alexey-pelykh

rafaelbn:
@victormmtorres this is the first module.

alexey-pelykh:
@rafaelbn unfortunately this one already merged since early December, not sure what's the policy for renaming - I think only in v13?

rafaelbn:
@pedrobaeza how to manage this change user to employee and this renaming?

pedrobaeza:
We can have both modules now in v12 and deprecate the other. For existing installations, a migration script can be develop for moving from one module to the other.

alexey-pelykh:
I gave this some thought, and in my opinion it's good that we've proceeded with `res.users` instead of `hr.employee`, since in project module (and this is project-related module), tasks are being assigned to users, not employees, and similarly - role on the project should be assigned to user, not employee.

As not all potential assignees are employees. Potentially, there can be a module that restricts who can be assigned to tasks by role, etc.

Summarizing, I'm pretty sure that we should keep things as-is because that's the proper way.

PR #454, opened 2018-11-26
[12.0][IMP] project_key: do not regenerate project key if it's already set

alexey-pelykh:
In case project key is already generated or set manually, do not overwrite it


pedrobaeza (approved):

bodedra (approved):

nikul-serpentcs (approved):
LGTM

PR #472, opened 2019-02-01
[IMP][11.0][project_task_material_stock] Improve the workflow to consume materials in project task using a stock picking

jcarlosmontoya:
This PR Improve the workflow to consume materials in project task using a stock picking
This PR can be improved for example:
* Add a new special stock picking type for stock picking


pedrobaeza:
Please check Travis

project_task_material_stock/views/project_task_view.xml
```diff
@@ -48,6 +48,7 @@
                             </field>
                         </page>
                         <page string="Stock Moves" name="stock_moves">
+                            <field name="picking_id" readonly="1"/>
                             <field name="stock_move_ids"/>
```
cubells: I think is more usable if you create a tree like this:
```
<field name="stock_move_ids">
    <tree string="Stock Moves" editable="top">
        <field name="picking_id"/>
        <field name="date_expected"/>
        <field name="date"/>
        <field name="origin"/>
        <field name="product_id"/>
        <field name="product_uom_qty"/>
        <field name="state"/>
    </tree>
</field>
```  

And then is not necessary a picking_id field in task.

project_task_material_stock/models/project_task.py
```diff
@@ -177,10 +183,29 @@ def _prepare_stock_move(self):
 
     @api.multi
     def create_stock_move(self):
-        for line in self:
-            move_id = self.env['stock.move'].create(
-                line._prepare_stock_move())
-            line.stock_move_id = move_id.id
+        pick_type = self.env.ref('stock.picking_type_out')
+
+        if self:
```
cubells: This if is not necessary. If you arrive here if because self is set.

project_task_material_stock/models/project_task.py
```diff
@@ -177,10 +183,29 @@ def _prepare_stock_move(self):
...
+            picking_id = task.picking_id or self.env['stock.picking'].create({
+                'origin': "{}/{}".format(task.project_id.name, task.name),
+                'partner_id': task.partner_id.id,
+                'picking_type_id': pick_type.id,
+                'location_id': location_id,
+                'location_dest_id': location_dest_id,
+            })
+
+            for line in self:
+                if not line.stock_move_id:
```
cubells: I think is better to create a dict and fill it with all the data and then create a picking once, adding all dict data in a line on creating the picking as is shown:
```
...
'move_lines': [(0, 0, dict_with_move_date)]
...
```

project_task_material_stock/models/project_task.py
```diff
@@ -177,10 +183,29 @@ def _prepare_stock_move(self):
...
+        pick_type = self.env.ref('stock.picking_type_out')
+
+        if self:
+            task = self[0].task_id
+            location_id = task.location_source_id.id or task.project_id.location_source_id.id or self.env.ref(
+                'stock.stock_location_stock').id
+            location_dest_id = task.location_dest_id.id or task.project_id.location_dest_id.id or self.env.ref(
+                'stock.stock_location_customers').id
+
+            picking_id = task.picking_id or self.env['stock.picking'].create({
```
cubells: picking_id is not necessary if you follow my instructions bellow.

cubells (changes requested):

cubells:
@jcarlosmontoya review travis

project_task_material_stock/models/project_task.py
```diff
@@ -177,10 +183,29 @@ def _prepare_stock_move(self):
...
     def create_stock_move(self):
-        for line in self:
-            move_id = self.env['stock.move'].create(
-                line._prepare_stock_move())
-            line.stock_move_id = move_id.id
+        pick_type = self.env.ref('stock.picking_type_out')
+
+        if self:
+            task = self[0].task_id
+            location_id = task.location_source_id.id or task.project_id.location_source_id.id or self.env.ref(
```
nikul-serpentcs: FYI Here we direct Get Id: `self.ref('stock.stock_location_stock')`

project_task_material_stock/models/project_task.py
jcarlosmontoya: @cubells thank you, but when delete items, self is empty and crash all

project_task_material_stock/models/project_task.py
jcarlosmontoya: @nikul-serpentcs, stock location by default?, if you set an Especific location in project or task?
I think that picking type should be a new, new sequence, etc. (separate them from standard pick types)
What do you think?

project_task_material_stock/models/project_task.py
jcarlosmontoya: if you need add a new material in task? picking id it's necessary?
I need you  help in this point. We should be able to work with partial materials in tasks

project_task_material_stock/views/project_task_view.xml
jcarlosmontoya: Perfect!

jcarlosmontoya:
> @jcarlosmontoya review travis

What happen with Travis please?

jcarlosmontoya:
@pedrobaeza @cubells @nikul-serpentcs  check new changes please

project_task_material_stock/models/project_task.py
nikul-serpentcs: @jcarlosmontoya If you used this  `self.env.ref('stock.stock_location_stock')` then better to get direct ID like `self.ref('stock.stock_location_stock')`.

Not blocking but improvement

jcarlosmontoya: @nikul-serpentcs thank you for your tip!

project_task_material_stock/views/project_task_view.xml
jcarlosmontoya: Done!

project_task_material_stock/models/project_task.py
jcarlosmontoya: Done!, updated

project_task_material_stock/models/project_task.py
jcarlosmontoya: Updated

cubells (approved):
:+1: 

LGTM

Code review and test

nikul-serpentcs (approved):
Code Review LGTM :+1:

pedrobaeza:
3 people added as contributors for this PR? It seems too much, and also the commit history is a mess, with merge operations and so on. Please cleanup it.

jcarlosmontoya:
@pedrobaeza Ok Go ahead

PR #473, opened 2019-02-11
[12.0][FIX] project_task_code: invalid variable

alexey-pelykh:


project_task_code/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     'name': 'Sequential Code for Tasks',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
```
nikul-serpentcs, line 6: @alexey-pelykh No need to Update version, it's Code Error And you DIrectly Fixed that.

It's Improvement Not added Any functionality in Exit code.

nikul-serpentcs:
@pedrobaeza Can you please Look and Fast Tracking

project_task_code/__manifest__.py
alexey-pelykh: It's a bug fix for published module, version bump is required

pedrobaeza (approved):

PR #474, opened 2019-02-12
[12.0][MIG] project_timesheet_time_control - alternative

alexey-pelykh:
https://github.com/OCA/hr-timesheet/pull/179, https://github.com/OCA/hr-timesheet/pull/180, and this one should provide a seamless drop-in substitution to #447. Also depends on https://github.com/OCA/hr-timesheet/pull/139


project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -0,0 +1,15 @@
...
+# Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    # defines start date, start date + unit amount = end date
```
alexey-pelykh: @pedrobaeza I'll need your help over here

yajo (changes requested):
This module deletes the previous implementation to just add a pre-migration script. AFAICS, you should install the module in an updated database and then uninstall it once migration is passed.

If this one were the path to follow, doing it in a module makes no sense. It should be done in OpenUpgrade instead.

Sorry, :man_shrugging: but :-1:, IMHO this can be closed.

alexey-pelykh:
@Yajo thanks for the feedback! Probably OpenUpgrade is more proper way

yajo:
Can you close this then, please? I don't have permissions.

Thanks for your understanding! :wink:

alexey-pelykh:
Once there will be a code to reference as "superseded by" - sure

PR #478, opened 2019-02-28
[11.0][ADD] project_document_page

LoisRForgeFlow:
This module allow to link document pages to projects.


pedrobaeza:
Shouldn't be `document_page_project`?

LoisRForgeFlow:
why? main focus here is to extend project functionalities.

pedrobaeza:
In my mindset, you are extending `document_page` for covering projects.

pedrobaeza:
I would even put this in OCA/knowledge, not here.

LoisRForgeFlow:
Yes, both ways of thinking it are valid, and, I swear you I thought about it before putting the name, but I wasn't able to come up with a strong point for any of the options so I end up using project first because it was the trigger for me, I want the pages in my projects. I mean, do you really think this needs to be moved?

pedrobaeza:
My reasoning for going to OCA/knowledge and for renaming of here is the same, but can be unfolded with other sub-reasins: 

- To avoid cross dependencies between repositories.
- To find everything related in the same place.
- To have a correlative names that makes you find everything that use "pages" with the name `document_page_*`.
- I always take the module that defines the main model as reference. In this case, as everything is related to the `document.page` model, that's why I say to take that one as start.

Obviously, all of this is subjective, as there's no strict rule for this.

LoisRForgeFlow:
Apart from the first point, all the other can be argued in favor of project (everything related to project in same place, search stuff related to project with `project_*`, main model `project.project`). But, anyway, it is not a big deal for me so meved to: https://github.com/OCA/knowledge/pull/206

pedrobaeza:
Thanks, I didn't want to impose my criteria, but I appreciate that you consider it.

PR #480, opened 2019-03-02
[imp] Highlights the tasks with date_end > date_deadline

karlis-dreizis:
Helps the user see, straight from the timeline view, whether his/hers
set end_date meets the set date_deadline.

Depends on OCA/web#1206

### DIFF
--- a/project_timeline/__manifest__.py
+++ b/project_timeline/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Project timeline",
     "summary": "Timeline view for projects",
-    "version": "11.0.1.1.0",
+    "version": "11.1.0.0.0",
     "category": "Project Management",
     "website": "https://www.tecnativa.com",
     "author": "Tecnativa, Onestein, "
--- a/project_timeline/__manifest__.py
+++ b/project_timeline/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Project timeline",
     "summary": "Timeline view for projects",
-    "version": "11.1.0.0.0",
+    "version": "11.0.2.0.0",
     "category": "Project Management",
     "website": "https://www.tecnativa.com",
     "author": "Tecnativa, Onestein, "
--- a/project_timeline/models/project_task.py
+++ b/project_timeline/models/project_task.py
@@ -2,11 +2,13 @@
 # Copyright 2017 Tecnativa - Carlos Dauden
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import api, models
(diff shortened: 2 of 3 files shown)


project_timeline/__manifest__.py
```diff
@@ -5,7 +5,7 @@
 {
     "name": "Project timeline",
     "summary": "Timeline view for projects",
-    "version": "11.0.1.1.0",
+    "version": "11.1.0.0.0",
```
nikul-serpentcs: @InfernalLV `11.0.1.0.0`

project_timeline/__manifest__.py
```diff
@@ -5,7 +5,7 @@
...
     "category": "Project Management",
     "website": "https://www.tecnativa.com",
```
nikul-serpentcs, line 10: @InfernalLV Should be `https://github.com/OCA/project`

nikul-serpentcs (changes requested):
Improve Code

project_timeline/__manifest__.py
pedrobaeza: It should `11.0.1.2.0`

karlis-dreizis: I see what I got wrong. But shouldn't it then be `11.0.2.0.0`? This pull request addresses a breaking change in OCA/web#1206. Doesn't that mean that this pull request introduces a breaking change as a consequence of the first one? Which Is why I went for the bump on the highest digit.

pedrobaeza: Yes, 11.0.2.0.0 makes sense.

karlis-dreizis: Version bumped

karlis-dreizis:
Am I supposed to address the ci/runbot error? 
Seems this could be what causes the failure `fatal: Couldn't find remote ref refs/merge-requests/480/head`

karlis-dreizis:
I inspected some other PR's and took a look at runbot repo. Based on https://github.com/odoo/runbot/blob/master/runbot/models/build.py#L22 and its usage I'd venture to say that ci/runbot fails because of following line from the log
`2019-03-08 12:02:02,204 193 WARNING openerp_test odoo.modules.registry: Models have no table: report.project.task.user.`

Again, how do I address this? If I do at all? The warning is unrelated to my PR. The report.project.task.user model is a sql view backed report from the `project` module.

I'd like to get this merged as soon as reasonably possible. So any guidance would be greatly appreciated.

Thanks!

pedrobaeza:
Indeed, this is not your fault, so unrelated to your PR and you can leave it. Let's see if that's fixable in OCA itself or in Odoo.

PR #495, opened 2019-04-09
[12.0][IMP] project_timeline - Enable subtasks

### LINKED ISSUE #485: [IMP] project_timeline: Timeline view & Subtasks
Currently, the timeline view does not distinguish tasks and subtasks. 
It would be nice to have a filter to show only the parent tasks (without subtasks), as defined in #347 

Also, when accessing the subtasks of a specific task, the timeline view is not available.

emagdalenaC2i:
Fixes #485

### DIFF
--- a/project_timeline/views/project_task_view.xml
+++ b/project_timeline/views/project_task_view.xml
@@ -44,7 +44,13 @@
         <field name="view_mode">kanban,tree,form,calendar,timeline,pivot,graph</field>
     </record>
 
-    <record id="view_task_form2" model="ir.ui.view">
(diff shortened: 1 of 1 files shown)


tarteo:
Can you rebase please for the scss fix?

emagdalenaC2i:
@tarteo rebased ready

tarteo:
@emagdalenaC2i I still get the error in Runbot:

image

But in the code I see you're only adding an action. Don't you need a menu item too?

emagdalenaC2i:
I think I didn't rebase it properly and the scss error is still there

I will check it later

emagdalenaC2i:
Closed, see #498

PR #492, opened 2019-04-08
[FIX][12.0] project_timeline_hr_timesheet

emagdalenaC2i:
scss Issue

PR #496, opened 2019-04-09
[12.0][FIX] project_timeline_hr_timesheet: fix scss

alexey-pelykh:


alexey-pelykh:
Issue with scss approach is that variable used is defined in less code, not scss, so we need to use something else.

nikul-serpentcs:
@alexey-pelykh check Please https://github.com/OCA/project/pull/492#issuecomment-481114107

tarteo (approved):
This is the best way, reusing the colors enables themes.

Issue #500, opened 2019-04-17
New List of Project Modules (v12)(Feedback and Reviews Needed)


patrickrwilson:
I'm new to contributing to the community and I have a list of project modules that I would like some feedback and interest for reviewing before I submit some PR's for them. I'm looking to add the following modules for v12:

New (Quick descriptions of each):
- **project_type** | Adds a new object to projects that allows users to set project types. | project_category already serves this purpose. (#502 canceled)
- **project_milestone** | Adds milestones to projects | #501
- **project_progress** | Adds progress measurements to projects (closed vs open tasks)
- **project_status** | Adds status to projects | #503 
- **project_tags** | Adds the task tags to the project | #505 
- **project_task_checklist** | Adds a checklist list to tasks
- **project_task_date_shift** | Adds the ability to shift all active task dates when the start date changes on the project. | #506 
- **project_task_timer** | Adds a timer function on tasks where a user can start/stop timers and easily add a time entry. | Other modules exist so this is most likely not needed.
- **project_template** | Convert projects to templates and easily create a project form the template. Very powerful when combined with project_milestone and date_shifting. | #504 
- **project_autoforecast** | My first attempt at automatically creating forecasts when a task is assigned, has start/end dates and has planned hours.

Migrated to v12:
- **project_department** | migrated v11 to v12, adds

Please let me know your interests and who would be willing to review some of these. I can then initiate some pull requests. Thank you for your time.

OSevangelist:
Hi @patrickrwilson at least from the functional perspective i could help in reviewing. Did you follow any specific project management method when creating these modules or do you use them with a special methodological framework in mind? I suggest that you do a pull request for the module code. Looking forward to see it. Best Frederik

patrickrwilson:
Thanks Frederik, I'll try to get a PR tomorrow for these. We currently use a solution built for Dynamics 365 and I took the experience of that as well as some other project management tools to build these modules. Basically these modules fill in the gaps that were missing from Odoo compared to what we are currently using. The goal is for when we move over to Odoo, the users will already be familiar with the projects and hopefully others can use these too. Also, I tried to build some of these so they would be optional on each project. Like the milestone and shift dates specifically, those are enabled on each project.

pedrobaeza:
Some of the functionalities you mention are already in the repository (maybe under other names). For scheduling, see our R&D project #419 that is pretty finished.

JordiBForgeFlow:
[...]

PR #502, opened 2019-04-18
[ADD] Project Type Module

patrickrwilson:
Adds Project Type Module which provides a way to add project types to the project object.


patrickrwilson:
Original Discussion #500

patrickrwilson:
I'm not exactly sure why this is failing due to this error: 

2019-04-18 21:14:37,060 6476 ERROR openerp_test odoo.sql_db: bad query: b'ALTER TABLE "project_project" ALTER COLUMN "project_type_id" SET NOT NULL'
ERROR: column "project_type_id" contains null values.

Any insight would be appreciated.

max3903:
@patrickrwilson The field is required:
https://github.com/OCA/project/pull/502/files#diff-a6821d7bc2e9ca91cd0f974468ecabc7R10

and the project module provides demo data which have an empty value.

One option to solve this is to set a value in the `pre_init_hook`.

Here is an example:
https://github.com/OCA/product-attribute/blob/11.0/product_code_mandatory/__manifest__.py#L17
https://github.com/OCA/product-attribute/blob/11.0/product_code_mandatory/__init__.py#L7

pedrobaeza:
I see this very similar to https://github.com/OCA/project/tree/12.0/project_category in the purpose.

patrickrwilson:
Thank you max3903, I understand the required field issue now. pedrobaeza, you are correct, that module does server our needs and this would essentially be a duplicate of that one. Sorry I missed that when I was looking for this. I do like how that module works on tasks too. 

How do I go about cancelling this pull request since the module is no longer needed?

pedrobaeza:
I close the PR then.

PR #501, opened 2019-04-18
[ADD] Project Milestone Module

patrickrwilson:
New module to add milestones to projects. 

Original discussion at #500

PR #504, opened 2019-04-19
[ADD] Project Template Module

patrickrwilson:
Adds Project Template Module

Original Discussion #500


project_template/models/project.py
```diff
@@ -0,0 +1,70 @@
...
+            new_name = self.name.replace(" (TEMPLATE)", " (COPY)")
+
+        new_project = self.copy(default={'name': new_name,
+                                         'active': True,
+                                         'total_planned_hours': 0.0})
+        if new_project.subtask_project_id != new_project.id:
+            new_project.subtask_project_id = new_project.id
+        # Clear Email Alias if it was set in the Template
+        if new_project.alias_name:
+            new_project.alias_name = False
```
dreispt: maybe we can set this on the copy default, at line 15.

project_template/models/project.py
```diff
@@ -0,0 +1,70 @@
...
+            'target': 'current',
+            'res_id': new_project.id,
+            'type': 'ir.actions.act_window'
+        }
+
+    # ADD "(TEMPLATE)" TO THE NAME WHEN PROJECT IS MARKED AS A TEMPLATE
+    @api.onchange('is_template')
+    def on_change_is_template(self):
+        # Add "(TEMPLATE)" to the Name if is_template == true
+        if self.name:
```
dreispt, line 45: I feel that this `if` statement is not needed.

project_template/models/project.py
patrickrwilson: Good point, I removed lines 20-22 and added this to the copy function. Seems to work as intended. Thanks.

project_template/models/project.py
patrickrwilson: I think you are right, removing it had no impact on functionality so i've removed it. Thanks.

rafaelbn:
I think is better idea to restore this functionality

https://github.com/odoo/odoo/blob/8.0/addons/project/project.py#L281

Add a Boolean for templates is not bad an is simple, but imagine adding more Boolean for more similar and simple purposes? :smile: 

It is just my opinion for a more general approach.

Not blocking

tarteo (commented):
Would be nice to be able to make a template from a project (reversed create_project_from_template).

tbaden:
by coincidence we did something similar for v10 https://github.com/OCA/project/pull/512 😅

patrickrwilson:
@rafaelbn I'm not too familiar with project functionalities of older versions of Odoo but I do see the appeal of some of those features. I don't think it would be too much to improve this module with additional features. The boolean was a simple, straight forward way to manage the templates which is why I chose that route. I'm sure there are many other options out there, I just chose one that was similar to how our Dynamics 365 project solution works.

@tarteo This is a good idea for a possible improvement. Currently you can change any project to a template simply by the boolean field. To keep the project you want to create a template from, you would have to duplicate it first then change to template. I can see this function having a bit more logic as you probably would want it to 'reset' the tasks and such as if they are from a current project they would have additional data in them. I think this would require some discussion on the best routes.

@tbaden I see that pr was never merged, perhaps it wasn't fully completed?

tbaden:
If you go to "Configuration" -> "Projects" -> "Create Project" 
Bildschirmfoto 2019-04-26 um 16 03 01

tbaden:
@patrickrwilson no we created it just this week and today we created the PR

patrickrwilson:
@tbaden nice catch, I missed creating the project that way. I'll work on the fix. Thanks

This has been fixed, I recently removed a if statement "if self.name:" in that function cause a reviewer and I thought it wasn't needed, however now I know why I originally had it in there. The error shouldn't show anymore. :)

patrickrwilson:
Fixed creation error and squashed commits. Thanks for your reviews. Please let me know what else might be needed.

rafaelbn:
It's OK. I just pointed that the functionality was in Odoo v8, more complete as @tarteo said, because this could help is very important to have back and front view in the ecosystem :smile: :




I understand you, and it's OK but i want to explain that the solutions in the OCA should be thunk as much generic as possible with the idea that more companies can use it

> I just chose one that was similar to how our Dynamics 365 project solution works

Usually this process of thinking which is the more generic and expandable way to do something for more customers and companies make better solutions

Do not get me wrong, I do not want to discourage you, much less. I thank you and I am very happy to have a new contributor in OCA but I wanted to give you my point of view.

After testing as is:

- I think menu "Templates" is not necessary, it take a really big space in the menu for just a little functionality. Plus you have exacly the same in the first menu changing filter.



- The Boolean for is template could be added in other place? Like right side of the other Boolean "Allow timesheets" or inside the Tab "Settings" inside the "Configuration" section



or

patrickrwilson:
@rafaelbn great suggestions, I think these are good improvements and will make these changes shortly.

patrickrwilson:
@rafaelbn I removed the template menu item and it does look much cleaner. I moved the bool field up below the name which looks nicer. I didn't put it in the configuration area since, at least on my instance, it required debug mode to see that. I do think your suggested changes did give it a cleaner look. Thanks

rafaelbn:
Hi @patrickrwilson , thank you!

It really needed test `codecov/patch — 17.64%` if don't know how to achive this, please let us know @Yajo or @pedrobaeza will point you some info so you can learn.

By the way, maybe you don't know it, we use this module in all our instances of Odoo: https://github.com/OCA/server-ux/tree/12.0/base_technical_features . This show some needed stuff for the users that Odoo hide and only is shown in Debug mode as you said.

Regards

rafaelbn:
How to make test, I found this @patrickrwilson 

- https://www.odoo.com/documentation/12.0/reference/testing.html
- https://www.odoo.yenthevg.com/creating-automated-tests-odoo/

rafaelbn:
@tarteo I think you can review this. It's quite simple and easy to test.

> Would be nice to be able to make a template from a project (reversed create_project_from_template).

@patrickrwilson add this to Known issues

IMHO with test is Ready to merge

@dreispt what do you think? :smile:

rafaelbn (commented):
Missing test

patrickrwilson:
@rafaelbn I'm working on tests, might take me a bit to figure it out but thanks for the links.

@tarteo I can add a simple 'Create template from project' button but it may need further discussion. If we use the copy method, then we would need to 'clean' the tasks as you wouldn't want all the junk that gets added to them if the project wasn't new. Maybe i'm thinking too much into it or there is another way you were thinking?

tarteo (changes requested):
It works. The project_key module has a constraint on key so to make a project from a template you need to change the key to something unique. The only thing for me is this button:
image
It doesn't match Odoo's style. Maybe you can put the button here:
image
This way the user can't accidentally click it and looks good:

patrickrwilson:
@tarteo I'm fine with making the change on the button. I'm all for consistency and clean looks. I'll have that in my next commit.

I can add a check in there that if the project_key module is installed, it automatically changes the key to be used but i'm not sure what to. Maybe a random number? Otherwise it can change it to something that already exists. Ideas?

tarteo:
@patrickrwilson I think it is not a problem of this module, it should be changed in project_key. I think I already made the fix in 12.0 https://github.com/OCA/project/pull/426

patrickrwilson:
@tarteo sounds good, I do have it working with adding it in the card menu instead of the button, so that will be in the next commit. I'm currently struggling with the unit tests and will probably have quite a few more commits for that soon.

patrickrwilson:
I'm still stuck on the test, my last commit has the menu change, but i'm just not sure how to build out the test and could use your expertise.

tarteo (approved):
Thanks!

project_template/tests/test_project_template.py
```diff
@@ -0,0 +1,25 @@
...
+        super().setUp()
+        test_project = self.env['project.project'].create({
+            'name': 'TestProject'})
+        self.env['project.task'].create({
+            'name': 'TestTask',
+            'project_id': test_project.id})
+
+    # TEST 01: Set project to be a template
+    def test_template(self):
+        test_project.is_template = True
```
tarteo: test_project cannot to be reached from here it's out of scope. You need to make an instance variable using ```self```.  In ```setUp``` do self.test_project instead.

To run unit tests locally add ```--test-enable -u project_template -d {mydb}```. Replace {mydb} with your database name. You can find more info here: https://www.odoo.com/documentation/11.0/reference/testing.html

patrickrwilson:
Thanks for everyone's help. I'm slowly getting better at unit tests and this is much closer. 

I have a piece of code in there that i'm not sure how to put into the unit test. If the project_milestone module is also installed (PR #501) then it will link the new tasks to the new milestones when a project is created from a template. I had to do this since deepcopy was a dangerous option I didn't feel like I should dive into when creating this. It ignores that code if the milestone module is not installed.

Does anyone know how to add some testing in unit tests for this? Suggestions?

max3903:
@patrickrwilson You can create a `project_template_milestone` module with:
* your piece of code
* `depends = project_template, project_milestone`
* `auto_install = True`

patrickrwilson:
@max3903 that makes sense and seems to be the proper way to do it. I've removed the code and will work on another module that implements that. Thanks. I've squashed the commits.

max3903 (approved):

PR #512, opened 2019-04-26
[ADD] project_template module

tbaden:
This Module implements templates for projects. You can create templates from 
scratch or from an existing project. After setting up the templates, users can 
create new projects based on the predefined templates. This is saving time in 
the project creation process and improves user friendlyness.


tarteo:
Is this a back port of #504 ?

tbaden:
no it's not, just a funny coincidence. so we have to check which is better 🤔

rafaelbn:
Runbot doesn't work in order to test it :cry:

tbaden:
@rafaelbn any idea what could be the issue with runbot? there are no errors in the logs.
I tried some other PRs for v10 and they behave the same.
maybe a general issue for v10?

https://github.com/OCA/account-invoicing/pull/468 and https://github.com/OCA/account-invoicing/pull/456 both are broken. I rebuild them and at the first connect you get an "internal server error" and right away the login screen.

tbaden:
not sure what did it, but runbot is now working @rafaelbn

tbaden:
closed to prevent duplicate

PR #514, opened 2019-04-29
[10.0][FIX] project_model_to_task compute method name

tbaden:
fix of compute method name & missing license in manifest

### DIFF
--- a/project_model_to_task/models/project.py
+++ b/project_model_to_task/models/project.py
@@ -12,7 +12,7 @@ class ProjectTask(models.Model):
 
     @api.multi
     @api.depends('model_reference')
-    def _get_origin(self):
+    def _compute_task_origin(self):
         for rec in self:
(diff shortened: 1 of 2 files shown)


project_model_to_task/models/project.py
```diff
@@ -32,7 +32,7 @@ def _authorised_models(self):
         help="Action called to go to the original window.")
     model_reference = fields.Reference(
         selection='_authorised_models')
-    task_origin = fields.Char(compute='_get_origin')
+    task_origin = fields.Char(compute='_compute_get_origin')
```
pedrobaeza: It should be `_compute_task_origin`

tbaden: thanks for the fast response. fix is pushed

bealdav (approved):

pedrobaeza (approved):

PR #515, opened 2019-04-29
[10.0][FIX] timesheet_currency compute function & missing newline

tbaden:
fix of computed function name and missing newline in xml file


project_timesheet_currency/models/analytic.py
```diff
@@ -22,12 +22,15 @@ def _get_timesheet_cost(self, values):
                 date=values['date']).compute(res['amount'], base_currency)
         return res
 
-    amount_currency = fields.Float(compute="_get_amount_currency", store=True)
+    amount_currency = fields.Float(
+        compute="_compute_get_amount_currency",
```
pedrobaeza: It should be `_compute_amount_currency`

tbaden: thanks for the fast response. fix is pushed

pedrobaeza (approved):
Fast-tracking this

PR #517, opened 2019-04-29
[FIX][10.0] sale_order_project Redefining warning

tbaden:
fix Redefining built-in 'Warning'
fix encoding header C8201(no-utf8-coding-comment)


sale_order_project/models/sale.py
```diff
@@ -4,7 +4,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
 from odoo import api, models, _
-from odoo.exceptions import Warning
+from odoo import exceptions
```
pedrobaeza, line 6: You can put this in the first row

sale_order_project/models/sale.py
```diff
@@ -28,7 +28,7 @@ def action_create_project(self):
         project_obj = self.env['project.project']
         for order in self:
             if order.project_id:
-                raise Warning(_(
+                raise exceptions.Warning(_(
```
pedrobaeza: Use `exceptions.UserError` as the other is deprecated

pedrobaeza (changes requested):

tbaden:
thanks for the feedback. changed it.

tbaden:
runbot/travis checks all green with the requested changes

sale_order_project/models/sale.py
pedrobaeza: Well, I meant to put combined with the first row:
```python
from odoo import api, exceptions, models, _
```
But it's not so important.

pedrobaeza (approved):
Fast-tracking this

sale_order_project/models/sale.py
tbaden: will do on the next one.

PR #526, opened 2019-04-30
[FIX][12.0] project_key relative path

tbaden:
fix project_key/tests/test_controller.py:6: [W7950(odoo-addons-relative-import), ] Same Odoo module absolute import. You should use relative import with "." instead of "openerp.addons.project_key"


pedrobaeza (approved):
Fast-tracking this

PR #528, opened 2019-04-30
[FIX][12.0] project_task_code missing api.multi

tbaden:
fix of
project_task_code/models/project_task.py:34: [W8102(copy-wo-api-one), ProjectTask.copy] Missing api.one or api.multi in copy function.

### DIFF
--- a/project_task_code/models/project_task.py
+++ b/project_task_code/models/project_task.py
@@ -31,7 +31,9 @@ class ProjectTask(models.Model):
                 )
         return super().create(vals_list)
 
+    @api.multi
     def copy(self, default=None):
+        self.ensure_one()
         if default is None:
             default = {}
         default['code'] = self.env['ir.sequence'].next_by_code('project.task')


pedrobaeza:
I think this check is no longer needed since >=10.0. What do you think, @moylop260 ?

moylop260:
You are right!
Since that by default all methods are api.multi
-- 
Moisés López Calderón
Mobile: (+521) 477-752-22-30
Twitter: @moylop260
hangout: moylop260@vauxoo.com
http://www.vauxoo.com - Odoo Gold Partner
Twitter: @vauxoo

pedrobaeza:
Can you disable then that check?

dalonsod (approved):
LGTM

moylop260:
@pedrobaeza 
I just checked a little more about this check and it is still valid since that the `copy` method doesn't allow copy multi-records.

https://github.com/odoo/odoo/blob/ca57218b6122ec00be51fa177067dfba7d168e91/odoo/models.py#L4290

It is a beta check so it is not required to fix it.

If you think that it should be entirely disabled even a beta check then you should create a new issue to MQT:
 - https://github.com/oca/maintainer-quality-tools/issues/new

moylop260 (approved):

PR #536, opened 2019-05-10
[ADD] project_template_milestone

patrickrwilson:
This module is an addition to the project_template and project_milestone modules. It allows the milestones to be copied when a project is created from a template that contains milestones.


max3903:
@patrickrwilson You can remove the `models/__pycache__` directory. This should not be included in the commit as it contains compiled code.

No need to check if the `project_milestone` is installed, it will be installed as it is a dependency.

You should call `super()` somewhere in your method to avoid duplicate code.

patrickrwilson:
I'll work on @max3903 suggestions, might take a bit but i'll let you all know when it's ready for more reviewing.

project_template_milestone/tests/test_project_template.py
```diff
@@ -0,0 +1,68 @@
...
+        # Verify that the project was created successfully
+        self.assertEqual(len(new_project), 1)
+
+        # Verify that the Milestones were created successfully
+        new_milestones = self.env['project.milestone'].search([(
+            'project_id', '=', self.test_project.id)])
+        self.assertEqual(len(new_milestones), 2)
+
+        # Verify that the tasks were created successfully with milestones
+        new_tasks_milestone1 = self.env['project.task'].search([(
```
max3903:
```suggestion
        self.env['project.task'].search([(
```

project_template_milestone/tests/test_project_template.py
```diff
@@ -0,0 +1,68 @@
...
+            'milestone_id', '=', self.test_milestone_1.id)])
+        self.assertEqual(len(new_project), 2)
+
+        new_tasks_milestone2 = self.env['project.task'].search([(
```
max3903:
```suggestion
        self.env['project.task'].search([(
```

max3903 (approved):

patrickrwilson:
Green! Ready for more reviews. Thanks

osi-scampbell (approved):

PR #540, opened 2019-05-16
12.0 mig project_recalculate

ernestotejeda:
* readme by fragments
* adapt python code to use the new methods of the calendar class
* adapt tests

Cc @Tecnativa


project_recalculate/models/project_task.py
```diff
@@ -57,14 +60,17 @@ def _dates_onchange(self, vals):
         start = from_string(date_start)
         end = from_string(date_end)
         resource, calendar = self._resource_calendar_select()
+
+        if not start.tzinfo:
```
pedrobaeza: For saving a lot of code lines, call directly in all places to _resource_timezone, and there, you check:
```python
if dt.tzinfo:
    return dt
```

project_recalculate/models/project_task.py
```diff
@@ -82,13 +88,17 @@ def _dates_onchange(self, vals):
                     return vals
                 project_date = from_string(self.project_id.date)
                 start, end = start, project_date
+
```
pedrobaeza: Don't add useless empty lines

project_recalculate/models/project_task.py
```diff
@@ -193,39 +202,71 @@ def _interval_context_tz(self, interval):
         start = fields.Datetime.context_timestamp(self, start)
         end = datetime.now().replace(hour=interval[1])
         end = fields.Datetime.context_timestamp(self, end)
-        return (start.hour, end.hour)
+        result = Intervals()
+        result._items.append((start, end))
```
pedrobaeza: Should you access to the protected variable `_items`? Isn't there a way to add elements directly through a public API?

project_recalculate/models/project_task.py
```diff
@@ -193,39 +202,71 @@ def _interval_context_tz(self, interval):
...
-        intervals = calendar.get_working_intervals_of_day(
-            start_dt=day_date, compute_leaves=True, resource_id=resource.id,
-            default_interval=default_interval,
-        )
-        return intervals and intervals[0] or False
-
-    def _calendar_schedule_days(self, days, day_date,
-                                resource=None, calendar=None):
+        intervals = self._get_work_intervals(day_date, resource, calendar)
+        return intervals._items and intervals._items[0] or False
```
pedrobaeza: Isn't the first element accessible without needing to access protected variable? (`intervals[0]`?)

project_recalculate/models/project_task.py
```diff
@@ -235,23 +276,30 @@ def task_recalculate(self):
...
-                end = self._calendar_schedule_days(
-                    task.estimated_days, date_start, resource, calendar)[1]
-                if end:
-                    date_end = end[1]
+                end_planned_dt = self._calendar_plan_days(
+                    task.estimated_days, date_start, resource, calendar)
+                if end_planned_dt:
+                    date_end = self._last_interval_of_day_get(
+                        end_planned_dt, resource, calendar)[1]
+
```
pedrobaeza: Remove useless empty lines inside a method

project_recalculate/models/project_task.py
```diff
@@ -193,39 +202,71 @@ def _interval_context_tz(self, interval):
...
+            return False
+        # date to datetime
+        if days >= 0:
+            day_dt = datetime.combine(day_date, datetime.min.time())
+        else:
+            day_dt = datetime.combine(day_date, datetime.max.time())
+
+        if not day_dt.tzinfo:
+            day_dt = self._resource_timezone(day_dt, resource)
+
```
pedrobaeza: Remove empty lines inside methods (do it for all the cases)

project_recalculate/models/resource_calendar.py
```diff
@@ -25,13 +26,49 @@ def get_working_days_of_date(self, start_dt=None, end_dt=None,
...
         return days
+
+    @api.multi
+    def plan_days_to_resource(self, days, day_dt, compute_leaves=False,
+                              resource=None, domain=None):
+        day_dt, revert = make_aware(day_dt)
+
+        # which method to use for retrieving intervals
+        if compute_leaves:
+            get_intervals = partial(self._work_intervals, resource=resource,
```
pedrobaeza, line 39: Why accessing protected variable `_work_intervals`?

project_recalculate/models/resource_calendar.py
```diff
@@ -25,13 +26,49 @@ def get_working_days_of_date(self, start_dt=None, end_dt=None,
...
+                                    domain=domain)
+        else:
+            get_intervals = partial(self._attendance_intervals,
+                                    resource=resource)
+
+        if days > 0:
+            found = set()
+            delta = timedelta(days=14)
+            for n in range(100):
```
pedrobaeza, line 48: This double loop is not very performant. Isn't a better solution?

project_recalculate/tests/test_project_project.py
```diff
@@ -11,14 +10,14 @@ class TestProjectProjectBegin(base.BaseCase):
     project_final_dates_one_task = {
         'date_begin': [
             # name, date_start, date
-            ['pj_0', '2015-08-01', '2015-08-05'],
-            ['pj_1', '2015-08-02', '2015-08-05'],
-            ['pj_2', '2015-08-03', '2015-08-05'],
+            ['pj0', date(2015, 8, 1), date(2015, 8, 5)],
```
pedrobaeza: Why changing key names?

project_recalculate/tests/test_project_task.py
```diff
@@ -23,23 +23,23 @@ def __init__(self, methodName='runTest'):
         #   both are valid
         self.task_days_res = {
             'date_begin': {
-                'pj_0': list(self.task_days['date_begin']),
-                'pj_1': list(self.task_days['date_begin']),
-                'pj_2': list(self.task_days['date_begin']),
+                'pj0': list(self.task_days['date_begin']),
```
pedrobaeza: Why changing key names? Better to reduce the diff

project_recalculate/wizard/recalculate_wizard.py
```diff
@@ -1,11 +1,11 @@
-# -*- coding: utf-8 -*-
 # See README.rst file on addon root folder for license details
 
 from odoo import models, fields, api
 
 
 class ProjectRecalculateWizard(models.TransientModel):
     _name = 'project.recalculate.wizard'
+    _description = 'project recalculate wizard'
```
pedrobaeza: Use upper in first letter

pedrobaeza (changes requested):

project_recalculate/models/project_task.py
ernestotejeda: Thinking better I think it does not make much sense to use the _interval_default_get method. Since in v12, the `resource.calendar` methods no longer receive a` default_interval` as a parameter.
So, it is not necessary to create an Intervals object

project_recalculate/models/resource_calendar.py
ernestotejeda: I have added the `plan_days_to_resource` method to resource.calendar with the intention that it does the same as plan_days but for a specific resource that is passed by parameter. Therefore, the code of that method is almost the same as that of the existing plan_days method, that's why there is the call to the protected method _work_intervals.
I have done this with the intention that the planning of the days take into account the leaves of the resource

project_recalculate/tests/test_project_project.py
ernestotejeda: because if this module is tested with the project_key module, then the tests do not pass. The project_key module adds the `key` field at the project level and the value of that field must be unique. What happens is that this field is automatically filled depending on the module name and for these module names, the value generated for the `key` field is the same.

ernestotejeda:
Changes done

project_recalculate/tests/test_project_project.py
pedrobaeza: This must be definitively fixed on `project_key` which is causing a lot of problems with the rest of the modules. Please do another PR disabling the features if on test mode and not testing the module itself (via a context key). Example: https://github.com/OCA/bank-payment/blob/7ce8bbb88bd8c889347a7550aa973124b847af58/account_payment_transfer_reconcile_batch/models/payment_order.py#L28

pedrobaeza:
You can rebase now and leave test data as it was

ernestotejeda:
Changes done

project_recalculate/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,19 @@
+You can define working calendar at Setting > Technical > Resource > Working time
```
yajo:
```suggestion
You can define working calendar at Setting > Technical > Resource > Working time.
```

project_recalculate/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,19 @@
...
+Then assign this calendar to a resource (related with an user), a project or
```
yajo:
```suggestion
Then assign this calendar to a resource (related with a user), a project or
```

project_recalculate/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,19 @@
...
+a company
```
yajo:
```suggestion
a company.
```

project_recalculate/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,19 @@
...
+  ends at 18:00
+* If not user assigned or resource hasn't working calendar, search first
+  working calendar of the company
+* If no working calendar found, then every day is workable and work starts at
+  08:00 and ends at 18:00
+
+Also you can define which task stages are included in recalculation when
+'Project recalculate' button is clicked. By default, all are included.
+To change this go to Project > Configuration > Stages > Task Stages and change
+'Include in project recalculate' field
```
yajo:
```suggestion
the 'Include in project recalculate' field.
```

yajo (commented):
Some typos

ernestotejeda:
Changes done

project_recalculate/models/project_task.py
```diff
@@ -0,0 +1,284 @@
...
+        else:
+            if not self.project_id.date:
+                raise exceptions.UserError(
+                    _('End Date field must be defined.')
+                )
+            project_date = self.project_id.date
+            days = self.from_days * (-1)
+        return increment, project_date, days
+
+    def _resource_timezone(self, dt, resource=None, calendar=None):
```
Tardo, line 190: Here perhaps can use Odoo Style:
```python
    tz = 'UTC'
    if resource or calendar:
        tz = (resource or calendar).tz
    record_user_timestamp = env.user.sudo().with_context(tz=tz)
    return fields.Datetime.context_timestamp(record_user_timestamp, dt)
```

ernestotejeda: What happens is method should not return the converted dt to the timezone of the resource or calendar. The method must return the same dt but aware of the timezone of the resource or calendar.
Example:
for
    `dt = 2019-06-23 00:00:00`
    `tz = 'Europe / Brussels'`
I want it to return
2019-06-03 00:00:00+02:00 instead of 2019-06-23 02:00:00+02:00

project_recalculate/models/resource_calendar.py
```diff
@@ -0,0 +1,74 @@
...
+                                 resource=None):
+        self.ensure_one()
+        if start_dt is None:
+            start_dt = datetime.now().replace(hour=0, minute=0, second=0)
+        if end_dt is None:
+            end_dt = datetime.now().replace(hour=23, minute=59, second=59)
+        days = 0
+        current = start_dt
+        while current <= end_dt:
+            if id is None:
```
yajo: :warning: [`id` is a built-in function](https://docs.python.org/3/library/functions.html#id) that will never be `None`. I'm not sure what this code was supposed to do, but it's wrong. :sweat_smile:

project_recalculate/models/project_task.py
```diff
@@ -235,23 +249,29 @@ def task_recalculate(self):
...
+            if date_start:
+                date_start = date_start.astimezone(utc)
+            if date_end:
+                date_end = date_end.astimezone(utc)
+
             task.with_context(task.env.context, task_recalculate=True).write({
-                'date_start': date_start and to_string(date_start) or False,
-                'date_end': date_end and to_string(date_end) or False,
-                'date_deadline': date_end and to_string(date_end) or False,
+                'date_start': date_start and date_start or False,
```
yajo: Oops, did you mean this?

```suggestion
                'date_start': date_start or False,
```

project_recalculate/models/project_task.py
```diff
@@ -0,0 +1,284 @@
...
+                    date_end = self._last_interval_of_day_get(
+                        end_planned_dt, resource, calendar)[1]
+            if date_start:
+                date_start = date_start.astimezone(utc)
+            if date_end:
+                date_end = date_end.astimezone(utc)
+
+            task.with_context(task.env.context, task_recalculate=True).write({
+                'date_start': date_start and date_start or False,
+                'date_end': date_end and date_end or False,
```
yajo:
```suggestion
                'date_end': date_end or False,
```

project_recalculate/models/project_task.py
```diff
@@ -0,0 +1,284 @@
...
+                'date_deadline': date_end and date_end or False,
```
yajo:
```suggestion
                'date_deadline': date_end or False,
```

project_recalculate/models/project_task.py
```diff
@@ -0,0 +1,284 @@
...
+
+    @api.constrains('estimated_days')
+    def _estimated_days_check(self):
+        for task in self:
+            if task.estimated_days <= 0:
+                raise exceptions.ValidationError(
+                    _('Estimated days must be greater than 0.')
+                )
+
+    def _dates_onchange(self, vals):
```
yajo: I don't like this method name. It's not an onchange, so it's confusing. Could you change it?

project_recalculate/models/project_project.py
```diff
@@ -0,0 +1,66 @@
...
+
+            NOTE: Do not perform any write operations to DB
+        """
+        vals = {}
+        self.ensure_one()
+        if not self.tasks:
+            return vals
+        # Here we consider all project task, independently from the value in
+        # include_in_recalculate
+        start_task = min(self.tasks, key=lambda t: t.date_start or t.date_end)
```
yajo, line 34: Shouldn't you filter out tasks without `include_in_recalculate`? :thinking:

project_recalculate/models/project_project.py
```diff
@@ -0,0 +1,66 @@
...
+            if (project.calculation_type == 'date_begin' and not
+                    project.date_start):
+                raise UserError(_("Cannot recalculate project because your "
+                                  "project doesn't have date start."))
+            if (project.calculation_type == 'date_end' and not
+                    project.date):
+                raise UserError(_("Cannot recalculate project because your "
+                                  "project doesn't have date end."))
+            for task in project.tasks:
+                task.task_recalculate()
```
yajo: Why not just `project.tasks.task_recalculate()`?

project_recalculate/models/project_task.py
```diff
@@ -0,0 +1,284 @@
...
+        # Calculate from_days depending on project calculation type
+        calculation_type = self.project_id.calculation_type
+        if calculation_type:
+            invert = False
+            increment = calculation_type == 'date_begin'
+            if increment:
+                if not self.project_id.date_start:
+                    # Can't calculate from_days without project date_start
+                    return vals
+                project_date = from_string(self.project_id.date_start)
```
yajo: It should return a datetime object already, isn't it?

```suggestion
                project_date = self.project_id.date_start
```

project_recalculate/models/project_task.py
```diff
@@ -0,0 +1,284 @@
...
+                start, end = project_date, start
+            else:
+                if not self.project_id.date:
+                    # Can't calculate from_days without project date
+                    return vals
+                project_date = from_string(self.project_id.date)
```
yajo:
```suggestion
                project_date = self.project_id.date
```

project_recalculate/models/resource_calendar.py
```diff
@@ -0,0 +1,74 @@
...
+                days += 1
+            else:
+                end_day = current.replace(hour=23, minute=59, second=59)
+                end = end_dt if end_day > end_dt else end_day
+                obj = self.with_context(tz='UTC')
+                working_intervals = obj._work_intervals(current, end, resource)
+                if len(working_intervals):
+                    days += 1
+            next_dt = current + timedelta(days=1)
```
yajo:
```suggestion
```

project_recalculate/models/resource_calendar.py
```diff
@@ -0,0 +1,74 @@
...
+            current = next_dt
```
yajo:
```suggestion
            current += timedelta(days=1)
```

yajo (changes requested):
I have to admit I mostly don't understand this source code, but if tests pass... OK.

However there's one point that totally seems a bug, and there are some suggestions too.

project_recalculate/models/project_project.py
ernestotejeda: Well, as I've seen in the commit history, it has always been considered all project tasks, the ones in a stage with `include_in_recalculate = False` and the ones with `include_in_recalculate = True`. I do not know if I should change that.

rafaelbn:
Hello @patrickrwilson maybe you are interested in this one. As commented in https://github.com/OCA/project/issues/500

project_recalculate/models/project_task.py
Tardo: I don't understand why need timezone if you like hours in UTC...

2019-06-03 00:00:00+02:00 in UTC is 2019-05-03 22:00:00+00:00

patrickrwilson:
@ernestotejeda thanks for working on converting this module. Do you mind squashing all your commits after your next changes please.

rafaelbn (commented):
Tested in runbot IMHO

- [x] If the task don't have any user assigned the module don't calculate anything
- [x] If the task have an user assigned with a working time 30 hour and the project have a working time of 40 hours the task are planned with the working time of the project and not with the working time of the user

This should be fixed or added to know issues as you could have no calculation or a wrong planning calculation based in a working plan thatis diferent of the user. This cases have been never deliberate from v8

pedrobaeza:
I think you should use user working time if present, and project if not.

ernestotejeda:
@rafaelbn , check now, I've already made those changes

rafaelbn:
@ernestotejeda in the project form the fields Start date and Expiration date are duplicated
(tested in http://3373450-540-efe2ea.runbot2-2.odoo-community.org/web?debug=true#id=5&action=231&model=project.project&view_type=form&menu_id=154 )

- It looks like project_recalculate and project_timeline add the same fields two times

rafaelbn:
@ernestotejeda even with debug mode (developer mode) activate user cannot see in the TASKS start date or end date in order to modify manually a single task. They should, check please:

rafaelbn:
@ernestotejeda The sequence of tasks is incorrect, "Installation" task starts in July but the the Kanban view the order is by sequence so you cannot see first the task that begins first. IMHO we could set a sequence by start date.  The task for example, "Follow-up socios potenciales" starts in June but in the kanban view doesn't appear as is in the lowest place

ernestotejeda:
> @ernestotejeda in the project form the fields Start date and Expiration date are duplicated

@rafaelbn  These fields are going to be repeated when the project_recalculate and project_timeline modules are installed at the same time. These two modules add to the project form view these two fields after the `Working Time` field.
A solution could be to set this module to depend on project_timeline, which is not essential for its operation but I don't think it would be bad. Providing a timeline view, for example, would allow to see quickly how a planning was made by this module.
Another solution I can think of is to add a new module to this repository that only aims to show the start and end dates of the project and perhaps those dates for each task in their form view. Then, project_recalculate and project_timeline, will depend on the new module, so there is no need for this project_recalculate  to depend on project_timeline. Perhaps there are cases in which the timeline view does not interest or it's prefered another instead of this one. The problem with this solution is that I don't know what repercussion to add a new module has for the repository. Or what repercussion it has to set project_timeline to depend on this module now that it is already migrated and that someone could be using it.

ernestotejeda:
> @ernestotejeda even with debug mode (developer mode) activate user cannot see in the TASKS start date or end date in order to modify manually a single task...

@rafaelbn The only way I have found to edit those fields in the form view is by installing the project_timeline module, which have the same problem mentioned in the previous comment..

ernestotejeda:
> @ernestotejeda The sequence of tasks is incorrect, "Installation" task starts in July but...

@rafaelbn , yes, I think it could be ordered by start_date, what do you think @pedrobaeza ?

pedrobaeza:
No, the order in the kanban is one different from the "time" order. You have timeline view for this purpose.

chienandalu (commented):
@ernestotejeda Please resolve done conversations to ease review of what's left. Also, try to increase test coverage if possible

pedrobaeza:
Let's merge this as is and let improvements for a second phase, as this migration has been very time consuming without customer real uses for most of the cases.

PR #544, opened 2019-05-21
project_key: add test context key

ernestotejeda:
Cc @Tecnativa

### DIFF
--- a/project_key/tests/test_common.py
+++ b/project_key/tests/test_common.py
@@ -9,8 +9,10 @@ class TestCommon(TransactionCase):
     def setUp(self):
         super(TestCommon, self).setUp()
 
-        self.Project = self.env['project.project']
-        self.Task = self.env['project.task']
+        self.Project = self.env['project.project'].with_context(
+            test_project_key=True)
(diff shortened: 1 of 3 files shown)


project_key/tests/test_common.py
```diff
@@ -9,8 +9,9 @@ class TestCommon(TransactionCase):
     def setUp(self):
         super(TestCommon, self).setUp()
 
-        self.Project = self.env['project.project']
-        self.Task = self.env['project.task']
+        test_ctx = dict(test_project_key=True)
```
pedrobaeza: Don't do it this way, as you are replacing full context dictionary passing a dictionary argument to next lines.

pedrobaeza (changes requested):
The required attribute is being force through a post_init_hook:

https://github.com/OCA/project/blob/12.0/project_key/hooks.py#L9

Remove that line and put the `required` attribute on views.

ernestotejeda:
Changes done

pedrobaeza (approved):
Fast-tracking this for solving compatibility problems in the rest of the modules tests. What a pain!

PR #503, opened 2019-04-18
[Add] Project Status Module

patrickrwilson:
This module adds Statuses to projects.

Original Discussion #500

PR #547, opened 2019-05-28
[11.0] [ADD] project_phase

dalonsod:
With this module you can define different project phases or stages.


carlosdauden (approved):

project_phase/models/project_project.py
```diff
@@ -0,0 +1,27 @@
...
+    sequence = fields.Integer(string='Sequence')
+
+
+class ProjectProject(models.Model):
+    _inherit = 'project.project'
+
+    @api.model
+    def _get_default_phase_id(self):
+        """ Gives default stage_id """
+        return self.env['project.phase'].search([('sequence', '=', '1')])
```
ValentinVinagre: I would change the function search for this:
search([], order='sequence', limit=1)
because if the search return more of one phases generates an error.

ValentinVinagre (commented):
Only comments...

project_phase/models/project_project.py
oihane: It is done as in `crm_claim` module, there is no problem changing it, LGTM

dalonsod: Done

ValentinVinagre (approved):
LGFM

lmiguens-solvos (approved):
LGTM

project_phase/__manifest__.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2018 Oihane Crucelaegui - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    "name": "Project Phase",
+    "version": "11.0.1.1.0",
```
nikul-serpentcs: @dalonsod If a new Module then` "version": "11.0.0.0.0",`

And Add `Summary` key

nikul-serpentcs (commented):
Nitpicking

project_phase/__manifest__.py
oihane: @nikul-serpentcs He has to change it to "11.0.1.0.0" and summary is not mandatory. The thing is that this module has been migrated from other repo.

dalonsod: As said [here](https://github.com/OCA/account-invoicing/pull/533#discussion_r288586399) I've only changed version number to 11.0.1.0.0

nikul-serpentcs (approved):
Code LGTM :+1:

project_phase/models/project_project.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2018 Oihane Crucelaegui - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import api, fields, models
+
+
+class ProjectPhase(models.Model):
```
daramousk: Models should be defined in their own files. https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#15file-naming

daramousk (approved):
Looks great, although I think it would be a little better being named as stage rather than phase since Odoo follows this convention for documents that go through different stages (eg. tasks have `stage_id` field.)

Other than that works great, thanks!

project_phase/models/project_project.py
dalonsod: @oihane what do you think about file naming comment? Should I make that refactoring?

oihane: odoo has named stages (no specified that they are for tasks) and that's why they are named project.phase and not stage. Otherwise putting each object in a different model/file, we do not see any problem.

cc @anajuaristi

dalonsod:
@pedrobaeza @rafaelbn is this ready to merge?

nikul-serpentcs:
@dalonsod Squash Commit's

HaraldPanten (approved):
Functional review 👍

dalonsod:
> 
> 
> @dalonsod Squash Commit's

Done!

rafaelbn:
Hi @dalonsod , thanks for this contribution.

Which is the difetence between this module and project_status : https://github.com/OCA/project/tree/12.0/project_status

@patrickrwilson could you give a review here please?

rafaelbn (changes requested):
I think this module makes the same has project_status made before: https://github.com/OCA/project/pull/503

So the best approach is a back-port IMHO

dalonsod:
@rafaelbn I think you're right :(
Anyway, still waiting for @patrickrwilson and @oihane reviews

patrickrwilson:
@dalonsod I agree with @rafaelbn that this seems to have the same functions as the project_status module for v12 and that should be fairly straight forward to back port with little effort to keep things consistent.

rafaelbn:
Hi all, as this functionality is duplicated I close this PR. Please @oihane if you answer we can re-open gladly :smile:

PR #550, opened 2019-07-05
[12.0][IMP] project_role: revert manual options in favor of web_m2x_options

alexey-pelykh:
@pedrobaeza what do you think of this?

### DIFF
--- a/project_role/README.rst
+++ b/project_role/README.rst
@@ -53,6 +53,10 @@ To manage assignments:
 #. Go to *Project > Assignments*
 #. Add/remove assignments as needed
 
+Also, it's recommended to consider using ``web_m2x_options`` module in order to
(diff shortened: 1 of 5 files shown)


pedrobaeza:
Well, you can leave it this way and take note for newly modules, but don't mind removing it again.

alexey-pelykh:
I won't be able to sleep good if I'll leave it as is :)

pedrobaeza (approved):
OK then, all for your sleep health!

pedrobaeza:
But please indicate also that the system parameter should be put.

alexey-pelykh:
Added to the readme

rafaelbn (approved):

rafaelbn:
Please don't forget this https://github.com/OCA/project/pull/452#issuecomment-487146507 for v13 :smile:

PR #553, opened 2019-07-11
[IMP] project_timeline: scss instead of less

sbidoul:
`lessc` is not a mandatory dependency of Odoo 12, so let's use scss instead.


rvalyi (approved):

pedrobaeza:
You can't do this change without switching syntax

sbidoul:
@pedrobaeza what should I change exactly? This file looks like regular css to me.

pedrobaeza:
OK, if there's no variable, then call it simply css, not scss, but it's weird that it's called previously less without having less variables.

sbidoul:
I don't know. Looks like a tradition to name these files scss everywhere. Admitedly I've no idea what I'm doing here :)

pedrobaeza:
Indeed the file contains plain CSS, so I still think it should be called `.css`

pedrobaeza:
Let's invoke @Yajo that knows about this stuff.

yajo (approved):
The fix is OK, in this case it's only using syntax that is shared between Less and Scss.

BTW the file is not valid CSS! It seems you guys are so used to Less/Scss that can't distinguish it from CSS now... :yum: This is the CSS it produces when compiled:

```css
.o_project_timeline_item small[name='planned_hours'] {
  font-size: 85%;
}
```

pedrobaeza:
And isn't better then to put the plain CSS and avoid any compilation?

yajo:
No, let the Scss. It's more maintainable and readable. I doubt you're gonna save 1ms.

pedrobaeza:
Thanks for the clarifications, @Yajo, and thanks @sbidoul for proposing this improvement. Merging!

PR #559, opened 2019-08-08
[11.0][UPD] travis.yml

MiquelRForgeFlow:
Greenify repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ sudo: false
 cache: pip
 
 addons:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -28,13 +29,9 @@ env:
   - TRANSIFEX="1"
```
pedrobaeza: Adapt this part also

MiquelRForgeFlow: done

pedrobaeza (approved):

PR #562, opened 2019-08-12
[11.0][IMP] project_category

aln-bmya:
update for newest


rafaelbn:
Hi @aln-bmya , thanks for this contribution.

Please:

- Explain what the PR makes, why, etc.... (sometimes a screenshot is a good idea)
- See that I have change you tittle in order to be more descriptiveso reviewer can select better

Thanks

tarteo (commented):
I don't understand what the purpose is of this PR is. If you want to remove some deprecated functionality instead of commenting it you can remove it entirely

dreispt (changes requested):
Please explain change, and remove rather than comemnting code.

PR #563, opened 2019-08-12
[12.0][IMP] project_role: hierarchy, archived roles and assignments

alexey-pelykh:
After few discussions, concept of roles has been improved to more cover needs of complex project task management.

### DIFF
--- a/project_role/models/project_assignment.py
+++ b/project_role/models/project_assignment.py
@@ -1,5 +1,5 @@
-# Copyright 2018 Brainbean Apps (https://brainbeanapps.com)
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+# Copyright 2018-2019 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models, _
 from odoo.exceptions import ValidationError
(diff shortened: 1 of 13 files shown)


alexey-pelykh:
@mreficent @astirpe @pedrobaeza I believe you may be interested in this improvement

project_role/models/project_assignment.py
```diff
@@ -18,50 +21,111 @@ class ProjectAssignment(models.Model):
...
-    @api.constrains('role_id', 'user_id')
-    def _check_assignable(self):
+    @api.constrains('company_id', 'project_id', 'role_id', 'user_id')
+    def _check(self):
         for assignment in self:
+            conflicting_assignment = self.search(
+                assignment._get_conflicting_domain(),
+                limit=1,
+            )
+            if conflicting_assignment:
```
tarteo, line 122: Isn't this redundant considering the constraint. Or do I misunderstand it :smile: maybe

tarteo (approved):
Nice addition / improvement! LGTM.

project_role/models/project_assignment.py
alexey-pelykh: It's only a bit redundant, in SQL constraint it's hard (if even possible) to express logic like "if there's global assignment of user U to role R, don't allow project-specific assignment". So I could've removed SQL constraints but I think it's good to keep things consistent as much as possible on database level as well :)

rafaelbn (approved):

PR #568, opened 2019-08-20
[MIG] project_hr: Migration to v12

martintorresvm:


project_hr/README.rst
```diff
@@ -53,13 +53,13 @@ Configuration
 
 For setting employee categories at project level:
 
-#. Go to *Project > Dashboard*.
-#. Click on the 3 vertical dots for unfolding options and select "Settings".
+#. Go to *Project > Configuration > Settings*.
```
pedrobaeza: This is not correct, and previous steps were correct. The only changed thing is that instead of `Dashboard`, main menu is now called `Projects` (*Project > Projects* in summary)

project_hr/README.rst
```diff
@@ -53,13 +53,13 @@ Configuration
 
 For setting employee categories at project level:
 
-#. Go to *Project > Dashboard*.
-#. Click on the 3 vertical dots for unfolding options and select "Settings".
```
pedrobaeza: If not specifically enough, rephrase it to (also changing the menu name):
```
#. Click on the 3 vertical dots of one of the project kanban cards for unfolding options and select "Edit".
```

project_hr/static/description/index.html
```diff
@@ -3,7 +3,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
-<meta name="generator" content="Docutils 0.15.1: http://docutils.sourceforge.net/" />
+<meta name="generator" content="Docutils 0.14: http://docutils.sourceforge.net/" />
```
pedrobaeza: Regenerate it with new library

project_hr/tests/test_project_hr.py
```diff
@@ -59,6 +59,7 @@ def test_task(self):
             'user_id': self.user_2.id,
             'category_ids': [(6, 0, self.hr_category.ids)],
         })
+        self.task._compute_allowed_user_ids()
```
pedrobaeza: This mustn't be called manually. If you need to do this, something in the depends is not correct. Other option is that certain operations must call a cache invalidation.

pedrobaeza (changes requested):

martintorresvm:
Changes done @pedrobaeza

project_hr/tests/test_project_hr.py
```diff
@@ -59,6 +59,7 @@ def test_task(self):
             'user_id': self.user_2.id,
             'category_ids': [(6, 0, self.hr_category.ids)],
         })
+        self.task.invalidate_cache()
```
pedrobaeza: You are invalidating cache here, but you should do it when creating the employee on regular code. If not, on a production database, computed field is not going to be refreshed.

martintorresvm:
Changes done @pedrobaeza

project_hr/models/hr_employee.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class HrEmployee(models.Model):
+    _inherit = 'hr.employee'
+
+    @api.model
+    def create(self, vals):
+        employee = super().create(vals)
+        if employee.category_ids:
+            tasks = self.env['project.task'].search([
```
pedrobaeza: No need of searching involved tasks. `invalidate_cache` works at model level (it's `@api.model`).

pedrobaeza: And what happens when you modify category_ids of an employee? Add a test to check if you need to invalidate cache as well.

pedrobaeza (changes requested):

project_hr/models/hr_employee.py
martintorresvm: Think already have https://github.com/OCA/project/pull/568/files?file-filters%5B%5D=.py#diff-21c8ed16cfa23508bf379ef5143fdc4eR45

martintorresvm:
Changes done @pedrobaeza

project_hr/models/hr_employee.py
pedrobaeza: No, do it a second time on same test and it will be then the proper test, as first time there's no value cached.

martintorresvm:
Changes done @pedrobaeza

pedrobaeza (approved):

pedrobaeza:
@ernestotejeda please review

ernestotejeda (approved):
Functionally OK
Code OK

rafaelbn:
ocabot /merge

PR #569, opened 2019-08-21
[12.0][ADD] project_deadline

tarteo:
Currently the project_timeline shows the hidden date and date_start on the project form view. It's a problem for customers who don't want the timeline view (yes they exists) but want to specify deadlines on projects. That's why I created an extra module project_deadline. ~~With this PR I also changed the dependencies in the project_timeline module.~~

Features:

* Change start date and deadline of projects
* Shows the deadline in the project kanban view
* Show the start date and deadline of the project on the task form view


pedrobaeza:
I don't agree on changing the dependencies of `project_timeline`. What you can do is to overwrite `fields_view_get` in both modules, and add by code the fields in the view if they don't exist yet.

tarteo:
@pedrobaeza I understand your point. Let's see what other think of it. Thanks.
IMO the view in another module + dependency in others is not a problem. To migrate the module it'll take not much time.

pedrobaeza:
@tarteo this doesn't depend on what others think. The module has been merged in the branch since a long. You can't break all the instances that have is installed forcing a new dependency that might not even be in their addons path (if they don't add massively all the folder, like my company does).

Then, the only option for having this accepted is the way I'm telling you to do.

tarteo:
Sure @pedrobaeza.

tarteo:
@pedrobaeza Are you happy with this? I moved it to `fields_view_get`. There are no tests yet, I'll make them when you're happy to prevent wasting time.

pedrobaeza:
Yes, thanks!

project_deadline/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
alexey-pelykh, line 1: I still prefer to have license in these files, yet that's just a comment

alexey-pelykh (approved):
Code LGTM

rafaelbn:
Hi @tarteo , could you add an Usage section in order to test it? Let me know thanks!

tarteo:
@rafaelbn Hi, I've added a usage section in the readme.

tarteo:
@rafaelbn Can you retest it, please?

rafaelbn (approved):

PR #570, opened 2019-08-26
[12.0][ADD] project_task_project_required

alexey-pelykh:
This module allows to configure task as having project as a mandatory field.

### DIFF
--- a/project_task_project_required/__manifest__.py
+++ b/project_task_project_required/__manifest__.py
@@ -1,0 +1,21 @@
+# Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Project: require Project on Task',
+    'summary': """
+        Set project on task as a mandatory field""",
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author':
+        'Brainbean Apps, '
+        'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/project',
+    'depends': [
+        'project',
+    ],
+    'data': [
+        'views/project_task.xml',
+        'views/res_config_settings.xml',
+    ],
+}
--- a/project_task_project_required/models/project_task.py
+++ b/project_task_project_required/models/project_task.py
@@ -1,0 +1,22 @@
+# Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models, _
+from odoo.exceptions import ValidationError
+
+
+class ProjectTask(models.Model):
+    _inherit = 'project.task'
+
+    is_project_required = fields.Boolean(
+        string='Is Project Required',
+        related='company_id.is_project_task_project_required',
+    )
+
+    @api.constrains('project_id')
+    def _check_project_id(self):
+        for task in self:
+            if task.is_project_required and not task.project_id:
+                raise ValidationError(_(
+                    'You must specify a project for the task.'
+                ))
--- a/project_task_project_required/models/res_config_settings.py
+++ b/project_task_project_required/models/res_config_settings.py
@@ -1,0 +1,14 @@
+# Copyright 2018-2019 Brainbean Apps (https://brainbeanapps.com)
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    is_project_task_project_required = fields.Boolean(
+        string='Require Projects on Tasks',
+        related='company_id.is_project_task_project_required',
+        readonly=False,
+    )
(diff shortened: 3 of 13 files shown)


project_task_project_required/models/res_config_settings.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = 'res.config.settings'
+
+    is_project_task_project_required = fields.Boolean(
+        string='Require Projects on Tasks',
+        related='company_id.is_project_task_project_required',
+        readonly=False,
```
tarteo, line 13: Isn't this the default for related fields?

tarteo (approved):
Very nice!

project_task_project_required/models/res_config_settings.py
alexey-pelykh: Sinve v12 default is `readonly=True`, see https://github.com/odoo/odoo/blob/897feb553e6ef75af9434cb77436b26abde3a761/odoo/fields.py#L427

tarteo: I see, thanks :+1:

project_task_project_required/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'summary': """
+        Set project on task as a mandatory field""",
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author':
+        'Brainbean Apps, '
+        'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/project',
+    'depends': [
+        'project',
```
Saran440, line 15:
```suggestion
        'project',
        'hr_timesheet',
```
Error when I not install hr_timesheet module.

project_task_project_required/models/project_task.py
```diff
@@ -0,0 +1,22 @@
...
+
+    is_project_required = fields.Boolean(
+        string='Is Project Required',
+        related='company_id.is_project_task_project_required',
+    )
+
+    @api.constrains('project_id')
+    def _check_project_id(self):
+        for task in self:
+            if task.is_project_required and not task.project_id:
```
Saran440, line 19: I don't know how can it happen in case?

Saran440 (commented):
Functional Test  :+1: 
and Code Review

project_task_project_required/models/project_task.py
alexey-pelykh: Odoo allows creating tasks and unsetting project_id after task has been created, so this check is the basically very essence of the module.

project_task_project_required/__manifest__.py
alexey-pelykh: Fixed origin of the issue

project_task_project_required/models/project_task.py
Saran440: My steps test
1.Go to Project -> All Task -> Create Task (don't depend project_id)
2.Go to Project -> Configuration -> check `Require Projects on Tasks`
3.Back to Task -> Project will required field.

Project is required field already and why are you want to check this function? or any case that used this function.

alexey-pelykh: That's marking UI field as required, it has nothing to do with function that ensures data integrity if I want to create task without project programmatically

alexey-pelykh:
Right now I have no idea why this build started to fail due to field that is not referenced anywhere:
```
Field 'allowed_user_ids' used in attributes must be present in view but is missing:
- 'allowed_user_ids' in domain="[('id', 'in', allowed_user_ids)]"
```

alexey-pelykh:
For me it's being green: https://travis-ci.org/brainbeanapps/project/builds/578649199
In OCA it fails: https://travis-ci.org/OCA/project/builds/578649217

alexey-pelykh:
@OCA/project-service-maintainers I guess this needs some high-level attention, if possible. 
```
<field name="allowed_user_ids" invisible="1"/>
```
adding this fixed OCA Travis, but breaks Runbot and our Travis. Not having that line (which seems really correct to me) is perfectly fine for Runbot and our Travis, but breaks OCA Travis. Any ideas?

Saran440 (approved):

NL66278:
@alexey-pelykh As far as I can see the field is defined here: project_hr/models/project_task.py:34. So your module should depend on project_hr.

alexey-pelykh:
@NL66278 thanks for a hit, I have no idea how I've missed it, I've done rgrep on all fresh OCA repos and rebased. Yet I'm not using that field, and this module has no relation at all with `project_hr` - basically I'll force to install unneeded module in this case. Could it be a not quite correct approach in `project_hr` to set domain on field instead of setting it on views?

alexey-pelykh:
@pedrobaeza that module is related to you, I believe. Any ideas on that?

rafaelbn:
Hi, thanks for this contribution and fixing

I'm not sure if [[FIX] project_hr: domain on field breaks other modules](https://github.com/OCA/project/pull/570/commits/3f4d22200659848b2c364ce6b587f519fbd70305) should be fixed in a separate PR

@pedrobaeza is out of office

alexey-pelykh:
@rafaelbn easily workaroundable https://github.com/OCA/project/pull/573

rafaelbn:
Please @alexey-pelykh https://github.com/OCA/project/pull/573 is merged. Rebase

alexey-pelykh:
@rafaelbn all good

rafaelbn (approved):

PR #573, opened 2019-09-01
[12.0][FIX] project_hr: domain on field breaks other modules

alexey-pelykh:
W/o this fix any module that would inherit views that include `user_id` field of `project.task` and won't depend on `project_hr` - will have failing builds

### DIFF
--- a/project_hr/views/project_task_views.xml
+++ b/project_hr/views/project_task_views.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Copyright 2018 Tecnativa - Pedro M. Baeza
+     Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 
 <odoo>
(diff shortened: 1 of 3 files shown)


yajo:
Code looks good. May you explain why exactly this is needed?

alexey-pelykh:
@Yajo sure, see https://github.com/OCA/project/pull/570#issuecomment-526454611

project_hr/__init__.py
```diff
@@ -1,3 +1,3 @@
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
```
yajo: Please keep the diff minimal and avoid this change, which adds no real value. The previous link works perfectly fine.

The same across all files.

project_hr/__manifest__.py
```diff
@@ -1,11 +1,11 @@
 # Copyright 2018 Tecnativa - Pedro M. Baeza
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 {
     'name': 'Project HR',
     'summary': "Link HR with project",
     'development_status': 'Production/Stable',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
```
yajo: This isn't wrong, but nowadays the bot will bump the version number automatically, so it is not really needed.

project_hr/views/project_task_views.xml
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!-- Copyright 2018 Tecnativa - Pedro M. Baeza
-     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
-
+<!--
+    Copyright 2018 Tecnativa - Pedro M. Baeza
+    Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+    License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+-->
```
yajo: Please do not reformat this code chunk. Just add a line with your copyright if you want.

This was following what's OCA standard, [as you can find in the official module template](https://github.com/OCA/maintainer-tools/blob/771508e99876115c418888bbac7899174662482b/template/module/views/report_name.xml#L2-L3).

yajo (changes requested):
I think the patch is OK, but there's just some unnecessary diff to remove.

Thanks!

ernestotejeda:
Why remove domain from the model (https://github.com/OCA/project/pull/573/commits/676af34d31f8c1e3196a8e146475110465152fff#diff-001090485663461057b077232b3f33b1L25) and add it to the view (https://github.com/OCA/project/pull/573/commits/676af34d31f8c1e3196a8e146475110465152fff#diff-13c4d9368e1fc91cc0fc7dda883e1961R18) ?
Isn't enough adding `allowed_user_ids` to the views quick_create_task_form (https://github.com/OCA/project/pull/573/commits/676af34d31f8c1e3196a8e146475110465152fff#diff-13c4d9368e1fc91cc0fc7dda883e1961R33) and view_task_tree2 (https://github.com/OCA/project/pull/573/commits/676af34d31f8c1e3196a8e146475110465152fff#diff-13c4d9368e1fc91cc0fc7dda883e1961R47)

alexey-pelykh:
Mainly because having domain on a field requires all views that refer this field to depend on `project_hr` and include the field manually. This is not good at all.

pedrobaeza:
OK for the patch.

rafaelbn (approved):

PR #579, opened 2019-09-17
[IMP] project_task_pull_request: Track PR URI changes

yajo:
@Tecnativa

### DIFF
--- a/project_task_pull_request/models/project_task.py
+++ b/project_task_pull_request/models/project_task.py
@@ -10,6 +10,7 @@ class ProjectTask(models.Model):
 
     pr_uri = fields.Char(
         string='PR URI',
+        track_visibility='onchange',
     )
 
     pr_required_states = fields.Many2many(
--- a/project_task_pull_request/static/description/index.html
+++ b/project_task_pull_request/static/description/index.html
@@ -420,6 +420,7 @@ If you spotted it first, help us smashing it by providing a detailed and welcome
 <li>Kelly Lougheed &lt;<a class="reference external" href="mailto:kelly&#64;smdrugstore.com">kelly&#64;smdrugstore.com</a>&gt;</li>
 <li><a class="reference external" href="https://www.tecnativa.com">Tecnativa</a>:<ul>
 <li>Ernesto Tejeda</li>
+<li>Jairo Llopis</li>
 </ul>
 </li>
 </ul>
--- a/project_task_pull_request/README.rst
+++ b/project_task_pull_request/README.rst
@@ -73,6 +73,7 @@ Contributors
 * `Tecnativa <https://www.tecnativa.com>`_:
 
   * Ernesto Tejeda
+  * Jairo Llopis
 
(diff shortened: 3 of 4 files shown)


project_task_pull_request/static/description/index.html
```diff
@@ -3,7 +3,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
-<meta name="generator" content="Docutils 0.15.1: http://docutils.sourceforge.net/" />
+<meta name="generator" content="Docutils 0.14: http://docutils.sourceforge.net/" />
```
pedrobaeza: Please update your local docutils library for not having this changed

project_task_pull_request/models/project_task.py
```diff
@@ -10,6 +10,7 @@ class ProjectTask(models.Model):
 
     pr_uri = fields.Char(
         string='PR URI',
+        track_visibility='onchange',
```
pedrobaeza, line 13: URLs are properly parsed?

yajo: No idea, but I don't think that's relevant for this case. :thinking:

pedrobaeza: Well, we want it for being clickable, so not having it it's not very useful

yajo: Right now this happens:

Captura de pantalla de 2019-09-17 12-24-44

Yeah it should be clickable :thinking:

pedrobaeza: That's what I suspect, so I'm not sure this is very helpful for what we want.

yajo: I'm digging through the code and it's not gonna be straightforward to find the place where to fix that.

Could we consider this an improvement and leave the clickability for the future?

pedrobaeza: Well, if you can do both together, we can then change the procedure. Meanwhile, this as is doesn't serve for our purpose.

yajo: If you recieve the notification via email, though, I think yes it will be clickable, because Gmail does that for you.

yajo: Also my purpose was more avoiding to write twice the link, that can be done with this as it is, as you can still click it in the field.

pedrobaeza: Yes, but I don't want to need to enter Odoo for clicking to the URL. If Gmail parses it (I'm doubtful about this), then it's enough.

rafaelbn (approved):
:+1:

rafaelbn:
@Tardo please review thank!

Tardo (approved):
LGTM :+1: I think that links feature needs be implemented in other module

PR #583, opened 2019-09-30
[12.0] Add addons summary table to main README file

dreispt:


pedrobaeza:
I removed this because now the list in the folder list itself, and this was creating extra commits with no added value.

dreispt:
I'm sad to hear that. Was there a discussion on that?
The folder list does not provide the same information, so it's an inferior, and not effective alternative.

pedrobaeza:
I put a mail on the contributors mailing list when creating 11.0 branches and nobody said anything about that. We have the OCA apps store for "superior" search alternative and I prefer it, but any PSC can decide in their repos any way.

dreispt:
OK. I'm PSC here, so let's go ahead with this.
I feel that we could discuss if this should be brought back again, but it can be discussed at a later time.

emagdalenaC2i:
Glad to see this PR and this discussion. 
As I said in https://github.com/OCA/oca-github-bot/issues/48 I think this is very usefull 
But I don't know if there will be too much noise in commits, so if you don't mind we could check and if there are too much noise change again

emagdalenaC2i (approved):

emagdalenaC2i:
@sbidoul Could you review this PR?
As whe mentioned in https://github.com/OCA/oca-github-bot/issues/48 this will add the addons table again
@pedrobaeza Please, tell us if you are not agree. If there are too much noise we could change again

pedrobaeza:
Yes, I don't like it, but PSC can proceed as they think.

sbidoul (approved):

sbidoul:
LGTM. The bot should populate it as soon as it is merged. Let's see.

sbidoul:
Yes: https://github.com/OCA/project/commit/e4e01ae315931edc1db4f76da84fdddeedd89473

emagdalenaC2i:
I must say it again... This table is very helpfull

pedrobaeza:
And I must say it also: for me it's noise in everything: commits, repo, etc

emagdalenaC2i:
Ok @pedrobaeza  ;-)
Let's see that noise with #604

pedrobaeza:
I don't get the link between that and this.

emagdalenaC2i:
Yes, there is no new version in that PR

pedrobaeza:
Because it's added after...

dreispt:
Thank you Stéphane.
Discoverability is currently one of our biggest issues.
This table is a good step in the right direction.
I also think that we could go further, and make it more informative. In particular, I miss cross version availability information. But that's another discussion...

emagdalenaC2i:
@pedrobaeza #603 merged and I didn't see that noise you mentioned
Please, could you explain what that "noise" is?
I'm sorry if I insist on this issue, I just want everyone to agree with the solution adopted

pedrobaeza:
This commit isn't noise?: https://github.com/OCA/project/commit/1bfb6943a006cb3be87d5b020e666b6d058b987b

emagdalenaC2i:
Yes, it is. In someway.
But are these commits also noise? https://github.com/OCA/project/commit/7e22f9cd4611ea37637ecc30f2ee5e5c5dea1146 https://github.com/OCA/project/commit/90b107b2c57c547bd5e7121d50fda216595f8518 https://github.com/OCA/project/commit/480608355a06453075645c5a60bda8da0d3c395e
https://github.com/OCA/project/commit/d03884946f464bfa2d6307c2b7067ed8b010aeee

pedrobaeza:
Yes, they are, but the balance profit/loss in that case is positive, as we get something needed for the module itself (well, kind of - better to not enter again in the pip debate -). And they can be avoided if the contributor does everything in their migration commit (I do). I also advice in the migration steps to squash that commits in the next version migration.

But for this I don't see that benefit, as the need was when we have mixed branches with uninstallable/installable modules and we didn't have https://odoo-community.org/shop. The discoverability is further better there and the list of addons is the list of directories in the repository. You only get a useless version and a summary that is not properly put in most modules.

But as said, that's my personal opinion.

emagdalenaC2i:
Ok, thanks for the explanation.

Now that I have your doubts clear. **Would it be possible for that awesome bot to merge that commits automatically?**
I admit that on many occasions I forget to merge translation commits, just like many other users.

The discoverability is better in https://odoo-community.org/shop of course, but this table is excellent for tracking the evolution of a repository and its different modules and versions

sbidoul:
> summary that is not properly put in most modules

They should :) If making them visible helps us getting better summaries, then that's an added value for that table.

> Would it be possible for that awesome bot to merge that commits automatically?

@emagdalenaC2i what do you mean with this?

pedrobaeza:
> They should :) If making them visible helps us getting better summaries, then that's an added value for that table.

The problem is in the summary concept itself. Few words are most of the times not enough for expressing the module intent, and if we have problems having contributors naming properly variables or putting README in English, imagine also for having the summarizing ability.

emagdalenaC2i:
If the bot could squash those commits when a PR is merged... then not only would the noise of the commit "[UPD] addons table in README.md" be silenced. But also everything else, and we would have a cleaner commit history

pedrobaeza:
@emagdalenaC2i that would mean to break for example PGP signed commits. It's allowed when migrating, as we are rewriting commits any way. That's why some of the contributors (including @sbidoul) don't want to clean the repo, but there are lot of other problems not cleaning the repo that we have already discussed in other threads.

sbidoul:
I know making good summaries is hard. Yet making them more visible could encourage contributors with good English / abstraction skills to improve them. One can dream, right? :)

pedrobaeza:
Hehe, indeed. If you see my wishlist...

sbidoul:
> squash those commits when a PR is merged

Squashing specific commits automatically is hard, and we want to respect contributors who create a good commit history. But the bot could maybe group some commits. I created https://github.com/OCA/oca-github-bot/issues/89 to track that.

There is also https://github.com/OCA/oca-github-bot/issues/86 and https://github.com/OCA/oca-github-bot/issues/43 about squashing in ocabot merge.

emagdalenaC2i:
> group some commits

That's de idea, not group all the commits

We are all dreaming here ... with better information in the repos, improve the bot, and don't mess it all (or as Archimedes said 'Do not disturb my circles' ;)

dreispt:
This is going a little off-topic, but oh well.
To be honest I don't see the point for this type of commit messages grooming.
The attrition it creates outweigh the eventual benefits the practice may have.

Note that I see the point for individual changesets with proper descriptions - the point of  commit is to explain why a change was made.
I would have concerns if a commit mixed together a module change with these "administrative" changes.

But if the "administrative" changes are made in their own commit, they don't compromise the "why" explanation of the actual changes. They would be "white noise" for me - it's noise, but won't really be any real nuisance, for practical purposes.

PR #586, opened 2019-09-30
[IMP] project_timesheet_time_control: ❤️ 4 UX

yajo:
This is a huge improvement in this module's usability, although basic functionality stays the same. Let me summmarize:

- In `account.analytic.line` views, the *Stop* button is replaced by colorful *Resume/Stop* buttons.
- The new *Resume* button will display a wizard that will duplicate all meaningful fields from the previous `account.analytic.line` record and auto-stop the previous running timer, if any.
- If the previous running timer is very old, some warnings will be displayed.
- In `project.project` and `project.task` views, I added the same buttons (well... it's *Start* instead of *Resume*, because you could not have worked before in that project or task, but it behaves the same).
- If using it from a `project.task`, *Start* will duplicate meaningful fields from your latest line in that task.
- If using it from a `project.project`, *Start* will duplicate meaningful fields your latest line in that project that had no task associated.
- All kanban, kanban mobile, list and form views are covered.
- I removed, BTW, some Bootstrap 3 and bad replacements in some views.

---

@Tecnativa TT19205
Inspired by https://github.com/OCA/timesheet/pull/139.

### DIFF
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -3,33 +3,89 @@
 # Copyright 2016-2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from odoo import api, fields, models
 from datetime import datetime
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 
 
 class AccountAnalyticLine(models.Model):
     _inherit = 'account.analytic.line'
     _order = 'date_time desc'
 
-    date_time = fields.Datetime(default=fields.Datetime.now,
-                                string='Date time')
+    date_time = fields.Datetime(
+        default=fields.Datetime.now,
+        copy=False,
+    )
+    show_time_control = fields.Selection(
+        selection=[("resume", "Resume"), ("stop", "Stop")],
+        compute="_compute_show_time_control",
+        help="Indicate which time control button to show, if any.",
+    )
 
-    def eval_date(self, vals):
+    @api.model
+    def _eval_date(self, vals):
         if vals.get('date_time'):
-            vals['date'] = fields.Date.to_date(vals['date_time'])
+            return dict(vals, date=fields.Date.to_date(vals['date_time']))
         return vals
 
     @api.model
-    def create(self, vals):
-        return super(AccountAnalyticLine, self).create(self.eval_date(vals))
+    def _running_domain(self):
+        """Domain to find running timesheet lines."""
+        return [
+            ("date_time", "!=", False),
+            ("employee_id", "in", self.env.user.employee_ids.ids),
+            ("project_id.allow_timesheets", "=", True),
+            ("unit_amount", "=", 0),
+        ]
+
+    @api.model
+    def _duration(self, start, end):
+        """Compute float duration between start and end."""
+        try:
+            return (end - start).total_seconds() / 3600
+        except TypeError:
+            return 0
+
+    @api.depends("employee_id", "unit_amount")
+    def _compute_show_time_control(self):
+        """Decide when to show time controls."""
+        for one in self:
+            if one.employee_id not in self.env.user.employee_ids:
+                one.show_time_control = False
+            elif one.unit_amount or not one.date_time:
+                one.show_time_control = "resume"
+            else:
+                one.show_time_control = "stop"
+
+    @api.model_create_multi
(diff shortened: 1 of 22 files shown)


project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -3,33 +3,79 @@
...
+                one.show_time_control = False
+            elif one.unit_amount:
+                one.show_time_control = "resume"
+            else:
+                one.show_time_control = "stop"
+
     @api.model
     def create(self, vals):
-        return super(AccountAnalyticLine, self).create(self.eval_date(vals))
+        return super(AccountAnalyticLine, self).create(self._eval_date(vals))
```
Tardo:
```suggestion
        return super().create(self._eval_date(vals))
```

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -3,33 +3,79 @@
...
 
     @api.multi
     def write(self, vals):
-        return super(AccountAnalyticLine, self).write(self.eval_date(vals))
+        return super(AccountAnalyticLine, self).write(self._eval_date(vals))
```
Tardo:
```suggestion
        return super().write(self._eval_date(vals))
```

chienandalu (commented):
These improvements are great!

A little issue:

- Go to timesheet and create two lines succesively.
- The timer of the first one doesn't stop so the times overlap.

Is it intended?

Tardo (changes requested):

yajo:
> A little issue:
> 
>     * Go to timesheet and create two lines succesively.
> 
>     * The timer of the first one doesn't stop so the times overlap.
> 
> 
> Is it intended?

There's no constraint that forbids creating 2 lines with 0 time. Maybe I should add a unique constraint to avoid that?

pedrobaeza:
I don't think we should control that.

yajo:
That's what I though at first also. After having these controls in place, it's more likely that you create your lines from projects/tasks view than timesheet ones.

:thinking: Maybe I could add a button to create a line from scratch from the menu, opening the wizard without smar fills, to let user autostop latest line.

pedrobaeza:
Uhm, maybe we can make a similar screen to the attendance one with big button for starting and stopping.

yajo:
That'd be much harder than just adding a menu item that opens the wizard. It would involve JS. I think the menu item should be enough

pedrobaeza:
Well, let's put it on the roadmap. I don't see the menu too much added value.

yajo:
I thought about it and it makes sense. I added this menu: *Timesheets > Timesheet > Start work*

It opens this wizard, without autocompletes. In the screenshot you can see the obvious benefit of doing it this way instead of just creating a new record: it warns about an old running timer and lets you go to it or stop it and start a new timer with one click:

Captura de pantalla de 2019-10-01 12-31-38


So, although I think we will use it less because the new buttons in tasks and projects are so cool, I still think it's good to have it for those that want to input a new line from scratch.

After all, it's just an action and menu item, so easy it's done already.

Tardo (approved):
LGTM :+1: Awesome improvements!

pedrobaeza:
Please add Spanish translation

yajo:
Done. I just have to add some tests and try to ensure the wizard always features a hotkey to make it more handy.

yajo:
> try to ensure the wizard always features a hotkey

Enforcing a hotkey is very error-prone in a wizard, it's easier to rely on the automatic one, as far as there are available, so I opened https://github.com/OCA/timesheet/pull/278 and patched a little bit kanban and tree views here to add more chances for the wizard button to get a hotkey. Now it will have one almost always.

pedrobaeza:
Can you improve a bit more test coverage?

yajo:
Yes, it's on my roadmap, do not merge still please.

rafaelbn (approved):
Awesome! Functionally :+1:

yajo:
If I'm not wrong, this should go 100% :umbrella: now.

yajo:
Please do not merge yet. I still have to change one thing to make it compatible with the upcoming patch for `crm_timesheet`.

yajo:
Ready to merge.

rafaelbn:
@Yajo travis is :red_circle:

yajo:
:heavy_check_mark: now.

yajo:
Once merged let's continue with https://github.com/OCA/timesheet/pull/279.

PR #589, opened 2019-10-01
[MIG] project_task_project_required : Migration to 13.0

Saran440:


project_task_project_required/views/project_task.xml
```diff
@@ -33,12 +33,13 @@
...
     </record>
 
-    <record model="ir.ui.view" id="quick_create_task_form">
+    <record id="quick_create_task_form" model="ir.ui.view">
         <field name="name">project.task.form.quick_create</field>
         <field name="model">project.task</field>
         <field name="inherit_id" ref="project.quick_create_task_form"/>
         <field name="arch" type="xml">
             <field name="name" position="before">
+                <field name="company_id" invisible="1"/>
```
pedrobaeza, line 42: Why do you need this?

Saran440: @pedrobaeza In odoo13, field project_id have new attribute check_company but this view not have company_id

pedrobaeza: But isn't that an upstream problem?

Saran440: No, I think. upstream v12 not required but v13 is required company_id.

dreispt (approved):

dreispt:
Ping!

legalsylvain:
@Saran440 : could you review #884 (and cherry-pick the commit if you're ok with the change)

thanks !

PR #593, opened 2019-10-02
[WIP] 12.0 mig sale order project

daramousk:


sale_order_project/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Sale Order Project',
+    'version': '12.0.1.0.0',
+    'category': 'Generic Modules/Others',
+    'license': 'AGPL-3',
+    'author': 'Akretion, '
+              'AvanzOSC, '
+              'Serv. Tecnol. Avanzados - Pedro M. Baeza, '
```
rafaelbn, line 11: @pedrobaeza do you want to change?

sale_order_project/README.rst
```diff
@@ -0,0 +1,81 @@
...
+    :target: https://translation.odoo-community.org/projects/project-12-0/project-12-0-sale_order_project
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runbot-Try%20me-875A7B.png
+    :target: https://runbot.odoo-community.org/runbot/140/12.0
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module is a base module that give the possibility to create a project
+from a quotation, if a contract is set with use_tasks checked it will show
```
rafaelbn, line 29: Please don't use the work `contract` Now in v12 Account analytic is one thin a contract is another model

sale_order_project/README.rst
```diff
@@ -0,0 +1,81 @@
...
+the related project.
```
rafaelbn, line 30: how have you adapted this module to the changes in v12?

Now you could have 1 analytic account in the sale order but this is not realted with the project anymore because now 1 analytic account could be related with n project

In fact take care and test it the module functionally in depth ! You cannot create a task in an existing project any more in Odoo if you don't use a specific product per project :cry: 

A lot of changes in v12 in this part

rafaelbn (commented):
This module should improve the readme and re USAGE to explain that is needed only in the case you can to create the project before confirming the sale order

sale_order_project/__manifest__.py
pedrobaeza: Yes, please put `Tecnativa`

rafaelbn:
@daramousk , now you have project_id in the sale.order.line this is new in v12

daramousk:
@rafaelbn Thanks for your comments, I have been opening some PRs today some of which are WIP, I will mark this one as well

sale_order_project/README.rst
daramousk: Let me see if I get this straight:

> Now you could have 1 analytic account in the sale order but this is not realted with the project anymore because now 1 analytic account could be related with n project

I see, but how is this a problem? We register the sale on the same analytic account as the project and that's it.

> You cannot create a task in an existing project any more in Odoo if you don't use a specific product per project

While the `project.project` has a `sale_line_id` field, this can be left empty, it does not seem to be required. Shouldn't this be the best course of action for this module?

daramousk: Or perhaps this module should only create a project with the same account as the sale and thats were the functionality should end.

I am not really sure here, your opinion would be much appreciated. @rafaelbn

PR #595, opened 2019-10-12
13.0 mig project parent task filter

emagdalenaC2i:


emagdalenaC2i:
@astirpe @Saran440 @schout-it and @nikul-serpentcs could you review this PR?

astirpe (approved):
Tested in Runbot and it works fine!

Just one think, not blocking: the filters for parent_id should be visible only if user belongs to group "Use Subtasks".

emagdalenaC2i:
@astirpe Thanks, I will take a look to add that option later

schout-it (approved):
looks good, but in a future iteration look at the filters as @astirpe said.
showing filter options for something a user can't see on the views is just strange

project_parent_task_filter/views/project_task.xml
```diff
@@ -0,0 +1,47 @@
...
+             <field name="stage_id" position="after">
+                <field name="parent_id" string="Parent task"/>
+            </field>
+
+            <xpath expr="//filter[@name='rating_task']" position="after">
+                <filter string="Parent tasks" name="parent_only" domain="[('parent_id','=',False)]"/>
+            </xpath>
+
+            <xpath expr="//filter[@name='company']" position="after">
+                <filter name="parent_task" string="Parent tasks" domain="[]" context="{'group_by' : 'parent_id'}"/>
```
Saran440:
```suggestion
                <filter name="parent_task" string="Parent tasks" context="{'group_by' : 'parent_id'}"/>
```
You can delete domain function, If you don't used it.

Saran440 (approved):
Functional test. :+1:

emagdalenaC2i:
I don't know why now but travis says no...
Seems something about pre-commit
image
I will take a look later

PR #596, opened 2019-10-15
[IMP]  project_timesheet_time_control: Tell the user no running line was found

yajo:
Imagine this scenario:

1. In tab 1 of the browser, you have opened task 1.
2. In tab 2 of the browser, you have opened task 2.
3. You go to tab 1 and start a timer.
4. Work, work, work...
5. You go to tab 2 and start a timer, stopping that of task 1.
6. Work, work, work...
7. You go to tab 1 and see that timer as running (it is not, but you didn't refresh). You hit stop.

Before this commit, it just seemed like the timer was actually stopped. What did happen behind the scenes is that your view was refreshed, but no timer was touched fortunately.

After this commit, you get a message telling you that there's no timer to stop and that your browser is most likely out of sync. This mimics the behavior previously found when doing the same, but directly in the AAL. Now it's present in projects and tasks too.

@Tecnativa TT19205 https://github.com/OCA/timesheet/pull/280

### DIFF
--- a/project_timesheet_time_control/i18n/es.po
+++ b/project_timesheet_time_control/i18n/es.po
@@ -9,8 +9,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 10.0\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-10-02 09:29+0000\n"
-"PO-Revision-Date: 2019-10-02 10:31+0100\n"
+"POT-Creation-Date: 2019-10-15 09:35+0000\n"
+"PO-Revision-Date: 2019-10-15 10:55+0100\n"
 "Last-Translator: Jairo Llopis <yajo.sk8@gmail.com>\n"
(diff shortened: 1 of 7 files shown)


pedrobaeza (changes requested):
This is what I told you it was happening!

Please include POT + es.po updates as well, and I'll fast-track

pedrobaeza:
Also there's no possibility of avoiding code duplication?

yajo:
Maybe with a mixin. Do you think it's worth it?

pedrobaeza:
Well, I'm seeing a lot of code duplication, also in OCA/timesheet#280, so I think we can explore that possibility

yajo:
Can we merge and leave that for later? :roll_eyes:

pedrobaeza:
I prefer to do it now.

yajo:
OK, done.

pedrobaeza (approved):
Thank you very much. This is much lighter! Fast-tracking it as said.

PR #603, opened 2019-10-30
[12.0][IMP] project_role: possible roles

alexey-pelykh:
Allows depending modules to more consistently get list of roles that are available to particular user on particular project, taking into account inheritance and other advanced features.

### DIFF
--- a/project_role/__manifest__.py
+++ b/project_role/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     'name': 'Project Roles',
-    'version': '12.0.2.0.0',
+    'version': '12.0.2.1.0',
     'category': 'Project',
     'website': 'https://github.com/OCA/project',
     'author':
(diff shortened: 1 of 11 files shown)


pedrobaeza:
Explanation of the change?

alexey-pelykh:
@pedrobaeza added

alexey-pelykh:
@tarteo if possible, please review

pedrobaeza:
I'm feeling the module is getting very complicated for not such a difficult feature, but I can't judge. What I can ask is for 2 things:

- Add docstrings to methods for knowing what they do. For example, what `_check` does?
- Expand README for all the new stuff.

alexey-pelykh:
Technically, added method is exposing what is already there - company-wide assignments that were added as we discussed previous improve. Why it's added: because when other modules need list of roles that user can assume, they would need to manually implement company-wide assignments inheritance, and that would get error-prone.

alexey-pelykh:
@pedrobaeza adjusted readme and added some docstring on non-obvious methods

artemsenko (approved):

artemsenko:
Functional test LGTM

pedrobaeza (approved):

PR #610, opened 2019-11-12
[12.0][FIX]  project_wbs: minor fix in init_hook

MiquelRForgeFlow:
Fast-track.

Doesn't affect to who has already the module installed, as the fix is in the init_hook.

This patch makes sense as no need to override analytic accounts that already have a code.


AaronHForgeFlow (approved):
:+1: looks reasonable to me. Fast-tracking.

PR #615, opened 2019-11-28
[12.0][IMP] project_timeline: unset `date_end` safer

naglis:
We have a case where another addon removes `date_end` in `update_date_end()`, and when `update_date_end` from `project_timeline` is called it results in a `KeyError`. This makes the removal safer.


pedrobaeza (approved):
No need to change manually module version. There's now a bot for that.

Saran440 (approved):

PR #619, opened 2019-12-03
[12.0][UPD] project_wbs: add migration scripts to v12

MiquelRForgeFlow:
Fast-track.

No need to increase the number version when merge!


alexey-pelykh (approved):

alexey-pelykh:
@mreficent won't `oldname` do a better job?

MiquelRForgeFlow:
No, `oldname` does an awful job. That's why `rename_fields` method exists and that's why `oldname` is removed in v13.

alexey-pelykh:
Sounds reasonable 👍

PR #627, opened 2019-12-19
[FIX] project_template: New Name Error

patrickrwilson:
There is an error if the template name gets modified so needed to add a new else statement for the new_name.


patrickrwilson:
ping @osi-scampbell @max3903

max3903 (approved):

osi-scampbell (approved):

PR #631, opened 2020-01-02
WSJF prioritization model

gabrielcardoso21:


project_wsjf/__manifest__.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Kmee Informática LTDA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Project Wsjf',
+    'summary': """
+        This module implements the Weighted Shortest Job First prioritization model.
```
mileo, line 8: Can u please move the summary to the README.rst and fill the other questions like, Installation, Configure, Contributors and etc.. ?

And check the travis erros.

mileo (changes requested):

mileo:
More info about https://www.scaledagileframework.com/wsjf/

project_wsjf/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+    """,
+    'version': '12.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'Kmee Informática LTDA,Odoo Community Association (OCA)',
+    'website': 'www.kmee.com.br',
```
mileo: Change the website to https://github.com/OCA/project/

mileo (changes requested):

project_wsjf/__manifest__.py
dreispt: Summay should stay here. Description is to the replaced by the readme.

project_wsjf/models/project_task.py
```diff
@@ -0,0 +1,29 @@
...
+class ProjectTask(models.Model):
+    _name = 'project.task'
+    _inherit = [
+        'project.task',
+        'project.wsjf.mixin',
+    ]
+    _order = 'wsjf desc, sequence'
+
+    project_id = fields.Many2one(
+        inverse='_inverse_project_id',
```
dreispt: Can you explain this?
AFAIK project_id is not a computed field, why define an Inverse computation function for it?

project_wsjf/models/project_project.py
```diff
@@ -0,0 +1,23 @@
...
+class ProjectProject(models.Model):
+    _name = 'project.project'
+    _inherit = [
+        'project.project',
+        'project.wsjf.mixin',
+    ]
+    _order = 'wsjf desc, sequence'
+
+    internal_pressure = fields.Selection(
+        inverse='_inverse_internal_pressure'
```
dreispt: I see this is extending the mixin. But it is a not a computed field, why the inverse computation?

project_wsjf/README.rst
```diff
@@ -0,0 +1,91 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
dreispt: Ot would me nice to use the new readme structure, that automatically generates the top level README:
https://github.com/OCA/maintainer-tools/tree/master/template/module/readme

dreispt (changes requested):
Can you please fix the build?

PR #654, opened 2020-03-09
[10.0][IMP] enable quick search by code field for tasks

ntsirintanis:
An update for `product_task_code`, enabling the user to search for tasks using the `code` field. This was done by editing `name_search()`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 cache: pip
 
 python:
-  - "2.7.13"
+  - "2.7"
 
 addons:
   postgresql: "9.6"
@@ -23,6 +23,9 @@ env:
   - LINT_CHECK="1"
   - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1"
   - TESTS="1" ODOO_REPO="OCA/OCB"
+
+virtualenv:
+  system_site_packages: true
 
 install:
   - git clone --depth=1 https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
--- a/project_task_code/models/project_task.py
+++ b/project_task_code/models/project_task.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2016 Tecnativa <vicent.cubells@tecnativa.com>
+# Copyright 2020 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from openerp import api, fields, models, _
@@ -28,3 +29,41 @@ class ProjectTask(models.Model):
             default = {}
         default['code'] = self.env['ir.sequence'].next_by_code('project.task')
         return super(ProjectTask, self).copy(default)
+
+    @api.model
+    def name_search(self, name='', args=None, operator='ilike', limit=100):
+        args = args or []
+        tasks = []
+        code = self._name_search_parse_code(name)
+        if code:
+            # We are searching for codes
+            tasks = self.search(
+                [('code', '=ilike', code + '%')] + args,
+                limit=limit).name_get()
+            if limit and len(tasks) == limit:
+                return tasks
+        additional_args = args + [('id', 'not in', [i[0] for i in tasks])]
+        return tasks + super(ProjectTask, self).name_search(
+            name=name,
+            args=additional_args,
+            operator=operator,
+            limit=limit and limit - len(tasks) or limit)
+
+    @api.model
+    def _name_search_parse_code(self, name):
+        """ Check if user input looks like code """
+        # Remove any whitespace from name
+        name = name.lstrip()
+        # Get the data record
+        sequence_id = self.env.ref('project_task_code.sequence_task')
+        prefix = sequence_id.prefix or ''
+        # if user is typing only digits, return right away
+        if name.isdigit():
+            return prefix + name
+        # Check if input could be code
+        if (
+            name.startswith(prefix) and
+            name[len(prefix): len(name)].isdigit(),
+        ):
+            return name
+        return False
--- a/project_task_code/models/project_task.py
+++ b/project_task_code/models/project_task.py
@@ -38,7 +38,7 @@ class ProjectTask(models.Model):
         if code:
             # We are searching for codes
             tasks = self.search(
-                [('code', '=ilike', code + '%')] + args,
+                [('code', '=ilike', code)] + args,
                 limit=limit).name_get()
             if limit and len(tasks) == limit:
                 return tasks
--- a/project_task_code/tests/test_project_task_code.py
+++ b/project_task_code/tests/test_project_task_code.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2016 Tecnativa <vicent.cubells@tecnativa.com>
+# Copyright 2020 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 import openerp.tests.common as common
@@ -34,3 +35,16 @@ class TestProjectTaskCode(common.TransactionCase):
         project_task_copy = self.project_task.copy()
         self.assertNotEqual(project_task_copy.code, self.project_task.code)
         self.assertEqual(project_task_copy.code, code)
+
+    def test_name_search(self):
+        tasks = self.project_task_model
+        for i in range(10):
+            task = tasks.create({'name': 'task0%s' % str(i)})
+            tasks += task
+        for task in tasks:
+            self.assertEqual(
+                task,
+                tasks.browse(tasks.name_search(name=task.code)[0][0]))
+            self.assertEqual(
+                tasks.name_search(name=task.code),
+                tasks.name_search(name=task.name))
--- a/project_task_code/tests/test_project_task_code.py
+++ b/project_task_code/tests/test_project_task_code.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # Copyright 2016 Tecnativa <vicent.cubells@tecnativa.com>
-# Copyright 2020 Therp BV <http://therp.nl>
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+# Copyright 2020 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import openerp.tests.common as common
 
--- a/project_task_code/__init__.py
+++ b/project_task_code/__init__.py
@@ -5,6 +5,7 @@
 from . import models
 from odoo import api, SUPERUSER_ID
 
+
 def create_code_equal_to_id(cr):
     """
     With this pre-init-hook we want to avoid error when creating the UNIQUE
--- a/project_task_code/__manifest__.py
+++ b/project_task_code/__manifest__.py
@@ -1,10 +1,11 @@
 # -*- coding: utf-8 -*-
 # Copyright 2016 Tecnativa <vicent.cubells@tecnativa.com>
+# Copyright 2020 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
     "name": "Sequential Code for Tasks",
-    "version": "10.0.1.0.0",
+    "version": "10.0.1.0.1",
     "category": "Project Management",
     "author": "OdooMRP team, "
               "AvanzOSC, "
@@ -17,6 +18,7 @@
         "Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>",
         "Ana Juaristi <ajuaristo@gmail.com>",
         "Vicent Cubells <vicent.cubells@tecnativa.com>",
+        "Nikos Tsirintanis <ntsirintanis@therp.nl>",
     ],
     "depends": [
         "project",


ntsirintanis:
@hbrunn @gfcapalbo @NL66278

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,22 @@ def copy(self, default=None):
...
+            operator=operator,
+            limit=limit)
+        if not args:
+            args = []
+        tasks = self.env['project.task']
+        # Remove here a leading character, if name is not purely digits
+        name = name if name.isdigit() else name.lstrip('T')
+        # We are searching for codes
+        # operator is always 'ilike'
+        tasks = self.search([('code', operator, name)] + args, limit=limit)
```
hbrunn: I would expect this only to happen if `name` looks like a code

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,22 @@ def copy(self, default=None):
...
+            args=args,
+            operator=operator,
+            limit=limit)
+        if not args:
+            args = []
+        tasks = self.env['project.task']
+        # Remove here a leading character, if name is not purely digits
+        name = name if name.isdigit() else name.lstrip('T')
+        # We are searching for codes
+        # operator is always 'ilike'
```
hbrunn: you don't control that, the caller of the function controls that. And the way I envision this override, it should for the code part always search `('code', '=ilike', name + '%')`

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,22 @@ def copy(self, default=None):
             default = {}
         default['code'] = self.env['ir.sequence'].next_by_code('project.task')
         return super(ProjectTask, self).copy(default)
+
+    @api.model
+    def name_search(self, name='', args=None, operator='ilike', limit=100):
+        res = super(ProjectTask, self).name_search(
```
hbrunn: it's quite wasteful to always do two searches. I think the logic of this function should be something like

```python
result = None
if name.looks_like_code():
    result = our_search()
if not result:
    result = super()
return result
```

This way we usually only do one search, and only when something that looks like a code but isn't a code (and that should be the least usual case) is filled in, do both.

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,22 @@ def copy(self, default=None):
...
+            name=name,
+            args=args,
+            operator=operator,
+            limit=limit)
+        if not args:
+            args = []
+        tasks = self.env['project.task']
+        # Remove here a leading character, if name is not purely digits
+        name = name if name.isdigit() else name.lstrip('T')
```
hbrunn: if we make assumptions about how data looks, better do this in its own function. Then we for example can override this an change it to '#'.

I'd suggest `_name_search_parse_code(name)` which either returns the code to be searched for or something falsy, then `name` isn't a code

project_task_code/tests/test_project_task_code.py
```diff
@@ -34,3 +35,19 @@ def test_copy_task_code_assign(self):
...
+            task = tasks.create({'name': 'task0%s' % str(i + 1)})
+            code_list.append(task.code)
+            tasks += task
+        task_names = [task.name for task in tasks]
+        for code in code_list:
+            [(task_id, task_name)] = tasks.name_search(name=code)
+            self.assertIn(task_id, tasks.ids)
+            self.assertNotIn(task_id, all_tasks.ids)
+            self.assertIn(task_name, task_names)
+        self.assertEqual(tasks + all_tasks, tasks.search([]))
```
hbrunn: these tests are not very interesting. you should indeed test that it works to find a task by test, but if this woks one time or 10 times doesn't matter so much.

It matters much more that the test tests that searching for something not like a code yields the right products still (and only those).

hbrunn (changes requested):

project_task_code/models/project_task.py
hbrunn: (and btw. also sort by code descending, then we'll get higher numbers = newer tasks higher up, which probably is what we want)

NL66278:
Something else to consider: can codes and names overlap? Assume there is a code AB and names like Ability testing, Abuse prevention. Searching with ilike on codes and then returning will make you can not find the names starting with ab, AB or Ab. This is not a problem when names and codes cannot overlap. In case they can overlap, you can still search for code first, but then extend the search results with a pure name search, decreasing the search limit (if any!) by the amount of records already found for code.

project_task_code/models/project_task.py
hbrunn: another thought about this: I think what the parsing function can best do is look at https://github.com/OCA/project/blob/12.0/project_task_code/data/task_sequence.xml#L7, see if `name` starts with the prefix and suffix configured there, and the middle is only numbers.

This is not perfect because sequence prefixes can contain some variables, but we don't care about this for now.

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,36 @@ def copy(self, default=None):
...
+                limit=limit)
+            if tasks:
+                return tasks.name_get()
+        return super(ProjectTask, self).name_search(
+            name=name,
+            args=args,
+            operator=operator,
+            limit=limit)
+
+    def _name_search_parse_code(self, name):
```
hbrunn, line 53: please explicitly mark this as `@api.model`, and add a docstring

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,36 @@ def copy(self, default=None):
...
+        name = name.lstrip()
+        # Get the data record
+        sequence_id = self.env.ref('project_task_code.sequence_task')
+        prefix = sequence_id.prefix or ''
+        suffix = sequence_id.suffix or ''
+        # Check if it looks like code
+        if not all([
+            name.startswith(prefix),
+            name.endswith(suffix),
+            name.lstrip(prefix + suffix).isdigit(),
```
hbrunn: consider

```
>>> 'abc42bca'.lstrip('cab')
'42bca'
```

vs

```
>>> 'abc42bca'[len('abc'):-len('bca')]
'42'
```

The corner case is when the length of `suffix` is 0, for that you'll need to add an `or len(name)`

hbrunn (changes requested):

project_task_code/models/project_task.py
ntsirintanis: I do not understand your last sentence.

hbrunn:
```python
>>> prefix = 'pre'
>>> suffix = 'suffix'
>>> name = 'pre4242suffix'
>>> name[len(prefix):-len(suffix)]
'4242'
>>> suffix = ''
>>> name = 'pre4242'
>>> name[len(prefix):-len(suffix)]
''
>>> name[len(prefix):-len(suffix) or len(name)]
'4242'
```

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,46 @@ def copy(self, default=None):
...
+            operator=operator,
+            limit=limit + (limit - len(tasks)) or limit)
+
+    @api.model
+    def _name_search_parse_code(self, name):
+        """ Check if user input looks like code """
+        # Remove any whitespace from name
+        name = name.lstrip()
+        # if user is typing only digits, return right away
+        if name.isdigit():
```
gfcapalbo: proposal for internal particular case: or (name[0] == '#' and  name[1:].isdigit())  (so we cover also #1234).
quite specific, can ignore.

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,46 @@ def copy(self, default=None):
...
+            return name
+        # Get the data record
+        sequence_id = self.env.ref('project_task_code.sequence_task')
+        prefix = sequence_id.prefix or ''
+        suffix = sequence_id.suffix or ''
+        # Check if input could be code
+        if not all([
```
gfcapalbo: you check that user wrote prefix , number in the middle and suffix before allowing to return name?
if you have all three 

prefix, digit, and suffix,  the ALL will return TRUE, and the function will return false, so this works in the case of partial codes , but does not work if user writes T1234SUF   (the completed code....)

wich can happen often when user copy/pastes.

``
find a cleaner way to express this.

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,46 @@ def copy(self, default=None):
...
+        code = self._name_search_parse_code(name)
+        if not code:
+            return super(ProjectTask, self).name_search(
+                name=name,
+                args=args,
+                operator=operator,
+                limit=limit)
+        # We are searching for codes
+        tasks = self.search(
+            [('code', '=ilike', '%' + code + '%')] + args,
```
gfcapalbo: [('code', '=ilike',  code)] is OK. 

But it is case-insensitive. you may want to make it case insensitive with 'like' in case someone writes the prefix in the wrong case. (if you want to do that you must also change _name_parse_code to support that......)

the case-insensitive prefix seems a good usability improvement, i believe it is quite common for users to write case-insensitive prefix.

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,46 @@ def copy(self, default=None):
...
+            limit=limit)
+        if limit and len(tasks) == limit:
+            return tasks.name_get()
+        return tasks.name_get() + super(ProjectTask, self).name_search(
+            name=name,
+            args=args,
+            operator=operator,
+            limit=limit + (limit - len(tasks)) or limit)
```
gfcapalbo: error: limit = limit-len(tasks), you are actually increasing the limit in this case...... what you want is the limit of this search  to be   =  total limit  - the amount of results you have already found.

also how can you be sure this new name_search will not return results you have already found?
Of course the overloaded "+ on records" operator will eliminate duplicates for you, but your total results would be less than "limit". wich is not a good look.  we want consistent results.

Add to args:  [('id', 'not in',  tasks.ids)]

gfcapalbo (changes requested):
a couple of suggestions and 1-2 errors...
otherwise approved...

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,46 @@ def copy(self, default=None):
...
+
+    @api.model
+    def _name_search_parse_code(self, name):
+        """ Check if user input looks like code """
+        # Remove any whitespace from name
+        name = name.lstrip()
+        # if user is typing only digits, return right away
+        if name.isdigit():
+            return name
```
hbrunn: this is good, but return `prefix + name`. This way we can still search with only '%' which I prefer very much

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,46 @@ def copy(self, default=None):
             default = {}
         default['code'] = self.env['ir.sequence'].next_by_code('project.task')
         return super(ProjectTask, self).copy(default)
+
+    @api.model
+    def name_search(self, name='', args=None, operator='ilike', limit=100):
```
hbrunn, line 34: to avoid having to write the super call twice, better do

```python
tasks = []
...
if code:
    tasks = our_search.name_get()
    if limit and len(tasks) == limit:
        return tasks
return tasks + super()
```

project_task_code/models/project_task.py
hbrunn: right. We shouldn't care about the suffix at all, as it will never add new information to search, so people also won't type it. If name without prefix is a number, it's a code, that's it.

Thanks for this one!

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,40 @@ def copy(self, default=None):
...
+        # Get the data record
+        sequence_id = self.env.ref('project_task_code.sequence_task')
+        prefix = sequence_id.prefix or ''
+        # if user is typing only digits, return right away
+        if name.isdigit():
+            return prefix + name
+        # Check if input could be code
+        if (
+            name.startswith(prefix) or
+            name.lstrip(prefix).isdigit(),
```
hbrunn: still do the slicing, `lstrip` strips any permutation of the characters in `prefix`

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,41 @@ def copy(self, default=None):
...
+        name = name.lstrip()
+        # Get the data record
+        sequence_id = self.env.ref('project_task_code.sequence_task')
+        prefix = sequence_id.prefix or ''
+        # if user is typing only digits, return right away
+        if name.isdigit():
+            return prefix + name
+        # Check if input could be code
+        if (
+            name.startswith(prefix) or
```
hbrunn: this wants to be an `and`

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,41 @@ def copy(self, default=None):
...
+            name[len(prefix): -len(name)].isdigit(),
```
hbrunn:
```python
>>> prefix='hello'
>>> name='hello world'
>>> name[len(prefix): -len(name)]
''
```

project_task_code/tests/test_project_task_code.py
```diff
@@ -34,3 +35,13 @@ def test_copy_task_code_assign(self):
...
+
+    def test_name_search(self):
+        tasks = self.project_task_model
+        for i in range(10):
+            task = tasks.create({'name': 'task0%s' % str(i)})
+            tasks += task
+        for task in tasks:
+            self.assertEqual(
+                tasks.name_search(name=task.code),
+                tasks.name_search(name=task.name))
```
hbrunn, line 50: this test can succeed if both `name_search`es return an empty result by some weird bug in the code for example.

So when you assert things, always assert not only relations between things (they can be equally wrong), but also check something 'absolute'. In accounting you'd do indeed just numbers. Here, I think asserting that those searches are equal and return at least `task` seems prudent to me.

hbrunn (changes requested):
I just restarted the builds as they failed because of problems over there

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,41 @@ def copy(self, default=None):
...
+    def name_search(self, name='', args=None, operator='ilike', limit=100):
+        args = args or []
+        tasks = []
+        code = self._name_search_parse_code(name)
+        if code:
+            # We are searching for codes
+            tasks = self.search(
+                [('code', '=ilike', code + '%')] + args,
+                limit=limit).name_get()
+            args += [('id', 'not in', [i[0] for i in tasks])]
```
hbrunn: never change a mutable object you got passed as function parameter. I think this is also a weird place to do this. I'd assign this to a new variable before the last `return`

.travis.yml
```diff
@@ -24,6 +24,9 @@ env:
   - TESTS="1" ODOO_REPO="odoo/odoo" MAKEPOT="1"
   - TESTS="1" ODOO_REPO="OCA/OCB"
 
+virtualenv:
```
NL66278, line 27: I doubt this is the way to resolve this. We try to prevent the need for system site packages wherever possible to create a really isolated environment. @hbrunn you have an opinion on this one?

hbrunn: I'm the author of this. v10 uses a legacy version of python, so I also don't see a problem to use a legacy python option here

hbrunn (approved):

project_task_code/models/project_task.py
```diff
@@ -28,3 +29,41 @@ def copy(self, default=None):
...
+
+    @api.model
+    def name_search(self, name='', args=None, operator='ilike', limit=100):
+        args = args or []
+        tasks = []
+        code = self._name_search_parse_code(name)
+        if code:
+            # We are searching for codes
+            tasks = self.search(
+                [('code', '=ilike', code + '%')] + args,
```
NL66278: No need to add `'%'`. This is already implicit in the ilike operator. Unlike in plain SQL, where you have to add it explicitly (but with more control, you can put it at the end, at the beginning, both, or basically anywhere in the comparison string).

project_task_code/tests/test_project_task_code.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 # Copyright 2016 Tecnativa <vicent.cubells@tecnativa.com>
+# Copyright 2020 Therp BV <http://therp.nl>
```
NL66278: Change these to https:

NL66278 (changes requested):
:+1: LGTM but few changes would make it perfect.

NL66278 (approved):
:+1: LGTM

PR #417, opened 2018-08-13
[ADD][11.0]project_wbs_task

AaronHForgeFlow:
This module integrates tasks in the wbs structure to make easier the management of task when using wbs. It adds those features:
- The possibility to search for task by the WBS complete reference or name.

- A button in the project tree view that will conduct the user to the list view for the associated tasks.
Depends on
- [x] project_wbs https://github.com/OCA/project/pull/416 Do not merge until oca_dependency.txt explicit branch to that PR is removed

PR #655, opened 2020-03-10
[ADD][11.0]project_wbs_task

AaronHForgeFlow:
Supersedes https://github.com/OCA/project/pull/417


project_wbs_task/README.rst
```diff
@@ -0,0 +1,59 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+
```
MiquelRForgeFlow: add `================================`

AaronHForgeFlow:
The problem seems to be in the rec_name of the account_analytic_account,  it was set to the value 'display_name' in the module `account_analytic_parent` and now even if I overwrite it that seems not taking any effect. Bonus track: when same issue appeared in the module `project_wbs`, it was fixed in the same way, but now, it is not taking any effect. I am not saying its aliens...

PR #660, opened 2020-03-26
[FIX] Convert datetime into date with timezone

ddufresne:
Before this fix, as a user, if you entered a timesheet with a given datetime,
the timesheet could actually be stored as another date in the database.

This falsifies the grid view and other reports that group analytic lines per date.

### DIFF
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -25,8 +25,13 @@ class AccountAnalyticLine(models.Model):
     @api.model
     def _eval_date(self, vals):
         if vals.get('date_time'):
-            return dict(vals, date=fields.Date.to_date(vals['date_time']))
+            return dict(vals, date=self._convert_datetime_to_date(vals['date_time']))
         return vals
+
+    def _convert_datetime_to_date(self, datetime_):
+        if isinstance(datetime_, str):
+            datetime_ = fields.Datetime.from_string(datetime_)
+        return fields.Date.context_today(self, datetime_)
 
     @api.model
     def _running_domain(self):
--- a/project_timesheet_time_control/tests/test_project_timesheet_time_control.py
+++ b/project_timesheet_time_control/tests/test_project_timesheet_time_control.py
@@ -2,7 +2,7 @@
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0
 
 from odoo.tests import common
-from odoo import exceptions, fields
+from odoo import exceptions
 from datetime import timedelta, date, datetime
 
 
--- a/project_timesheet_time_control/tests/test_project_timesheet_time_control.py
+++ b/project_timesheet_time_control/tests/test_project_timesheet_time_control.py
@@ -60,15 +60,30 @@ class TestProjectTimesheetTimeControl(common.TransactionCase):
         self.line.employee_id = self.other_employee
         self.assertFalse(self.line.show_time_control)
 
-    def test_create_write_analytic_line(self):
-        line = self.env['account.analytic.line'].create({
-            'date_time': fields.Datetime.now(),
+    def test_create_analytic_line(self):
+        line = self._create_analytic_line(datetime(2016, 3, 24, 3), tz="EST")
+        self.assertEqual(line.date, date(2016, 3, 23))
+
+    def test_create_analytic_line_with_string_datetime(self):
+        line = self._create_analytic_line('2016-03-24 03:00:00', tz="EST")
+        self.assertEqual(line.date, date(2016, 3, 23))
+
+    def test_write_analytic_line(self):
+        line = self._create_analytic_line(datetime.now())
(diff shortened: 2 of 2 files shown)


pedrobaeza:
I don't really understand the problem...

@Yajo do you?

foutoucour (approved):

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -25,9 +25,14 @@ class AccountAnalyticLine(models.Model):
...
         if vals.get('date_time'):
-            return dict(vals, date=fields.Date.to_date(vals['date_time']))
+            return dict(vals, date=self._convert_datetime_to_date(vals['date_time']))
         return vals
 
+    def _convert_datetime_to_date(self, datetime_):
+        if isinstance(datetime_, str):
+            datetime_ = fields.Datetime.from_string(datetime_)
+        return fields.Date.context_today(self, datetime_)
+
```
yajo, line 35: This should be enough, no need to add an extra method:

```suggestion
            return dict(
                vals,
                date=fields.Date.to_date(fields.Datetime.to_datetime(vals['date_time']))
            )
        return vals
```

project_timesheet_time_control/tests/test_project_timesheet_time_control.py
```diff
@@ -60,15 +60,30 @@ def test_aal_from_other_employee_no_button(self):
...
+        self.assertEqual(line.date, date(2016, 3, 23))
+
+    def test_create_analytic_line_with_string_datetime(self):
+        line = self._create_analytic_line('2016-03-24 03:00:00', tz="EST")
+        self.assertEqual(line.date, date(2016, 3, 23))
+
+    def test_write_analytic_line(self):
+        line = self._create_analytic_line(datetime.now())
+        line.with_context(tz="EST").date_time = '2016-03-24 03:00:00'
+        self.assertEqual(line.date, date(2016, 3, 23))
```
yajo, line 74: @pedrobaeza Notice this test.

If an AAL is created with this timezone and this time, the AAL date field should be mirrored as being the previous day, because DB dates are always stored in UTC, and technically, in UTC it would be day 23.

yajo (changes requested):

project_timesheet_time_control/models/account_analytic_line.py
ddufresne: @Yajo I disagree about your suggestion. First, it breaks the behavior. Second, the extra method makes the intension more explicit.

yajo: What does it break?

ddufresne: @Yajo it does not use the timezone of the user. Your solution is similar to the original behavior before this pull request.

yajo: OK, how about this?

```suggestion
            ts = fields.Datetime.to_datetime(vals['date_time'])
            ts = fields.Date.context_today(ts)
            return dict(vals, date=ts)
```

ddufresne: @Yajo ok, but there's a small error in your proposal:

ts = fields.Date.context_today(ts)

should be

ts = fields.Date.context_today(self, ts)

dreispt:
@ddufresne We should use the builtin methods of the Date / Datetime fields, so @Yajo is on the right track here.

ddufresne:
@dreispt Yajo's solution does not make the unit tests pass.

Otherwise, you are implying that my solution does not use the build-in methods, which is not the case.

dreispt:
@ddufresne Yes you are, and your code is passing the tests checking the timezone.
So this seems correct to me.

dreispt (approved):

ddufresne:
@dreispt thanks!

yajo (approved):
Yes, I didn't mean your code was wrong. I just think that adding an extra method to fix this is not worth it. But nevermind, do as you please. :man_shrugging:

pedrobaeza:
Oh, wait, please squash the 2 commits.

pedrobaeza (changes requested):
Squash commits

marcelsavegnago (approved):

pedrobaeza:
Please squash commits for merging this.

pedrobaeza:
Why have you merged this?

ddufresne:
@pedrobaeza what do you mean?

You told to squash before merging. I squashed before merging.

The PR is approved by 4 people other than me.

pedrobaeza:
You must squash in your branch, but not merge the PR. That's done with the bot, and now there's no version bumping due to this. Please don't do this next time and just squash in your local branch.

ddufresne:
@pedrobaeza my appologies, did not know about this.

Couldn't this robot squash the PR by itself before merging?

pedrobaeza:
No, it's a limitation

yajo:
FTR https://github.com/OCA/oca-github-bot/issues/43

PR #661, opened 2020-04-04
[13.0][MIG] project_custom_info

bodedra:


bodedra:
Will open again when depend module is ready in V13.0

Dependant module is https://github.com/OCA/server-tools/tree/12.0/base_custom_info

PR #667, opened 2020-04-21
[FIX][13.0] project_category can now be migrated from 11.0 to 13.0

mlaitinen:
In commit 26a7e26481bc35306ed9aacfb0d2f26bbf78a37d the view `project_type_search` of `project.project` was renamed to `project_search` and there was a new view introduced with the same XML ID, but for a different model, `project.type`. This means that the `inherit_id` field of that view still contains the old value of `project.view_project_project_filter` and obviously that view shouldn't inherit any other view. Thus the `inherit_id` is explicitly set to `False`.

PR for version 12.0 here: #666 :smiling_imp:

### DIFF
--- a/project_category/views/project_type_views.xml
+++ b/project_category/views/project_type_views.xml
@@ -45,6 +45,7 @@
     </record>
     <record id="project_type_search" model="ir.ui.view">
         <field name="model">project.type</field>
+        <field name="inherit_id" eval="False"/>
         <field name="arch" type="xml">
             <search string="Type">
                 <field name="complete_name" />
--- a/project_category/views/project_type_views.xml
+++ b/project_category/views/project_type_views.xml
@@ -45,7 +45,7 @@
     </record>
     <record id="project_type_search" model="ir.ui.view">
         <field name="model">project.type</field>
-        <field name="inherit_id" eval="False"/>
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza:
Please check travis

pedrobaeza:
Is it 13.0 version really needed?

mlaitinen:
The same migration code is necessary in the 13.0 branch if you're migrating Odoo from 11.0 directly to 13.0 and skipping 12.0.

To me it looks like all the PR's targeting `13.0` aren't passing the Travis check for the same reason:
```
/home/travis/.travis/job_stages: line 878: syntax error near unexpected token `fi'
/home/travis/.travis/job_stages: line 878: `fi'
```

emagdalenaC2i (approved):

project_category/views/project_type_views.xml
```diff
@@ -45,6 +45,7 @@
     </record>
     <record id="project_type_search" model="ir.ui.view">
         <field name="model">project.type</field>
+        <field name="inherit_id" eval="False"/>
```
dreispt:
```suggestion
        <field name="inherit_id" eval="False" />
```

PR #684, opened 2020-06-18
[MIG] project_template: Migration to 13.0

ghost:


okuryan (approved):

HaraldPanten (changes requested):
Hi @alfadil 

It seems that it exists an incompatibility with project_key module. I describe the behaviour:

1- Open the options of a project and set "is template" checkbox to True.
2- Once the project has been defined as a template, click to "Create project from template". It doesn't matter if you do that by the form view or the Kanban view. It will work properly.
3- Try to create a second project using the same template as before: An error occurs as a restriction popup will tell the user that "Project key must be unique".

If you create one project from a template it is working fine. If you try to create a second project using the same template, user is going to be restricted by the unique key.

Could you check?

Thanks!

ghost:
hi @HaraldPanten 
the incompatibility with project_key module is inherited problem from previous versions, And frankly I think the user should change the name of the project generated from project template rather than trying to change the behavior of one of the modules 
Thanks for your comment

HaraldPanten (approved):
It's OK then.

Functional review 👍

patrickrwilson (approved):
LGTM

PR #689, opened 2020-06-29
[11.0][FIX] project_task_material_with_sale_timesheet: The value of super is not returned

igallart:
The value of super is not returned.


HaraldPanten (approved):
Small fix LGTM 👍

dreispt (approved):

PR #690, opened 2020-07-07
[12.0] [ADD] project_category: Added some fields in the Project type model.

Chandresh-SerpentCS:
**Project Type Model Enhancement**

### DIFF
--- a/project_category/models/project_type.py
+++ b/project_category/models/project_type.py
@@ -38,6 +38,18 @@ class ProjectType(models.Model):
     task_ok = fields.Boolean(
         string='Can be applied for tasks'
     )
+    is_billable = fields.Boolean(
+        string="Is Billable"
+    )
+    is_internal = fields.Boolean(
+        string="Is Internal"
+    )
+    is_budgeted = fields.Boolean(
+        string="Is Budgeted"
+    )
+    code = fields.Char(
+        string="Code"
+    )
 
     @api.constrains('parent_id')
     def check_parent_id(self):
--- a/project_category/views/project_type_views.xml
+++ b/project_category/views/project_type_views.xml
@@ -1,14 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
-
     <record id="project_type_form" model="ir.ui.view">
         <field name="model">project.type</field>
         <field name="arch" type="xml">
             <form string="Type">
                 <sheet>
                     <div class="oe_title">
-                        <label for="name" class="oe_edit_only"/>
-                        <h1><field name="name"/></h1>
+                        <label class="oe_edit_only" for="name"/>
+                        <h1>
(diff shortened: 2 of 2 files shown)


Chandresh-SerpentCS:
@dreispt Request for review the PR.

HaraldPanten (approved):

alexey-pelykh (commented):
I don’t really understand value of the module except for adding checkboxes

alexey-pelykh:
Or how name of the module corresponds to content

project_category/models/project_type.py
```diff
@@ -38,6 +38,18 @@ class ProjectType(models.Model):
     task_ok = fields.Boolean(
         string='Can be applied for tasks'
     )
+    is_billable = fields.Boolean(
```
alexey-pelykh, line 41: Can’t be this deduced from sale order link?

alexey-pelykh (commented):
Cc @pedrobaeza calling fo extra opinion

dreispt:
@alexey-pelykh 
I prepared these requirements for Chandresh. I realize now I should have created an RFQ Issue to clarify the goals.
The requirement behind this is to enable Project time analysis.

- Billable projects are "productive time", they mean income, and the other mean expenses.
- Budgeted projects are internal projects that have a budget assigned to. So they are also "productive time", but they do not generated income.

Maybe the "Is internal" flag is redundant.

Thoughts?

alexey-pelykh:
Hm, thanks for the information. I got the idea, yet I don't understand how this implementation would work (on it's own) with other OCA modules. It feels like a valid business case, yet unlikely applicable (in current implementation) as-is to other deployments. Billable projects (well, 100% expenses projects or with some profit) is doable via analytic accounts with ease, for much stronger analysis capabilities. Budgeted projects are investment projects, in other words. Likewise, analytic accounts [with parenting] could be a way of differentiating cost centers. In current implementation I'm leaning towards not supporting this module: too broad name, too detached implementation, low value (IMHO).

dreispt:
Adding these field would lay the ground for this to be used in other places. Alone adds low value, but is as enabler for features adding actual values, such as dashboards or List view filters/groups.

IIUC your suggestion is to use Parent Analytic Accounts? This greatly limits your flexibility to freely setup your Analytic tree. Also, I'm having trouble finding this field on the default UI.

Now, I noticed we have a Analytic Account "Group" field, that could be used for this:
image

alexey-pelykh:
> Adding these field would lay the ground for this to be used in other places. Alone adds low value, but is as enabler for features adding actual values, such as dashboards or List view filters/groups.

Absolutely, if this module would've added these dashboards etc. - then it would've made sense.

Does Analytic Tags/Groups solve your case?

dreispt:
We're solving this is a different way - just using the Project Types, thanks.

PR #692, opened 2020-07-15
Time scheduling group is shown for all

cesarlr:
date_start and date are not limited by groups, so the their group should not be limited neither

### DIFF
--- a/project_timeline/views/project_project_view.xml
+++ b/project_timeline/views/project_project_view.xml
@@ -29,6 +29,9 @@
                 <field name="date_start"/>
                 <field name="date"/>
             </xpath>
+            <xpath expr="//field[@name='resource_calendar_id']/.." position="attributes">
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But `resource_calendar_id` is not used in `project_timeline`. Why are you removing the groups here?

cesarlr:
The [container group](https://github.com/odoo/odoo/blob/12.0/addons/project/views/project_views.xml#L153) was limited to system users. I remove the group there to allow all users to edit `date_start` and `date`. If I don't do it there is no way to show them because of the permissions of the container group.  `resource_calendar_id` is still restricted to system users

pedrobaeza (approved):
OK, thanks for the clarifications. I didn't see the `/..` part.

marcelsavegnago (approved):

PR #695, opened 2020-08-05
[FIX] project_timesheet_time_control: take users timezone into account when setting date field

davejames:
Port changes in https://github.com/OCA/project/pull/660 to 13.0


pedrobaeza (approved):

HaraldPanten (approved):
Seems to be OK

marcelsavegnago (approved):

PR #699, opened 2020-08-17
[12.0][ADD] project_leverage

### LINKED ISSUE #697: [RFC] project_leverage
# Description

This module allows you to report on the leverage of a project manager. It supports changes of Project Manager.

# Usage

* Go to Project > Reporting > Timesheets
* Select the measure Duration
* Group by Project Manager

# Technical Specifications

## account.analytic.line

- [ ] Add a many2one field "Project Manager" (project_user_id) to res.users. The field is read-only and defaults to the project manager of the related project.
- [ ] Add a pivot view in Project > Reporting > Timesheets

## project overview

- [ ] Add a "Leverage" section
- [ ] Add the previous pivot view with the duration (measure), grouped by project manager (row) and date/month (column) and filtered by the project.

Khalid-SerpentCS:
This module allows you to report on the leverage of a project manager. It supports changes of Project Manager.

Closes #697


bodedra (approved):

HaraldPanten (approved):

project_leverage/models/account_analytic_line.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+
+    project_user_id = fields.Many2one('res.users',
+                                      related='task_id.project_id.user_id',
+                                      string='Project Manager',
+                                      store=True,
+                                      readonly=True)
```
dreispt, line 13: README says this supports Project manager changes, but I'm not sure.
When the PM is changed, the related Analytic Lines will all be recomputed and updated.
I think we need a non computed field, and logic on create/write to set this value.

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+        else:
+            query += """
+                project_id = %s AND employee_id = %s AND is_manager IS %s
+            """ % (str(project_ids[0]), str(employee_ids.id), 'TRUE')
+        return query
+
+    def line_group(self):
+        return """
+            group by Year, month, date_mm, employee_id order by month
+        """
```
dreispt, line 94: Missing code?

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+        _group = self.line_group()
+        query = _select + _from + _where + _group
+        request.env.cr.execute(query)
+        raw_data = request.env.cr.dictfetchall()
+        project_data = {}
+        month_list = []
+        month_name_list = []
+        for rec in raw_data:
+            if str(rec.get('month')).strip() + ' ' + str(
+                    int(rec.get('year'))) not in month_name_list:
```
dreispt: If you add a "month_year" SQL column, you won't need to repeat these string operation all these times.

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+class SaleTimesheetController(SaleTimesheetController):
+
+    def _plan_prepare_values(self, projects):
+
+        employee_obj = request.env['hr.employee']
+        values = super(SaleTimesheetController,
+                       self)._plan_prepare_values(projects)
+        _select = self.select_line()
+        _from = self.line_from()
+        _where = self.line_where(projects.ids)
```
dreispt: You're passing an IDs list, to later browse it into a Rset again.
```suggestion
        _where = self.line_where(projects)
```

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def line_from(self):
+        return """
+            FROM account_analytic_line
+        """
+
+    def line_where(self, project_ids):
+        project_mananger_ids = [
+            project.user_id.id
+            for project in request.env['project.project'].browse(project_ids)]
```
dreispt: Work with recordsets as much as you can, and use ID lists only when needed, and as late as possible.
```suggestion
    def line_where(self, projects):
        project_manangers = projects.mapped("user_id")```

File: project_leverage/controllers/main.py [hunk 6]
```diff
@@ -0,0 +1,109 @@
...
+        employee_ids = request.env['hr.employee'].search(
+            [('user_id', 'in', project_mananger_ids),
+             ('user_id', '!=', False)])
+        query = """
+            WHERE
+        """
+        if not employee_ids:
+            raise UserError('Project Manager is not defined or no '
+                            'timesheet entry found for project manager.')
+        if len(project_ids) > 1 and len(employee_ids) > 1:
```
dreispt: To check for nonzero length, this is the Pythonic way
```suggestion
        if project_ids and employee_ids:
```

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+            query += """
+                project_id IN %s AND employee_id IN %s AND is_manager IS %s
+            """ % (str(tuple(project_ids)),
+                   str(tuple(employee_ids.ids)), 'TRUE')
+        elif len(project_ids) > 1 and len(employee_ids) == 1:
```
dreispt: No need for all these conditions, they're doing the same thing.
You can use IN with a one element list, or even an empty list...

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,109 @@
...
+             ('user_id', '!=', False)])
+        query = """
+            WHERE
+        """
+        if not employee_ids:
+            raise UserError('Project Manager is not defined or no '
+                            'timesheet entry found for project manager.')
+        if len(project_ids) > 1 and len(employee_ids) > 1:
+            query += """
+                project_id IN %s AND employee_id IN %s AND is_manager IS %s
```
dreispt: I don't get the is_manager filter.
This will exclude the data from the project team. 
I think that works against the goal of the module.

dreispt (changes requested):
Form code review, I believe this module won't work as expected.
Detailed comments and suggestions in the review.
Thanks!

project_leverage/views/account_analytic_view.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+        <record id="view_task_project_user_pivot" model="ir.ui.view">
+            <field name="name">account.analytic.line.pivot</field>
+            <field name="model">account.analytic.line</field>
+            <field name="arch" type="xml">
+                <pivot string="Timesheet Analysis">
+                    <field name="employee_id" type="row"/>
```
dreispt, line 8: The group is by project manager
```suggestion
                    <field name="project_manager_id" type="row"/>
```

project_leverage/models/account_analytic_line.py
```diff
@@ -0,0 +1,32 @@
...
+    def _compute_is_manager(self):
+        for rec in self:
+            if rec.employee_id.user_id == rec.project_user_id:
+                rec.is_manager = True
+
+    @api.model
+    def create(self, vals):
+        project_user_id = vals.get('task_id', False).project_id.user_id
+        vals.update({'project_user_id': project_user_id})
+        return super(AccountAnalyticLine, self).create(vals)
```
dreispt, line 30: Can we have this as a default value on the field instead?
That would have the benefit of populating the new field with initial values.

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,95 @@
...
+        month_name_list = []
+        for rec in raw_data:
+            month_name_list.append(rec.get('month_year'))
+            if str(rec.get('date_mm')).strip() + str(
+                    int(rec.get('year'))) not in month_list:
+                month_list.append(
+                    str(rec.get('date_mm')).strip() + str(
+                        int(rec.get('year'))))
+            if project_data.get(rec.get('employee_id') or 0):
+                project_data[rec.get('employee_id') or 0].update(
```
dreispt, line 37: @Khalid-SerpentCS 
This details the hours per employee, and not per project manager.
Should this be by project manager intead? cc @max3903

project_leverage/models/account_analytic_line.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @api.multi
+    def write(self, vals):
+        if vals.get('task_id', False):
+            project_user_id = vals.get('task_id', False).project_id.user_id
+            vals.update({'project_user_id': project_user_id})
```
dreispt: There is a bug here,as `vals.get('task_id') is not a browse record
```suggestion
        if "task_id" in vals:
            Task = self.env["project.task"]
            task_id = vals.get('task_id')
            task = Task.browse(task_id) if task_id else Task  
            project_user_id = task.project_id.user_id
            vals.update({'project_user_id': project_user_id and project_user_id.id else None})
```

dreispt (changes requested):
A couple of comments to address, thanks.
I have some concerns about the actual computations, but will need more tests.

Khalid-SerpentCS:
Hi @AmmarOfficewalaSerpentCS, Can you please review the comments.

project_leverage/README.rst
```diff
@@ -0,0 +1,28 @@
...
+Project Leverage
+================
+
+This module allows you to report on the leverage of a project manager. It supports changes of Project Manager.
+
+
+Installation
+============
+
+* No specific installation step required.
```
dreispt, line 15:
```suggestion
* No specific installation step is required.
```

dreispt:
@Khalid-SerpentCS Build is red, please fix it.

PR #700, opened 2020-08-19
[13.0][FIX] project_timesheet_time_control: fix compatibility with hr_timesheet_activity_begin_end module

davejames:
Fixes a compatibility issue with [this module](https://github.com/OCA/timesheet/tree/13.0/hr_timesheet_activity_begin_end)

Prior to this commit, trying to resume an activity which had a start time or end time set would cause the newly created timesheet to throw a validation error because time_stop - time_start must equal duration. This commit fixes that issues by adding those fields to the ignore list.

### DIFF
--- a/project_timesheet_time_control/wizards/hr_timesheet_switch.py
+++ b/project_timesheet_time_control/wizards/hr_timesheet_switch.py
@@ -114,6 +114,11 @@ class HrTimesheetSwitch(models.TransientModel):
                 # resume an invoiced AAL if that module is installed,
                 # and it doesn't hurt here
                 "timesheet_invoice_id",
+                # These fields are from the hr_timesheet_activity_begin_end
+                # module. Unless ignored, these fields will cause a validation
+                # error because time_stop - time_start must equal duration.
+                "time_start",
+                "time_stop",
             }
             inherited.read(_fields)
             values = inherited._convert_to_write(inherited._cache)


davejames:
@pedrobaeza @alexey-pelykh would it be possible to get a review on this?

pedrobaeza (approved):

alexey-pelykh:
I don't use modules in question, so out of context what is being fixed

davejames:
@alexey-pelykh 

The project_timesheet_time_control module adds a datetime field for recording the start time to the account_analytic_line model, and adds a button to set the task duration by calculating the time between the task start and the current time. So it acts like a task timer.

When clicked a second time, this button "resumes" the task by creating a duplicate account_analytic_line record with the start time set to the current time, using the same vals as the original record.

The hr_timesheet_activity_begin_end module adds a start time and end time field to the account_analytic_line model and has a constraint to ensure that the task duration matches the start and end time.

So what is happening is when the task is duplicated and the start_time and end_time fields are set, but the duration field is not set, then we get a validation error. So the solution in this commit is to remove those keys from the prepared vals (if present) before they are used to create the new timesheet line.

pedrobaeza:
@Yajo can review this

alexey-pelykh:
@davejames yeah, regarding `project_timesheet_time_control` there was a huge debate in https://github.com/OCA/timesheet/pull/139 :) Your explanation helps

alexey-pelykh (approved):

davejames:
oh wow, yes that is a big debate!

I have some other compatibility issues between these modules which I was planning on addressing in the near future. Ultimately it would be nice to have the ability to have a timer to start and stop tasks as this module does, but also keep track of the start and stop time in the hr_timesheet_activity_begin_end module. So my plan is to make these modules work a bit more seamlessly together. Currently when you close the timer you then have to update the start and stop times (and then be careful not to resume the task, until this PR is merged at least).

pedrobaeza:
For me, this is redundant, as you already have the start time, which is the datetime in `project_timesheet_time_control` module, and the stop is start + duration. If you wan to have both type of fields, I suggest you to expand `project_timesheet_time_control` for:

- Expanding/Changing label for datetime field to indicate explicitly that is the start time.
- Adding a computed/inverse field stop time that is computed as start + duration, and inverse changing duration according the introduced stop date.

pedrobaeza:
Anyway, let's merge this for now:

/ocabot merge patch

PR #702, opened 2020-08-21
[13.0][IMP] project_timesheet_time_control: add a field for the end time of a timesheet

davejames:
This commit adds a field to the timesheet to store the end time. It is not required to enter the end time, if it is not entered it is calculated automatically by comparing the start time and the duration, if only a start time and end time are entered, then the duration is calculated accordingly. The end time field is set to optional in the tree view.

### DIFF
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -5,6 +5,8 @@
 
 from datetime import datetime
 
+from dateutil.relativedelta import relativedelta
+
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
 
@@ -13,12 +15,42 @@ class AccountAnalyticLine(models.Model):
     _inherit = "account.analytic.line"
     _order = "date_time desc"
 
-    date_time = fields.Datetime(default=fields.Datetime.now, copy=False)
+    date_time = fields.Datetime(
+        string="Start Time", default=fields.Datetime.now, copy=False
+    )
+    date_time_end = fields.Datetime(
+        string="End Time",
+        compute="_compute_date_time_end",
+        inverse="_inverse_date_time_end",
+    )
     show_time_control = fields.Selection(
         selection=[("resume", "Resume"), ("stop", "Stop")],
         compute="_compute_show_time_control",
         help="Indicate which time control button to show, if any.",
     )
+
+    @api.depends("date_time", "unit_amount", "product_uom_id")
+    def _compute_date_time_end(self):
+        hour_uom = self.env.ref("uom.product_uom_hour")
+        for record in self:
+            if (
+                record.product_uom_id == hour_uom
+                and record.date_time
+                and record.unit_amount
+            ):
+                record.date_time_end = record.date_time + relativedelta(
+                    hours=record.unit_amount
+                )
+            else:
+                record.date_time_end = record.date_time_end
+
+    def _inverse_date_time_end(self):
+        hour_uom = self.env.ref("uom.product_uom_hour")
+        for record in self.filtered(lambda x: x.date_time and x.date_time_end):
+            if record.product_uom_id == hour_uom:
+                record.unit_amount = (
+                    record.date_time_end - record.date_time
+                ).seconds / 3600
 
     @api.model
     def _eval_date(self, vals):
--- a/project_timesheet_time_control/views/account_analytic_line_view.xml
+++ b/project_timesheet_time_control/views/account_analytic_line_view.xml
@@ -8,7 +8,8 @@
                 <attribute name="invisible">1</attribute>
             </field>
             <field name="date" position="after">
-                <field name="date_time" string="Date" required="1" />
+                <field name="date_time" required="1" />
+                <field name="date_time_end" optional="hide" />
             </field>
             <field name="unit_amount" position="after">
                 <field name="show_time_control" invisible="1" />
@@ -65,7 +66,8 @@
                 <attribute name="invisible">1</attribute>
             </field>
             <field name="date" position="after">
-                <field name="date_time" string="Date" required="1" />
+                <field name="date_time" required="1" />
+                <field name="date_time_end" />
             </field>
         </field>
     </record>
@@ -94,7 +96,8 @@
                 />
             </xpath>
             <field name="date" position="after">
-                <field name="date_time" required="1" />
+                <field name="date_time" />
+                <field name="date_time_end" />
             </field>
             <xpath
                 expr="//templates//t[@t-esc='record.date.value']"
@@ -104,6 +107,11 @@
             </xpath>
             <xpath expr="//templates//t[@t-esc='record.date.value']" position="after">
                 <t t-esc="record.date_time.value" />
+                <t t-if="record.date_time_end.value">
+                    <br />
+                    <span>to</span>
+                    <t t-esc="record.date_time_end.value" />
+                </t>
             </xpath>
         </field>
     </record>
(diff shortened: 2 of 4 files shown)


project_timesheet_time_control/views/account_analytic_line_view.xml
```diff
@@ -94,7 +96,8 @@
                 />
             </xpath>
             <field name="date" position="after">
-                <field name="date_time" required="1" />
+                <field name="date_time" />
```
astirpe, line 99: Why removing the `required="1"` here?

davejames: this is a kanban view, so required is redundant

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -13,13 +15,33 @@ class AccountAnalyticLine(models.Model):
...
         selection=[("resume", "Resume"), ("stop", "Stop")],
         compute="_compute_show_time_control",
         help="Indicate which time control button to show, if any.",
     )
 
+    @api.depends("date_time", "unit_amount")
+    def _compute_date_time_end(self):
+        for record in self:
+            record.date_time_end = record.date_time + relativedelta(
+                hours=record.unit_amount
```
astirpe: I have the feeling that the UOM should be included in the compute of `unit_amount`, to avoid errors case of UOMs other than time.

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -13,13 +15,33 @@ class AccountAnalyticLine(models.Model):
...
+            )
+
+    def _inverse_date_time_end(self):
+        for record in self:
+            record.unit_amount = (
```
astirpe: Here you are assuming that `unit_amount` is always in UOM expressed in hours, that could be not always the case.

astirpe (changes requested):

pedrobaeza (approved):
We can deprecate right now the other module with this.

project_timesheet_time_control/models/account_analytic_line.py
davejames: yeah you're right, i hadn't thought about non-timesheet related usage of this model

pedrobaeza: There's no sense in filling these fields for non-timesheet analytic lines. What has to be taken into account (it's also a mistake in previous module state), is if the timesheet unit is not hours, but days (or other compatible UoM).

davejames: I just pushed a commit to check if the uom is hourly, i think if the uom is daily then the end time isn't necessary anyway

pedrobaeza: I think it is, as it should put the fraction of the working hour day. Please check how core does when you select days as timesheet unit.

davejames: I don't see a lot of value in calculating the days. For example, you may put a quantity of 5 days, which would be converted into 40 hours by default, so automatically setting the end date would need to take the working schedule into account, but in reality it's guess work, that 5 days might not be a contiguous period of time.

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -13,13 +15,39 @@ class AccountAnalyticLine(models.Model):
...
+    @api.depends("date_time", "unit_amount")
+    def _compute_date_time_end(self):
+        hour_uom = self.env.ref("uom.product_uom_hour")
+        for record in self:
+            if record.product_uom_id == hour_uom:
+                record.date_time_end = record.date_time + relativedelta(
+                    hours=record.unit_amount
+                )
+            else:
+                record.date_time_end = False
```
astirpe: I think it should be like this:

```suggestion
                record.date_time_end = record.date_time_end
```

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -13,13 +15,39 @@ class AccountAnalyticLine(models.Model):
...
         selection=[("resume", "Resume"), ("stop", "Stop")],
         compute="_compute_show_time_control",
         help="Indicate which time control button to show, if any.",
     )
 
+    @api.depends("date_time", "unit_amount")
+    def _compute_date_time_end(self):
+        hour_uom = self.env.ref("uom.product_uom_hour")
+        for record in self:
+            if record.product_uom_id == hour_uom:
```
astirpe: You should also check that both `date_time` and `unit_amount` have a value to calculate `date_time_end`

project_timesheet_time_control/models/account_analytic_line.py
davejames: included in my latest commit

project_timesheet_time_control/models/account_analytic_line.py
davejames: included in my latest commit

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -13,13 +15,43 @@ class AccountAnalyticLine(models.Model):
...
+        compute="_compute_date_time_end",
+        inverse="_inverse_date_time_end",
+    )
     show_time_control = fields.Selection(
         selection=[("resume", "Resume"), ("stop", "Stop")],
         compute="_compute_show_time_control",
         help="Indicate which time control button to show, if any.",
     )
 
+    @api.depends("date_time", "unit_amount")
```
astirpe: UOM missing in the depends

project_timesheet_time_control/views/account_analytic_line_view.xml
```diff
@@ -104,6 +107,9 @@
             </xpath>
             <xpath expr="//templates//t[@t-esc='record.date.value']" position="after">
                 <t t-esc="record.date_time.value" />
+                <br />
```
astirpe: Not sure if this `<br />` is intentional. I think it shouldn't be here.

project_timesheet_time_control/models/account_analytic_line.py
davejames: uom is already a dependency of hr_timesheet

astirpe: I mean the dependency of the compute method

project_timesheet_time_control/views/account_analytic_line_view.xml
astirpe: Shouldn't this block:

```
               <span>to</span>
                <t t-esc="record.date_time_end.value" />
```

be displayed only if `date_time_end` has a value?

davejames: it's intentional, it's so that the times are both in line with each other. This is easier on the eyes in my opinion than having them on the same line.

image

as opposed to 

image

project_timesheet_time_control/models/account_analytic_line.py
davejames: oooh, yes you're right

project_timesheet_time_control/views/account_analytic_line_view.xml
astirpe: It's ok. But I think the `to` should not be displayed is `date_time_end` is false

davejames: I agree, I added that to the latest commit

project_timesheet_time_control/models/account_analytic_line.py
davejames: included in latest commit

astirpe (approved):
LGTM

davejames:
thanks @astirpe for the thorough review :)

davejames:
squashed

pedrobaeza:
Thanks, merging!

/ocabot merge minor

Can you make a PR for changing the README of `hr_timesheet_activity_begin_end` for deprecating it saying that it's not needed using this one?

PR #644, opened 2020-02-12
12.0 project parent

gfcapalbo:
project parent v12

PR #703, opened 2020-08-26
[13.0][MIG] project_parent module

tfossoul:


pedrobaeza:
Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0.

tfossoul:
@pedrobaeza  I squash everything into this commit. https://github.com/OCA/project/pull/703/commits/43794a3b5cdcb27ad5c5e0a6602d1eed6f63ed73 because there are 12 commits only for the v12 pull request (https://github.com/OCA/project/pull/644). It is better for readability to get only one.

pedrobaeza:
That's not valid, as readability here is a secondary point, as you haven't respected commit attribution to the original authors of the code, so please preserve such main commit history. You can squash administrative commits with their immediately main commit though.

tfossoul:
The PR in v12 is still not merged. The author didn't keep v10 module history.
So should I keep history from v10 or from the v12 PR ?
When the v12 PR will be merged we will break history.
So the only way to keep history is to ask the v12 PR author to do it, merge its work and then work on this PR v13.

I suppose i'm not the first one to get this kind of problem :/

pedrobaeza:
You can take one of these approaches:

- Take v10 history, and continue from there.
- Take v12 branch, and continue from there.

depending on which code base you are starting from. Obviously, the ideal thing is to get v12 version merge first, but if not possible, you can do this.

tfossoul:
Here is it.

project_parent/__manifest__.py
```diff
@@ -5,10 +5,9 @@
     "version": "13.0.1.0.0",
     "license": "LGPL-3",
     "category": "project",
-    "author": "Therp B.V., Elico Corp, Odoo Community Association (OCA)",
+    "author": "haulogy SA/NV, Therp B.V., Elico Corp, Odoo Community Association (OCA)",
```
dreispt: I'm not sure there is enough originality to award Authorship.
I would rather have you added to the contributors section, and not here.

dreispt (approved):

project_parent/__manifest__.py
tfossoul: Sure.

tfossoul:
@pedrobaeza @dreispt does we need more approver before been able to merge ?

HaraldPanten (approved):

PR #708, opened 2020-09-10
[12.0] FIX project_timesheet_time_control when not using employees

eLBati:
Otherwise, when not using employees:

stop

### DIFF
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -38,7 +38,7 @@ class AccountAnalyticLine(models.Model):
         """Domain to find running timesheet lines."""
         return [
             ("date_time", "!=", False),
-            ("employee_id", "in", self.env.user.employee_ids.ids),
+            ("user_id", "=", self.env.user.id),
             ("project_id.allow_timesheets", "=", True),
             ("unit_amount", "=", 0),
         ]


pedrobaeza:
AFAIK, the employee is a mandatory data for proper timesheeting (without this module), so I'm not sure this is relevant.

mrcast (approved):
Functional test ok!

eLBati:
The current behavior is wrong anyway: why are the users able to start timers but they are not able to stop them?

If employee is mandatory, then the "start work" button should not appear when employee is not set.

Nevertheless, I would not force to have an employee to use timers: there are cases where the users just need to track the quantity of worked hours, without needing to track employees data like hour cost

pedrobaeza:
As said, it's not me, but Odoo requiring it. Don't you have any problem doing that without this module?

mrcast:
@pedrobaeza excuse me but I don't understand "Odoo requiring it." may you explain me please?
I could be a user not related to an employee. if employee is mandatory, then the "start work" button should not appear when employee is not set or an alert "please select an employee" would be a better user experience when the start button is clicked
tks!

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -38,9 +38,11 @@ def _running_domain(self):
         """Domain to find running timesheet lines."""
         return [
             ("date_time", "!=", False),
-            ("employee_id", "in", self.env.user.employee_ids.ids),
             ("project_id.allow_timesheets", "=", True),
             ("unit_amount", "=", 0),
+            '|',
+            ("employee_id", "in", self.env.user.employee_ids.ids),
+            ("user_id", "=", self.env.user.id),
```
astirpe: Shouldn't searching by `user_id` be enough?
Searching by `employee_id` here seems not necessary to me, since `("employee_id", "in", self.env.user.employee_ids.ids)` is already covered by `("user_id", "=", self.env.user.id)` IMO.

eLBati: Right, changed, thanks

astirpe (approved):
LGTM

eLBati:
> Don't you have any problem doing that without this module?

@pedrobaeza I am not aware of any problem

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -38,7 +38,7 @@ def _running_domain(self):
         """Domain to find running timesheet lines."""
         return [
             ("date_time", "!=", False),
-            ("employee_id", "in", self.env.user.employee_ids.ids),
+            ("user_id", "=", self.env.user.id),
```
pedrobaeza, line 41: I remembered an error message when trying to add a timesheet with a user with no employee, but it seems that message has gone away.

Anyway, the duality is still there, so please put both conditions:

```suggestion
            '|',
            ("employee_id", "in", self.env.user.employee_ids.ids),
            ("user_id", "=", self.env.user.id),
```

pedrobaeza (changes requested):

project_timesheet_time_control/models/account_analytic_line.py
astirpe: In the timesheet lines, `user_id` and `employee_id` are always consistent.
See https://github.com/odoo/odoo/blob/13.0/addons/hr_timesheet/models/hr_timesheet.py#L134-L137 defined in `_timesheet_preprocess()` and invoked in create and write.
Also see https://github.com/odoo/odoo/blob/13.0/addons/hr_timesheet/models/hr_timesheet.py#L72-L78 in the create.

Having only `("user_id", "=", self.env.user.id)` is just fine IMO.

pedrobaeza: No if you don't have any user linked to your employee. Both searches are needed.

pedrobaeza: Thinking twice, although that can happen, not for "interactive" timesheeting, so OK to the switch.

pedrobaeza (approved):

pedrobaeza:
Please forward-port it to v13

eLBati:
There it is https://github.com/OCA/project/pull/709

Thanks

PR #709, opened 2020-09-11
[13.0] FIX project_timesheet_time_control when not using employees

eLBati:
From https://github.com/OCA/project/pull/708


pedrobaeza (approved):

pedrobaeza:
We can fast-track it:

/ocabot merge patch

PR #711, opened 2020-09-29
[10.0] project_issue_code: included parent filter_domain

fredzamoabg:
included parent filter_domain to avoid changing filter behavior

### DIFF
--- a/project_issue_code/views/project_issue_view.xml
+++ b/project_issue_code/views/project_issue_view.xml
@@ -40,7 +40,13 @@
         <field name="inherit_id" ref="project_issue.view_project_issue_filter" />
         <field name="arch" type="xml">
             <xpath expr="//field[@name='name']" position="attributes">
-                <attribute name="filter_domain">['|', ('name', 'ilike', self), ('issue_code', 'ilike', self)]</attribute>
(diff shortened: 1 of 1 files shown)


project_issue_code/views/project_issue_view.xml
```diff
@@ -40,7 +40,14 @@
         <field name="inherit_id" ref="project_issue.view_project_issue_filter" />
         <field name="arch" type="xml">
             <xpath expr="//field[@name='name']" position="attributes">
-                <attribute name="filter_domain">['|', ('name', 'ilike', self), ('issue_code', 'ilike', self)]</attribute>
+                <attribute name="filter_domain">['|', '|', '|', '|', '|', \
+                    ('partner_id','child_of',self), \
+                    ('description','ilike',self), \
+                    ('email_from','ilike',self), \
+                    ('name','ilike',self), \
+                    ('name', 'ilike', self), \
```
SimoRubi, line 47: This kind of notation is always a bit confusing, but it seems to me that all the conditions are put together with `OR`, so why is it needed to repeat this one twice?

SimoRubi (changes requested):
Thanks for you PR!
Please see comment below

SimoRubi (approved):
code review, ok for me.

PR #712, opened 2020-09-29
[REF] project_task_code: have Task Code extension friendly

dreispt:
This change does not affect any of the existing features, but it makes easier to extend the logic assigning the Task Code.


bodedra (approved):
LG :+1:

dreispt:
Tests are failing, and working on it, I found a better way to address my need.
Thanks for the review.

PR #731, opened 2020-10-20
[IMP] project_task_code: support already initialized code column

dreispt:
Support the case where the database table already has a `code` column.
Only populate new task codes on records without  a task number set.
Code Blackified.


skeller1 (approved):
Functional Review :+1:

bodedra (approved):

PR #746, opened 2020-11-09
[13.0][FIX] project_hr: forbid employee tag creation on project task form

skeller1:
**Impacted versions:**
13.0

**Steps to reproduce:**

1. Open OCA project runbot instance
2. Add employee category (tag) to a project
3. Create new task for that project
4. Open Employee Categories Drop Down and create new employee category
5. Save the task

Constraint warning appears

project_hr_with_create
project_hr_not_allowed_at_project_level

**Solution:**

Allowed employee categories must be assigned on project level.
We can solve this by forbidden the creation of an employee category in project task form.

project_hr_no_create

### DIFF
--- a/project_hr/views/project_task_views.xml
+++ b/project_hr/views/project_task_views.xml
@@ -13,7 +13,11 @@
             <field name="user_id" position="before">
                 <field name="allowed_hr_category_ids" invisible="1" />
                 <field name="allowed_user_ids" invisible="1" />
-                <field name="hr_category_ids" widget="many2many_tags" />
+                <field
+                    name="hr_category_ids"
+                    widget="many2many_tags"
+                    options="{'no_create': True}"
+                />
             </field>
             <field name="user_id" position="attributes">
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Or maybe pass the context to create the category with proper tags?

skeller1:
> Or maybe pass the context to create the category with proper tags?

What do you mean with "with proper tags"?

AFAIU: I should add a context information and in the extended create method of hr.employee.category model I start to:

1. Add the new category to the project
2. Add the new category to the employee

Is this correct?

Thank you

pedrobaeza:
Well, if I don't remember bad, the constraint happens because the constraint is not complied with, so the creation of the record should comply with that constraint, and that's why I say about auto-filling proper data for it, but not sure which data should be right now.

skeller1:
After checking the code this simple UI Improvement is the best solution to avoid errors generated by rookie users or not key users.

Would be perfect to merge this or give more input (to solve it in a better way).
Thank you

pedrobaeza (approved):

PR #756, opened 2020-11-15
[13.0][IMP] project_key: Allow copying of projects

### LINKED ISSUE #734: [13.0][project_key] unable to change key for project already created from a template (and having tasks)
current module doesn't play nice with project_template.

use case: 
1. create a project and set it as a template. Set key = "BRX"
2. using a service product configured with "Service Tracking" as "Create a task in sales" and select the template in point 1.
3. On SO (having the product of point 2) confirmation, the project is created with not nice key while would expect template provided key with either increment (BRX1, BRX2) or other addition to the provided key to render it unique.
4. unable to change the project key once there are existing tasks. getting error: "Task key must be unique!"

This is solved by adding the key to the task's form view and manually modifying the task's keys before having the project new key accepted.

skeller1:
1. Allow copying of projects (Add own sequence record `P1-PXXXXX`)
2. Fix updating task keys after changing project key containing `-` (replaced split_part by substring)
3. Fix updating sequence name after changing project name with existing project key

4. Fixing post_init_hook to set all project task keys (folded task_ids are ignored through domain see [here](https://github.com/odoo/odoo/blob/9241b92945edf695a534642b8ffe1977f9b32c8f/addons/project/models/project.py#L186))

**post_init_hook result before**

project_key_hook_before_fix

**post_init_hook result after fixing**

project_key_hook_after_fix

 Fixes #734

### DIFF
--- a/project_key/__manifest__.py
+++ b/project_key/__manifest__.py
@@ -10,6 +10,6 @@
     "author": "Modoolar, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/project/",
     "depends": ["project"],
-    "data": ["views/project_key_views.xml"],
+    "data": ["data/ir_sequence_data.xml", "views/project_key_views.xml"],
     "post_init_hook": "post_init_hook",
 }
--- a/project_key/data/ir_sequence_data.xml
+++ b/project_key/data/ir_sequence_data.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <data noupdate="1">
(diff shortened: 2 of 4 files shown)


ygol (commented):
I've runbot tested the following that works fine:

1. Duplicate an existing project:
    - a new key is proposed and there is no error.
2. Create a template project:
    - duplicate a project and configure it as a template.
    - create a service product with service tracking = "Create a task in sales order's project"
    - add the service product to a new SO.
    - confirm the SO and a project is created without error. 
   This solves will solve https://github.com/OCA/project/issues/808

PR #765, opened 2020-11-20
[14.0] [MIG] project stage closed

Yoshiro009:


project_stage_closed/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# Copyright 2017 Tecnativa - Pedro M. Baeza
+# Copyright 2017 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Project Task Stage Closed",
+    "summary": "Make the Closed flag on Task Stages available without "
+    "installing sale_service"
+    "",
+    "author": "ACSONE SA/NV," "Tecnativa," "Odoo Community Association (OCA)",
```
skeller1, line 9: Could you please fix the double quote issue in author string line. This is a result of black,isort run during migration from 12.0 to 13.0.
Thank you

skeller1 (changes requested):
LGTM: Functional and technical review on runbot and local machine.

Small change in manifestfile.
LGTM :+1: 

Thank you for supporting OCA

HaraldPanten (approved):

skeller1 (approved):

petrus-v:
I think this module is useless in version 14.0 as long `project` module has an `is_closed` field with business logic implement.

What do you think about ? what's the best approach here, at least I'll like to create a new PR with deprecated message in readme also to prevent migration to V15 ?

pedrobaeza:
Indeed it's totally useless. I would create in this version a new PR with a version 2.0.0, with migration scripts that copy data from `closed` to `is_closed` when set, and update README saying that this module is no longer needed since v14, and remove the rest of the logic.

petrus-v:
@pedrobaeza thanks for your prompt reply I'll prepare a PR as you suggested !

PR #766, opened 2020-11-22
[FIX] project_task_code: If code column exists, the update process stop

flachica:
init_hook should test if column code exist yet

### DIFF
--- a/project_task_code/hooks.py
+++ b/project_task_code/hooks.py
@@ -10,8 +10,10 @@ def pre_init_hook(cr):
     code constraint when the module is installed and before the post-init-hook
     is launched.
     """
-    cr.execute("ALTER TABLE project_task " "ADD COLUMN code character varying;")
-    cr.execute("UPDATE project_task " "SET code = id;")
+    cr.execute(
+        "ALTER TABLE project_task " "ADD COLUMN IF NOT EXISTS code character varying;"
+    )
+    cr.execute("UPDATE project_task " "SET code = id where code is null;")
 
 
 def post_init_hook(cr, registry):


alexey-pelykh (approved):

project_task_code/hooks.py
```diff
@@ -10,8 +10,10 @@ def pre_init_hook(cr):
     code constraint when the module is installed and before the post-init-hook
     is launched.
     """
-    cr.execute("ALTER TABLE project_task " "ADD COLUMN code character varying;")
-    cr.execute("UPDATE project_task " "SET code = id;")
+    cr.execute(
+        "ALTER TABLE project_task " "ADD COLUMN IF NOT EXISTS code character varying;"
```
alexey-pelykh, line 14: I'd keep all keywords CAPS to match styling

flachica: Sorry, I can't quite understand you. You mean this statement should look like this: ALTER TABLE project_task ADD COLUMN IF NOT EXISTS code CHARACTER VARYING;

skeller1 (commented):
+1 LGTM Code and functional review (both on runbot and locally)

skeller1 (approved):

HaraldPanten (approved):

dreispt:
This was implemented for 12.0 with https://github.com/OCA/project/pull/731
IMO it makes more sense to forward port that PR.

skeller1:
The code is nearly the same, but in 13 and 14 there are maybe different precommit style guides and such stuff.
Also this addon is still not migrated for V14, so after merging this PR this change is available for the migration.

But if you can forward port the V12 merge this would also be great.
I would prefer merging this PR. In the end the update process for this addon should be fixed ...
Thank you

skeller1:
sorry, PR #731 is also fixing selecting of task (only with empty code field), so port forward would be great. Thank you

PR #771, opened 2020-12-16
Remove unnecessary translated term that is error prone

ypapouin:
No translatable content


alexey-pelykh:
What about RTL languages?

ypapouin:
Oh right, didn't think about that.

PR #783, opened 2021-01-27
[13.0] [MIG] project_custom_info: MIgration to v13.0

CarlosRoca13:
cc @Tecnativa TT26771

please @Tardo @carlosdauden review this


Tardo (approved):

pedrobaeza:
Can you please take the occasion to add demo data to this module? You can take as inspiration demo data in base_custom_info.

CarlosRoca13:
@pedrobaeza demo data added

lfreeke (approved):
LGTM :+1: (Functional test)

project_custom_info/demo/custom.info.template.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model,model_id:id
+tpl_addressed,Addressed to...,res.partner,base.model_res_partner
```
pedrobaeza: This is not correct, as the model should be project. And the idea is to put some data related to project, not PEGI.

pedrobaeza (changes requested):

CarlosRoca13:
I changed the demo data, can you review it again @pedrobaeza

project_custom_info/demo/default.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2021 Tecnativa - Carlos Roca
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl). -->
+<odoo>
+    <record id="prop_time" model="custom.info.property">
+        <field name="default_value">3</field>
```
pedrobaeza: Why not including this on the CSV?

project_custom_info/demo/project_project.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2021 Tecnativa - Carlos Roca
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl). -->
+<odoo>
+    <record id="project_with_custom_info_demo" model="project.project">
+        <field name="name">Create a module for allow custom info on models</field>
```
pedrobaeza: After `for`, a gerund is always needed (`for allowing`), but let's shorten it:
```suggestion
        <field name="name">Module analysis</field>
```

project_custom_info/demo/default.xml
CarlosRoca13: I'm based on base_custom_info. In these module, the default value was added like here... I don't know the reason but I thought that's important.

pedrobaeza: Please try the other way to see. I don't find right now the reason for this.

CarlosRoca13:
Changes done @pedrobaeza the default value on csv is working fine

pedrobaeza (approved):

PR #790, opened 2021-02-12
[IMP] Allow negative task material quantity

wpichler:
Removed quantity check
Added logic to switch src and location on move
Added logic to use return picking type if configured
Split positive and negative moves into two pickings

### DIFF
--- a/project_task_material/models/project.py
+++ b/project_task_material/models/project.py
@@ -5,7 +5,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl-3.0).
 
 from odoo import api, fields, models, _
-from odoo.exceptions import ValidationError
 
 
 class Task(models.Model):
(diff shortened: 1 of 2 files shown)


project_task_material/models/project.py
```diff
@@ -30,12 +29,3 @@ class ProjectTaskMaterial(models.Model):
     product_id = fields.Many2one(
         comodel_name='product.product', string='Product', required=True)
     quantity = fields.Float(string='Quantity')
-
-    @api.multi
-    @api.constrains('quantity')
-    def _check_quantity(self):
-        for material in self:
-            if not material.quantity > 0.0:
```
petrus-v, line 38: Thanks for contribution,

I'm not sure what's the business logic to get negative task material quantity, any way you should probably add opt-in to let administrator allows or not negative quantity and forbid by defaut to avoid to break existing behaviour.

Could you fix test according if you want to get a chance to have your work merged ?

petrus-v (changes requested):

PR #797, opened 2021-03-06
 [11.0][MIG] backport project_template from 12.0 to 11.0

yweng8111:


marylla (approved):
I made a functional review and it looks good to me. :+1: 

Here is what I did:

* Create a new project and activate the flag for "Is Template"
-> It is not shown in the kanban list by default (because of the filter for non-templates)
* Add a description, set a project manager, define some stages and add tasks.
* Create a project from the template
-> The new project is a copy of the template, also the name gets a "(Copy)", everything is just as it is in the template (well, without the activated is template flag of course), even the tasks ands its information are copied.

chandni299 (approved):
Code Review :+1:

yweng8111:
@dreispt Can you please check this?

yweng8111:
@OCA/project-service-maintainers Can you please check this?

PR #799, opened 2021-03-09
[REF] Remove wrong 'web' submodule directory

dreispt:


dreispt:
Administrative fix, fast tracking

PR #801, opened 2021-03-15
[13.][IMP] project_timeline: add Start Date field, to replace Assigned Date (major change)

### LINKED ISSUE #774: [13.0] field "date_start" doesn't exist anymore in project_timeline
Field "date_start" dosn't exist anymore. This field appears to have been replaced by "date_assing" but this one is not editable. 
When a user want schedule a task to start on a specific date , they cannot.  The "date_assign" field is automatically set with the date corresponding to the moment in which the task is assigned to a user

dreispt:
Closes #774

### DIFF
--- a/project_timeline/hooks.py
+++ b/project_timeline/hooks.py
@@ -1,0 +1,16 @@
+# Copyright 2021 Open Source Integrators - Daniel Reis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def populate_date_start(cr, registry):
+    """
+    The date_start was introduced to be used instead of date_assign.
+    To keep same behaviour on upgrade, initialize it
+    to have the same data as before.
+    """
+    cr.execute(
+        "UPDATE project_task "
+        "SET date_start = date_assign "
+        "WHERE date_start IS NULL "
+        "AND date_assign IS NOT NULL"
+    )
--- a/project_timeline/models/project_task.py
+++ b/project_timeline/models/project_task.py
@@ -1,12 +1,15 @@
 # Copyright 2016-2017 Tecnativa - Pedro M. Baeza
 # Copyright 2017 Tecnativa - Carlos Dauden
+# Copyright 2021 Open Source Integrators - Daniel Reis
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import models
+from odoo import fields, models
 
 
 class ProjectTask(models.Model):
     _inherit = "project.task"
+
+    date_start = fields.Datetime("Start Date")
 
     def update_date_end(self, stage_id):
         res = super().update_date_end(stage_id)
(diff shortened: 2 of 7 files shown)


project_timeline/hooks.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 Open Source Integrators - Daniel Reis
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def populate_date_start(cr, registry):
```
Tardo, line 5: This should be moved to "migrations"

project_timeline/models/project_task.py
```diff
@@ -1,13 +1,16 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from odoo import models
+from odoo import fields, models
 
 
 class ProjectTask(models.Model):
     _inherit = "project.task"
 
+    date_start = fields.Datetime("Start Date")
```
Tardo, line 12: To ensure avoid problems with possible function signature changes.
```suggestion
    date_start = fields.Datetime(string="Start Date")
```

Tardo (changes requested):
Thanks! I think that what is being done in the hook should be moved to the migrations folder. Please, check travis errors.

anajuaristi (approved):

pedrobaeza:
@dreispt can you attend comments and make Travis green?

project_timeline/hooks.py
dreispt: What would the cons of having it as it is?

project_timeline/models/project_task.py
dreispt: I'm sorry, I don't see the point in the change here. Using positional arguments is documented and supported, and will not change in this Odoo version.

oihane (approved):
LGTM

project_timeline/hooks.py
oihane: What happens if you already have it installed, there is no migration script that will execute this.

project_timeline/hooks.py
```diff
@@ -0,0 +1,16 @@
...
+    """
+    The date_start was introduced to be used instead of date_assign.
+    To keep same behaviour on upgrade, initialize it
+    to have the same data as before.
+    """
+    cr.execute(
+        "UPDATE project_task "
+        "SET date_start = date_assign "
+        "WHERE date_start IS NULL "
+        "AND date_assign IS NOT NULL"
```
pedrobaeza, line 15:
```suggestion
        "UPDATE project_task "
        "SET date_start = COALESCE(date_assign, create_date) "
        "WHERE date_start IS NULL "
```

project_timeline/models/project_task.py
pedrobaeza: Not really needed, but it's always welcome to use keyword arguments as such.

project_timeline/hooks.py
pedrobaeza: Exactly. It should be as hook, but also call it from a migration script:

```python
from odoo.addons.project_timeline.hooks import populate_date_start
from odoo.modules import registry


def migrate(cr, version):
    populate_date_start(cr, registry)
```

pedrobaeza (changes requested):

pedrobaeza:
Already done in #860

PR #811, opened 2021-05-17
[13.0][IMP] project_timesheet_time_control: Get only base fields in wizard from account.analytic.line to prevent errors when custom fields exists

victoralmau:
Get only base fields in wizard from `account.analytic.line` to prevent errors when custom fields exists.

Please @pedrobaeza and @carlosdauden can you review it?

@Tecnativa TT29728


pedrobaeza:
Incorrect patch.

PR #815, opened 2021-05-27
[ADD] 14.0 new module project_analytic_code

petrus-v:
This module add related Analytic account code in project
name (`name_get`) and you can search projects by the analytic account
code (`name_search`) from any list (like in timesheets).


petrus-v:
I suppose this module is incompatible with project_key, is there a way to tell travis to test them separately ?

petrus-v:
> I suppose this module is incompatible with project_key, is there a way to tell travis to test them separately ?

Last commit should do the trick waiting from travis following `copier update` [instructions](https://odoo-community.org/groups/contributors-15/contributors-201141?mode=date&date_begin=2021-07-01&date_end=2021-08-01)

petrus-v:
> > I suppose this module is incompatible with project_key, is there a way to tell travis to test them separately ?
> 
> Last commit should do the trick waiting from travis following `copier update` [instructions](https://odoo-community.org/groups/contributors-15/contributors-201141?mode=date&date_begin=2021-07-01&date_end=2021-08-01)

in fact it was an issue while setting up the unit-test assuming an analytic account were existing, revert that change

project_analytic_code/models/project.py
```diff
@@ -0,0 +1,32 @@
...
+        related="analytic_account_id.code",
+        store=True,
+        index=True,
+    )
+
+    @api.model
+    def name_search(self, name="", args=None, operator="ilike", limit=100):
+        args = list(args or [])
+        if name:
+            args += ["|", ("analytic_account_code", "=", name)]
```
grindtildeath, line 21:
```suggestion
            args = OR([args, [("analytic_account_code", "=", name)]]
```
This is not the best way to modify a domain, because depending on the provided one, it could lead to inconsistent results.
You should better use `OR` from `osv.expression` to safely "extend" the domain if such is your intent.

project_analytic_code/models/project.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2021 Pierre Verkest
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
```
grindtildeath, line 5:
```suggestion
from odoo.osv.expression import OR

```

project_analytic_code/views/project.xml
```diff
@@ -0,0 +1,46 @@
...
+    <data>
+        <record id="view_project_project_filter" model="ir.ui.view">
+            <field name="inherit_id" ref="project.view_project_project_filter" />
+            <field name="name">project.project.select.cgi_project</field>
+            <field name="model">project.project</field>
+            <field name="arch" type="xml">
+                <field name="name" position="attributes">
+                    <attribute
+                        name="filter_domain"
+                    >["|", ("name", "ilike", self), ("analytic_account_code", "=", self)]</attribute>
```
grindtildeath, line 12: Can't we just add another field to be searched instead of hijacking the `name` one?

project_analytic_code/views/project.xml
```diff
@@ -0,0 +1,46 @@
...
+        <record model="ir.ui.view" id="view_project_kanban">
+            <field name="inherit_id" ref="project.view_project_kanban" />
+            <field name="name">project.project.kanban.project_analytic_code</field>
+            <field name="model">project.project</field>
+            <field name="priority">200</field>
+            <field name="arch" type="xml">
+                <field name="name" position="after">
+                    <field name="analytic_account_code" />
+                </field>
+                <xpath expr="//div[hasclass('o_primary')]//span" position="replace">
```
grindtildeath: Replacing an element can also break inheritance for other module (cf https://dev.to/opencode/don-t-break-your-and-others-odoo-xml-373i ) 

Can't you just change add the missing elements you need instead?

project_analytic_code/models/project.py
```diff
@@ -0,0 +1,32 @@
...
+        return super().name_search(name=name, args=args, operator=operator, limit=limit)
+
+    def name_get(self):
+        result = []
+        for project in self:
+            name = ""
+            if project.analytic_account_code:
+                name += f"[{project.analytic_account_code}] "
+            name += project.name
+            result.append((project.id, name))
```
grindtildeath:
```suggestion
        result = super().name_get()
        # Prepend analytic_account_code to display_name
        for i, (res_item, project) in enumerate(zip(result, self)):
            code = project.analytic_account_code
            if code:
                result[i] = (res_item[0], "[%s] %s" % (code, res_item[1]))
        return result
```
I guess we can still do the same with the call to `super` to avoid breaking inheritance. :wink:

project_analytic_code/views/project.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <data>
+        <record id="view_project_project_filter" model="ir.ui.view">
+            <field name="inherit_id" ref="project.view_project_project_filter" />
+            <field name="name">project.project.select.cgi_project</field>
```
petrus-v: wrong naming !
```suggestion
            <field name="name">project_analytic_code.project.select</field>
```

project_analytic_code/models/project.py
petrus-v: hum... the matter is that I really want to add the OR operator with the domain add by super `("name", operator, name)`, I don't want to use OR with provide domain by args!... I'll revert this change !

project_analytic_code/views/project.xml
petrus-v: I think user must be able to search on what it see. here the kanban view is the default and display analytic account code as part of the name. It seems legit to me to add that search (I can see that it conflict with the project_key module which does the same things)

I could move this magic in `search` method but it's probably too much intrusive.

If you really prefer I can move it in a dedicated filter?

project_analytic_code/views/project.xml
petrus-v: done !

project_analytic_code/views/project.xml
petrus-v: changed !

project_analytic_code/models/project.py
grindtildeath: I'm not sure to understand why you don't want to use the `OR` operator, can you please explain?

The issue using `+` to extend the domain is that the ensuing domain will not necessarily be syntactically correct as the `|` expects two operands.

Please look at the following example done using odoo shell on a demo DB:
```
>>> domain = ['|', ('description', '!=', False), ('name', 'ilike', 'dummy')]
>>> self.env["project.project"].name_search(args=domain)
[]
>>> append_domain = domain + ['|', ('analytic_account_id.code', '=', '12')]
>>> append_domain
['|', ('description', '!=', False), ('name', 'ilike', 'dummy'), '|', ('analytic_account_id.code', '=', '12')]
>>> self.env["project.project"].name_search(args=append_domain)
  File "<console>", line 1, in <module>
  File "/odoo/src/odoo/models.py", line 1902, in name_search
    ids = self._name_search(name, args, operator, limit=limit)
  File "/odoo/src/odoo/models.py", line 1918, in _name_search
    return self._search(args, limit=limit, access_rights_uid=name_get_uid)
  File "/odoo/src/odoo/models.py", line 4693, in _search
    if expression.is_false(self, args):
  File "/odoo/src/odoo/osv/expression.py", line 218, in is_false
    for token in reversed(normalize_domain(domain)):
  File "/odoo/src/odoo/osv/expression.py", line 210, in normalize_domain
    assert expected == 0, 'This domain is syntactically not correct: %s' % (domain)
AssertionError: This domain is syntactically not correct: ['|', ('description', '!=', False), ('name', 'ilike', 'dummy'), '|', ('analytic_account_id.code', '=', '12')]
```

OTOH, if you use the `OR` operator, you'll make sure of its correctness:
```
>>> from odoo.osv.expression import OR
>>> or_domain = OR([domain, [('analytic_account_id.code', '=', '12')]])
>>> or_domain
['|', '|', ('description', '!=', False), ('name', 'ilike', 'dummy'), ('analytic_account_id.code', '=', '12')
>>> self.env["project.project"].name_search(args=or_domain)
[]
```

project_analytic_code/views/project.xml
grindtildeath: To me, the analytic account code is part of the displayed name and not the name itself, so by doing this the user is not able to search in the name field only.
Anyway if other modules are doing it already, I guess you can keep it this way.

project_analytic_code/models/project.py
petrus-v: I agree that using osv.expression is safer but I'm not sure how I can do it in this case.

Domain I expect here is `["|", ("analytic_account_code", "=", name), ("name", "ilike", name)]` where the ` ("name", "ilike", name)` is provided by odoo https://github.com/OCA/OCB/blob/14.0/odoo/models.py#L1815

With your suggestion tests failed because assuming `args = []` it would produce the following domain at the end `[("analytic_account_code", "=", name), ("name", "ilike", name)]`.

In your odoo shell demonstration you haven't provide name parameter that why you get the error in this method if name is not provided this domain won't be added.

So I'm not sure how to express the OR with a part of the domain that I don't handle, any idea are welcome.

Thanks for your time

CasVissers-360ERP (approved):
Nice module, very easy in an analytic / project set-up!

project_analytic_code/models/project.py
grindtildeath: Sorry for the delay.

I had a look in the core of Odoo and the good news, is that the possibility to override `name_search` was improved for v16.0 in this PR: https://github.com/odoo/odoo/pull/86588

Looking at it, we can see that there were already multiple overrides with the goal of searching on multiple fields and not only the `_rec_name`. Can we then mimic what is done in Odoo core to avoid the issue? Something around these lines:
```
    @api.model
    def _name_search(self, name, args=None, operator='ilike', limit=100, name_get_uid=None):
        if not (operator == 'ilike' and not (name or '').strip()):
            args = args or []
            domain = ['|', ('name', operator, name), ('analytic_account_code', operator, name)]
            return self._search(expression.OR([domain, args]), limit=limit, access_rights_uid=name_get_uid)
        return super()._name_search(name, args=args, operator=operator, limit=limit, name_get_uid=name_get_uid)
```
If it doesn't work as suggested, can you then please explain the issue in the roadmap segment of the readme so that we can merge this PR?

petrus-v: sorry for the very, very late reply, your suggestion has been implemented from v16.0 and required change in base _name_search that I won't backport here!

As far I've seen practices in Odoo module is to not call super() which I don't find very kind in this case for this module while using _inherit.

I'm using it as it in production for years without any troubles. I'm on the way to migrate this module in version 17.0 so I'll take care of this.

Thanks for your time

PR #819, opened 2021-06-11
[13.0][IMP] project_status: Add multi-company support

joao-p-marques:
@Tecnativa
TT30382

ping @pedrobaeza @Yajo


pedrobaeza:
This misses:

- A migration script that empties current status from any company.
- Empty company_id in current data.
- Not directly related, but I see that the data doesn't have `force_create=False` for allowing to remove them. ¿Or maybe it's better to be demo data?

joao-p-marques:
Ok, changes made @pedrobaeza 

> ¿Or maybe it's better to be demo data?

IMHO it makes more sense as regular data, but without `forcecreate`, as suggested...

pedrobaeza (approved):

leemannd (approved):

pedrobaeza:
Let's not bump the version 2 times:

/ocabot merge nobump

PR #828, opened 2021-06-25
[ADD] project_task_markdown

KKamaa:
Add markdown to description field in project task as an option from default odoo version 13.0 base on [web_widget_text_markdown](https://github.com/OCA/web/pull/1965)


KKamaa:
@thomaspaulb This has been migrated, added the branch repo to oca-requirements.txt as stated in[ doc](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#185oca_dependenciestxt). Travis finds an error in another module not related to this ADD, not sure why.

project_task_markdown/models/project_task.py
```diff
@@ -0,0 +1,11 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo import fields, models
+
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    description = fields.Char(string="Description")
```
thomaspaulb: `fields.Text` seems more logical?

project_task_markdown/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* Tom Blauwendraat <tblauwendraat@therp.nl>
```
thomaspaulb: tom@sunflowerweb.nl

thomaspaulb (changes requested):
Hehe... seems you have just found a timezone bug in `project_timesheet_time_control` :-) Since you like to work around midnight, that's when bugs happen if date functions do not correctly take into account the time zone.

thomaspaulb:
Anyway that test is not ours to fix.. and if you push a fix around daytime I think Travis will be green.

project_task_markdown/readme/CONTRIBUTORS.rst
KKamaa: fixed

project_task_markdown/models/project_task.py
KKamaa: This works too, pushed the fix

thomaspaulb (approved):

gjotten:
First look on a running instance is promising but I'd like to do some more functional tests before final approval :-)

ajaniszewska-dev (approved):

PR #834, opened 2021-07-14
 Allow users without manager access right to track project time

konykon:
Resolves issue https://github.com/OCA/helpdesk/issues/228

### DIFF
--- a/project_timesheet_time_control/wizards/hr_timesheet_switch.py
+++ b/project_timesheet_time_control/wizards/hr_timesheet_switch.py
@@ -124,7 +124,7 @@ class HrTimesheetSwitch(models.TransientModel):
         ).running_timer_id.button_end_work()
         # Start new timer
         _fields = self.env["account.analytic.line"]._fields.keys()
-        self.read(_fields)
+        self.sudo().read(_fields)
         values = self._convert_to_write(self._cache)
         new = self.env["account.analytic.line"].create({
             field: value for (field, value) in values.items()


pedrobaeza:
I don't think this is the proper solution. This has nothing to do with project managers, but access to the involved objects, and the helpdesk module should add such permissions properly.

project_timesheet_time_control/wizards/hr_timesheet_switch.py
```diff
@@ -124,7 +124,7 @@ def action_switch(self):
         ).running_timer_id.button_end_work()
         # Start new timer
         _fields = self.env["account.analytic.line"]._fields.keys()
```
daniquilez, line 126: @pedrobaeza @konykon What about if we change this line by:

`fields = self.env["account.analytic.line"].fields_get().keys()`

This way we only get fields that user has permission to access.

pedrobaeza: But that way we will lose the information on that fields. I still don't get how to reproduce the problem.

daniquilez: Problem is: Module helpdesk_mgmt_timesheet adds 2 fields to account.analytic.line only accessible by group helpdesk_mgmt.group_helpdesk_user. https://github.com/OCA/helpdesk/blob/12.0/helpdesk_mgmt_timesheet/models/hr_timesheet.py

So when we activate this helpdesk module, a non helpdesk user will get permission problems when using project_timesheet_time_control. The call to ```self.env["account.analytic.line"]._fields.keys()```  will return fields you don't have permission to read, create or anything. 

A user not having access to helpdesk but to project should be able to get fields on account.analytic.line he has permission to read. Thats why ```fields_get``` is a better option than ```_fields```. ```fields_get``` already gives you fields you have permission to access.

Issue #836, opened 2021-07-16
Project creation fails


StephaneMangin:
Due to usage of the global project.task.type tree view that contains project_ids relation and generates bad values retrieved by the project create method.

Dict values was : 
```
    "type_ids": [
        [6, False, [53, 54, 55, 56, 57, 59, 60]],
        [
            1,
            53,
            {
                "sequence": 10,
                "name": "Analysis",
                "case_default": True,
                "fold": False,
                "mail_template_id": False,
                "rating_template_id": False,
                "auto_validation_kanban_state": False,
                "description": False,
                "project_ids": [
                    [6, False, [4718]],
                    [0, "virtual_745", {"display_name": False, "color": False}],
                ],
            },
        ],
```

PR #837, opened 2021-07-16
[14.0][FIX] Add custom embedded tree view in project.project for project.task.type

StephaneMangin:
issue is that OCA module adds field type_ids inside project form without defining the view, so it select odoo default tree view for project.task.type which include project_ids in it:

image (1)

By defining a custom embedded view, the problem is solved

Fixes : https://github.com/OCA/project/issues/836


grindtildeath (approved):
:+1:

simahawk (approved):
LG run pre-commit pls

PR #841, opened 2021-07-19
14.0 mig project budget

hurrinico:
It depends on https://github.com/OCA/account-budgeting/pull/50


afbenevides:
Hello, since https://github.com/OCA/account-budgeting/pull/50 is now merged, could that be re-opened?

PR #843, opened 2021-08-08
13.0 mig project wbs

juanpgarza:
Standard migration


bosd:
Runbot is red, 
Can you fix this one?

AaronHForgeFlow (changes requested):
The root analytic account seems not being calculated correctly. See this example of a project with a  child and two grandchildren:



The grandchildren project has its parent as root analytic account. It should be its grandparent instead:

PR #772, opened 2020-12-23
[14.0][MIG]project_timesheet_time_control

watthanun:

PR #846, opened 2021-08-18
[14.0][MIG] project_timesheet_time_control: Migration to 14.0

AntoniRomera:
Depends on OCA/timesheet#447


AntoniRomera:
Hello @pedrobaeza ,

I name you because I don't see project maintainers team, sorry for the inconvenience, in version 14 odoo has added a post_init_hook that creates two rows in "account.analytic.line" in the module "hr_timesheet" then when passing the test of this module indicated me that there was more than one work in progress, the solution that I have reached is to remove those two rows. How do you see it or if someone proposes another solution?

AntoniRomera:
@newtratip 

Hi, i see you migrated the module "project_key" i have a problem with the mock version, you know how to solve this?

Runbot log:

> [31mERROR: Could not find a version that satisfies the requirement mock==3.0.5 (from odoo14-addon-project-key) (from versions: none)[0m
> [31mERROR: No matching distribution found for mock==3.0.5[0m

pedrobaeza:
There's another PR for migrating the module in #772 (I haven't checked its status). Please also put a proper PR title for next times for easing the search.

AntoniRomera:
I close this PR in favour of the other then

pedrobaeza:
Well, maybe that one is abandoned and not correctly migrated, but we should give a bit more priority to it first.

AntoniRomera:
We can look at migrating this PR since we have no response from the other one, I am reopening it.
Looking at the travis log, I noticed that for the "project_key" module we have version 3.0.5 of the "mock" dependency which is only supported up to python version 3.5.x, currently using version 3.6.x.

pedrobaeza:
Why pinning a specific version of the library? Better to let it free.

AntoniRomera:
I don't really understand, in order to make the modification we have to do a PR of the other module to update the dependency?

pedrobaeza:
I see this CI correct, so where is the problem you are pointing?

AntoniRomera:
Sorry, the error is in the runbot, not in the CI.

pedrobaeza:
But I think that's another thing, as I don't see such error. I think we can continue for now.

AntoniRomera:
Oh wait, here's an image of the runbot log.
Error

pedrobaeza:
OK, then add an extra commit releasing such pinning in the module manifest and requirements.txt

pedrobaeza:
I'm thinking if such requirement is for tests, and then, it shouldn't be put on the manifest, and add `test_requirements.txt` file.

AntoniRomera:
Have I done it correctly?

pedrobaeza:
Oh, sorry, the files should be `test-requirements.txt`.

pedrobaeza:
Uhm, still erroneous... @simahawk do you know what is going on?

simahawk:
https://app.travis-ci.com/github/OCA/project/jobs/532833784#L1253

Still project_key controller test. What I don't get is why those tests are patching the request if they are using HttpCase. That's wrong IMO.

pedrobaeza:
Apart from what you comment, Simone, previously, there was a pinned version in mock library. @AntoniRomera maybe the problem is that in `test-requirements.txt`, there's no such pin. Fixing `project_key` tests would be the ideal, but it's more complicated, so let's see if with this, we can continue.

simahawk:
that's wrong too. `mock` is built-in lib so you should trash https://github.com/OCA/project/blob/14.0/project_key/__manifest__.py#L15 an

pedrobaeza:
That's what he is trying to do in the commit https://github.com/OCA/project/pull/846/commits/9604348c2388e73f60e7a52142c4e0cd091567ca with this result.

simahawk:
yes, but
1. there's no need for the requirement (it's built-in)
2. there's no need for mock -> stop mocking :)

simahawk:
last but not least, the way the request is mocked there is wrong... anyway, just get rid of it.

pedrobaeza:
Yeah, 1 is covered here, and it fails, and 2 is too much to be asked to a contributor of another module. Maybe do you have the magic recipe to do 2 in a blink?

simahawk:
no, 1 is NOT covered here: do not add `mock` to `test-requirements.txt` nor anywhere else (even if you still want to use mock).
2: remove the mocking and stop worrying about it. That test should not use it at all.
Anyway, here's a blind fix https://github.com/OCA/project/pull/847 let's see.

pedrobaeza:
Please rebase

AntoniRomera:
@pedrobaeza Can you view the logs of ci/runbot?

AntoniRomera:
@pedrobaeza 

Is there a way to avoid the installation of the "project_key" module so that the runbot passes correctly, until PR #848  proceeds?

FernandoRomera (approved):

pedrobaeza:
Please avoid merge commits in the PR, and you can squash administrative commits with the previous regular one.

MRomeera (approved):

AntoniRomera:
@pedrobaeza @hkapatel-initos 

You can do a review so we can do the merge because I think the runbot is in red and there is no one to do a review and the runbot is in red because of the "project_key" module.

pedrobaeza:
There's a contamination of other commits/files in this PR. Please clean up it.

pedrobaeza:
Please don't add merge commits in a PR.

AntoniRomera:
@pedrobaeza 

Sorry, can you tell me what is wrong, I don't see exactly what I have to clean up, I have done the commits merge.

pedrobaeza:
You have to rebase, not to merge/pull. Now the final commit is including again a lot of undesired changes.

AntoniRomera:
@pedrobaeza 

Did I get it right this time? Sorry for the inconvenience

project_timesheet_time_control/hooks.py
```diff
@@ -11,3 +13,26 @@ def post_init_hook(cr, registry):
...
+        analytic_line_ids = env["account.analytic.line"].search(
+            [
+                ("name", "=", _("Analysis")),
+                ("task_id", "in", project_id.task_ids.ids),
+                ("user_id", "=", admin.id),
+            ]
+        )
+
+        if analytic_line_ids:
+            analytic_line_ids.unlink()
```
pedrobaeza: This is not correct, as it's destructing data.

project_timesheet_time_control/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+access_hr_timesheet_switch_portal,access_hr_timesheet_switch portal,model_hr_timesheet_switch,base.group_user,1,1,1,1
```
pedrobaeza: This should be put to the timesheet user

pedrobaeza (changes requested):

hkapatel-initos:
Can you please resolve this  WARNING?
https://app.travis-ci.com/github/OCA/project/jobs/534054346#L1512

AntoniRomera:
> Can you please resolve this WARNING?
> https://app.travis-ci.com/github/OCA/project/jobs/534054346#L1512

Solved

project_timesheet_time_control/hooks.py
AntoniRomera: Sorry, i fix this by stop any running timer in testss

AntoniRomera:
@pedrobaeza Do you think the modification I have made is correct? Instead of deleting, I have stopped the analytical lines in the test so that it does not give errors.

pedrobaeza (approved):

PR #848, opened 2021-08-23
[14.0][IMP] project_key

AntoniRomera:
- Improve key generation, error on install with another modules

### DIFF
--- a/project_key/models/project_project.py
+++ b/project_key/models/project_project.py
@@ -3,6 +3,7 @@
 
 from odoo import _, api, fields, models
 from odoo.osv import expression
+from odoo.tools import config
 
 
 class Project(models.Model):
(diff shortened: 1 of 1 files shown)


AntoniRomera:
@pedrobaeza (Is there a group of maintainers of this repository?)

Hello,
I think the module "project_key" has important shortcomings and only gives me no error if the module is installed alone, for the test has a variable that adds to the context and I think it is not the way to do it, I have made these modifications to see if the community like it, as the PR #846  gives error because of this module.

pedrobaeza:
This module has been always conflicting, and I don't like too much nor use it, but it seems to have its user base, being migrated version per version, but the context + config part was put to isolate the behavior. This is not the solution then, and Travis is still red. The problem in runbot seems to be on demo data, isn't it?

AntoniRomera:
Travis was red because I have not made the modification in all the places where it generates the "key" field. 

I think that by generating a unique "key" field, i.e. by putting a number after the sequence if the project name generates a key equal to an existing one.

And remove the varaible code from the context or check if odoo has active tees.

AntoniRomera:
> but the context + config part was put to isolate the behavior.

The problem is that the PR #846  fails because of this module, the fix I think is the right one, because the problem comes at the time of making the tests, if so and we keep as it is now, in all tests we will have to add this variable in the context, out of this module.

pedrobaeza:
As said, you shouldn't disable the check about config and context key. Any other thing is a bad migration of the module, as this was working in 13.0. Please check differences. Anyway, it seems the error is only on runbot, so that's not blocking.

PR #857, opened 2021-10-15
[MIG] project_department: Migration to 14.0

polqubiq:


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0.

PR #864, opened 2021-11-25
[12.0][IMP] project_task_code: Added a migration script that will 'uniqify'…

ddejong-therp:
… the task codes, so that the new unique constraint can be applied without issues.

In 12.0, the project_task_code module adds a unique constraint to its code field that hasn't been unique before. This of course causes an issue when, for whatever reason, the DB contains 'duplicates', and you are migrating your database. I don't expect there to be a lot of duplicate task codes if nobody has messed with them manually, yet it can happen. However, empty task codes (with a value of NULL), _can_ be expected.
In either case, the duplicates get assigned a new task code.

The `pre_init_hook` does something similar, but does not handle duplicate codes that are not NULL.

### DIFF
--- a/project_task_code/migrations/12.0.1.1/pre-migration.py
+++ b/project_task_code/migrations/12.0.1.1/pre-migration.py
@@ -16,7 +16,7 @@ def uniqify_codes(env):
             env.cr.execute("SELECT id FROM project_task WHERE code = %s",
                 [task_code])
         else:
-            env.cr.execute("SELECT id FROM project_task"
+            env.cr.execute("SELECT id FROM project_task "
                            "WHERE code IS NULL OR code = '/'")
         similar_task_ids = [x['id'] for x in env.cr.dictfetchall()]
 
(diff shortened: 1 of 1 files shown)


ddejong-therp:
Additional comment:

During the migration of 10.0->11.0, the `project` module creates new tasks if you have `project.issue` issues in your database, because `project.issue`'s dissappear and get recreated as a `project.task`.

However, because these tasks get created in an Odoo 11.0 instance where the `project_task_code` module doesn't exist anymore, these tasks get created _without_ a new task code.

So that is a case where this migration script will help you out.

ddejong-therp:
If you were migration from Odoo 10.0, and have `project.issue`'s in you DB, those issues will be converted into new tasks.

Those tasks don't have the field `code` set, because they were created with an SQL query, not with the ORM api. Therefore, the inherited `create` function on `project.task` never gets called, which causes the `code` field to remain empty.

That should really be fixed in the `post-migration.py` of `project` module 11.0, but this script automatically takes care of it as well.

PR #854, opened 2021-09-30
[14.0] Migration of project_timeline_hr_timesheet

alan196:
### Proposed changes

Migration PR for module project_timeline_hr_timesheet.

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Migration Update

### Checklist

- [x] I have read the CONTRIBUTING doc
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation

PR #869, opened 2021-12-21
[14.0][FIX] Added timeline view type on project.task actions

ChrisOForgeFlow:
On version 14.0 exists other windows actions related to project.task, this commit ensure to add timeline view on every context
@ForgeFlow

Related with: #854

### DIFF
--- a/project_timeline/views/project_task_view.xml
+++ b/project_timeline/views/project_task_view.xml
@@ -50,6 +50,18 @@
             name="view_mode"
         >kanban,tree,form,calendar,timeline,pivot,graph,activity</field>
     </record>
+    <record id="project.action_project_task_user_tree" model="ir.actions.act_window">
+        <field name="view_mode">graph,pivot,timeline</field>
+    </record>
+    <record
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):

marcelsavegnago (approved):
LGTM

victoralmau (approved):

project_timeline/views/project_task_view.xml
```diff
@@ -58,11 +70,25 @@
             name="view_mode"
         >kanban,tree,form,calendar,timeline,pivot,graph,activity</field>
     </record>
+    <record id="project.action_view_all_task" model="ir.actions.act_window">
+        <field
+            name="view_mode"
+        >tree,kanban,form,calendar,timeline,pivot,graph,gantt,activity,map</field>
```
petrus-v, line 76: On my instance I haven't `map` nor `gantt` views available, so I'm not able to see task...

So there are missing dependencies or we should remove those views here and define that in an other module. I would prefer to avoid adding those dependencies here and move those view modes somewhere else.

PR #884, opened 2022-01-14
[IMP] project_task_project_required : project field mandatory by default

legalsylvain:
Change the default behaviour of ``project_task_project_required`` for new companies (or during installation).

the feature is now active by default.

**Rational**
 
- most of the odoo instances are mono-company. In that case, if we install this module, the expected behaviour is to have the feature enabled.
- for multi company instances, we can assume that if we install this module, we also want the feature enabled

**Impact**
- reduce configuration time.

CC : author : @alexey-pelykh

### DIFF
--- a/project_task_project_required/__manifest__.py
+++ b/project_task_project_required/__manifest__.py
@@ -19,4 +19,7 @@
         'views/project_task.xml',
         'views/res_config_settings.xml',
     ],
+    'demo': [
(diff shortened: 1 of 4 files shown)


HaraldPanten (approved):

MiquelRForgeFlow (approved):

andreampiovesana (approved):
merge?

PR #813, opened 2021-05-23
[14.0] project_stage_closed: deprecated module

petrus-v:
This module is useless as long odoo added the is_closed field
in version 14.0.

@pedrobaeza as discussed earlier in https://github.com/OCA/project/pull/765 !

PR #892, opened 2022-02-04
[15.0][MIG] project_task_dependency: migration

Mantux11:
There is only migration script to save old data. For now one Odoo 15 has dependencies, so this module is not longer needed.


pedrobaeza:
This should be done in OpenUpgrade, not here.

gaikaz:
@pedrobaeza
I see your point, tho Enterprise users upgrade via the official scripts. But then a module like this is needed in the new version, before it can be uninstalled, as uninstalling before the upgrade means losing data.
This is a very similar situation like one approved by yourself here: https://github.com/OCA/project/pull/813

Now it doesn't really hurt us as Odoo partners as we can do this anyway without the module actually being merged.
We just feel like helping the community too, as we do use these lovely quality modules! 🙂

pedrobaeza:
OK, go ahead then.

gaikaz:
Successfully migrated our internal Enterprise Odoo that is hosted odoo.sh with this module.

gaikaz (approved):
Successfully migrated our internal Enterprise Odoo that is hosted odoo.sh with this module.

BT-aestebanez:
Thanks for pointing me here:
Don't you think this query is also needed?

```
UPDATE project_project SET allow_task_dependencies=true
WHERE id in (SELECT project_id FROM project_task WHERE id in (SELECT task_id FROM project_task_dependency_task_rel));
```
Otherwise the page is not shown

gaikaz:
You are right. It would probably make sense to enable the setting like that.
Tho, for our Odoo, without much thought, I just enabled the option in "General Settings" and it seems to have enabled it on all the projects.

BT-aestebanez:
Yes, you are right
However.. @Mantux11 could you include this query?

Mantux11:
Good point here. I haven't thinked about that. @BT-aestebanez I have included your query.

BT-aestebanez (approved):
LGTM!
Thanks

ghost:
@dreispt @AaronHForgeFlow can you please merge this PR? Thank you

BT-aestebanez:
@rafaelbn Could you please review (and merge if you find it correct) this PR?

rafaelbn:
Hello @astirpe , what do you think as maintainer? Thanks in advance 😄

BT-aestebanez:
@rafaelbn it seems @astirpe does not have time. But @pedrobaeza said the approach was ok... I think we can merge it, don't you?

BT-aestebanez:
@dreispt Thanks.
This is the error message:
Error: reserveCache failed: Cache already exists. Scope: refs/heads/15.0-ocabot-merge-pr-892-by-dreispt-bump-nobump, Key: pre-commit-2-e233dd2121e64c861dfc8efd42bc49a71b455d4b158f53eb6973b5ed6703f473-6f934adb923502b30b518d023467313f206d1f64d803d89596ae85986792faeb, Version: 28cdb9f5496f334116f23e86f0063f5d3a9348c2e22425a33171e071aadada7e
 that prevents to merge

PR #901, opened 2022-02-09
[ADD] Auto fold task on Personal Stage when is closed

Shide:
This module adds the behavior to change the Personal Stage of the task to the first Folded one when the task reachs an stage with "Mark as Done".
If the Personal Stage is already folded, does nothing.

Simplifies the behavior of closing a task, go back to Personal Stages and move it to the Pesonal Done stage.

MT-240 @moduon


Shide:
Fold the Personal Task only applies to the user tat closes the task

project_task_personal_stage_auto_fold/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2022 Moduon - Eduardo de Miguel <edu@moduon.team>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Project Task Personal Stage auto Fold",
+    "summary": "Moves task to the first fold personal stage when done",
+    "version": "15.0.1.0.0",
+    "category": "Project",
+    "website": "https://github.com/OCA/project",
+    "author": "Moduon, " "Odoo Community Association (OCA)",
```
yajo:
```suggestion
    "author": "Moduon, Odoo Community Association (OCA)",
```

project_task_personal_stage_auto_fold/models/project_task.py
```diff
@@ -0,0 +1,42 @@
+from odoo import models
+
+
+class Task(models.Model):
+    _inherit = "project.task"
+
+    def write(self, vals):
+        result = super().write(vals)
+        if "stage_id" in vals and vals.get("stage_id"):
```
yajo: This can be simplified with:

```suggestion
        if vals.get("stage_id"):
```

project_task_personal_stage_auto_fold/models/project_task.py
```diff
@@ -0,0 +1,42 @@
...
+            if self.env["project.task.type"].browse(vals.get("stage_id")).is_closed:
```
yajo: AFAICS `is_closed` is a field in `project.task` instead.

https://github.com/odoo/odoo/blob/ad171e10f0f338c6fb81d98434e26a0860e801bc/addons/project/models/project.py#L1006

project_task_personal_stage_auto_fold/models/project_task.py
```diff
@@ -0,0 +1,42 @@
...
+                    ("fold", "=", True),
+                    ("is_closed", "=", True),
+                ],
+            )
+            .sorted(lambda ptt: ptt.is_closed, reverse=True)
+        )
+
+        for personal_not_folded_task in personal_tasks_not_folded:
+            if fold_stage:
+                personal_not_folded_task.stage_id = fold_stage[0]
```
yajo: This code contains a couple of `sudo()` calls that I think we could avoid. Also it uses the `is_closed` field in the wrong model, as explained above.

Also it doesn't take into account that you can close a task by moving it to "done" or to "cancelled". It's a bit tricky to get that 100% right, but there's some work that can be done to at least try it.

Besides, `project.task` already includes some helpers to easily move the personal stage, so we can reuse them.

I think this code will do. IMHO it's simpler and more accurate:

```suggestion
        # Find all folded personal stages
        folded_stages = (
            self.env["project.task.type"]
            .search(
                [
                    ("user_id", "=", self.env.user.id),
                    ("fold", "=", True),
                ],
            )
        )
        if not folded_stages:
            return
        # Allow to find personal stage with same name as shared stage
        stage_by_name = {stage.name: stage for stage in folded_stages}
        # Apply best matching personal stage
        for task in self:
            task.personal_stage_type_id = stage_by_name.get(
                task.stage_id.name,
                folded_stages[0],
            )
```

project_task_personal_stage_auto_fold/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This module adds the behavior to change the Personal Stage of the task to the first
+Closed or Folded (Closed Personal Stages has high priority than Folded) stage when
+the task reachs an Stage marked as Closed Stage.
```
yajo, line 3: If you apply my above suggestion, then add this please:

```suggestion
This module adds the behavior to change the Personal Stage of the task to the first
Closed or Folded (Closed Personal Stages has high priority than Folded) stage when
the task reachs an Stage marked as Closed Stage.

If there's a folded personal stage with the same name as the folded shared stage,
that one will be used.

```

yajo (changes requested):
I think I've spotted some bugs and you have a couple of nice enhancements.

Please also remove the readme sections that are empty (USAGE.rst and README.rst).

Other than that this is awesome and totally needed. Thanks! ❤️

project_task_personal_stage_auto_fold/models/project_task.py
Shide: `is_closed` field on Tasks means the task is folded, not on a closing stage. 
I want to move the task to a "done" personal stage when the task is really done.

This is why I'm checking the field `is_closed` on the stage, not on the task.

The field on the task should may be called `is_folded` to avoid confusion.

project_task_personal_stage_auto_fold/models/project_task.py
Shide: Your suggestion works fine and it's better. 
I'm just going to add the sorted for `is_closed` field first on the `project.task.type` search.

project_task_personal_stage_auto_fold/readme/DESCRIPTION.rst
Shide: Done

project_task_personal_stage_auto_fold/models/project_task.py
yajo: But AFAICS, `project.task.type` doesn't have `is_closed` field, right? See https://github.com/odoo/odoo/blob/ad171e10f0f338c6fb81d98434e26a0860e801bc/addons/project/models/project.py#L63-L105

So, in Odoo 15, folded and closed stages are synonyms. At least, that's what I understand... Am I missing something?

pedrobaeza: On current 15.0, there's a `is_closed` field on project.task.type:

https://github.com/odoo/odoo/blob/2ba7e054b75c7408588baca2179715b0e3af8160/addons/project/models/project.py#L101

We are talking about removing `project_task_stage_closed` OCA module due to this.

yajo: Ah! I must have been looking v14 code all this time 🤦🏼‍♂️

yajo: Ah no, I was looking on master. So, indeed, in master, they remove the "closed" concept and make folded = closed.

https://github.com/odoo/odoo/commit/f4c054e4731270a5103ec1d5a97c2789aef977e3#diff-dcdc6e8a2e65aff0582c53d72c3890c1b56725a8b9fcf7abbb0378097b5999afL965-R962

pedrobaeza: Oh, that's ugly. For me it's not the same...

yajo (approved):
Code & functional review, it looks perfect now!

rafaelbn (approved):
Functional review 👍 tested in runboat 🎉

PR #725, opened 2020-10-11
[13.0] mig project_task_material_stock

jobiols:
Hello @rvalyi Thank you for your support. I decided to start clean from a new PR, tell me if I am now on the right track.

PR #904, opened 2022-02-18
[13.0][MIG] project_task_material_stock: Migration to 13.0

victoralmau:
Superseed: https://github.com/OCA/project/pull/725

Migration to 13.0

@Tecnativa TT34554


victoralmau:
The error in runboat seems to be related to `project_task_markdown` the `web_widget_text_markdown` dependency is not migrated to 13.0 yet (https://github.com/OCA/web/pull/2136)

project_task_material_stock/data/data.xml
```diff
@@ -0,0 +1,33 @@
...
+            <field name="code">outgoing</field>
+            <field name="sequence_code">PTM</field>
+            <field name="sequence_id" ref="project_task_material_sequence" />
+            <field name="warehouse_id" ref="stock.warehouse0" />
+            <field name="default_location_src_id" ref="stock.stock_location_stock" />
+            <field
+                name="default_location_dest_id"
+                ref="stock.stock_location_customers"
+            />
+        </record>
```
pedrobaeza, line 31: On v13, this is not hardcoded on data, but populated on this method:

https://github.com/odoo/odoo/blob/b78ab99d1426d2fbabaa5b272ee9e42fc1a66113/addons/stock/models/stock_warehouse.py#L262

and here you inject a new picking type:

https://github.com/odoo/odoo/blob/b78ab99d1426d2fbabaa5b272ee9e42fc1a66113/addons/stock/models/stock_warehouse.py#L830

But I'm doubting if a new picking should be populated instead of using only `stock.move` records without picking. It's the technique used by `mrp` module. Even more, they present in the production form directly `stock.move` and  `stock.move.line` records to manipulate the quantities and products to be consumed or to produce.

I think we should get rid off the model `project.task.material`, and use instead `stock.move` / `stock.move.line` as the way to add products and quantities.

Check how it's done on `mrp`:

https://github.com/odoo/odoo/blob/b78ab99d1426d2fbabaa5b272ee9e42fc1a66113/addons/mrp/models/mrp_production.py#L168-L178

pedrobaeza (changes requested):

project_task_material_stock/data/data.xml
victoralmau: Finally, the suggested changes have not been introduced in this addon, therefore this MR would be ready for review.

A new addon https://github.com/OCA/project/pull/905 has been created with these comments using only stock moves.

chienandalu:
> The error in runboat seems to be related to `project_task_markdown` the `web_widget_text_markdown` dependency is not migrated to 13.0 yet ([OCA/web#2136](https://github.com/OCA/web/pull/2136))

The quick fix could be to remove the hard dependency.

pedrobaeza:
Closing as moved to the new addon #905

PR #877, opened 2022-01-05
[MIG][14.0] project_hr

AntoniRomera:
Standard migration

PR #910, opened 2022-03-17
[14.0][MIG] project_hr: Migration to 14.0

ernestotejeda:
Cc @Tecnativa TT34131

Main changes:
- This module adds a new field named allowed_user_ids at project.task level but a field with that name (and with another purpose) already exists in the Odoo core 'project' module, so I renamed the new field

Supersede https://github.com/OCA/project/pull/877 and https://github.com/OCA/project/pull/776 shuld be closed too


ernestotejeda:
ping @victoralmau

victoralmau (approved):
Code review OK

project_hr/models/project_task.py
```diff
@@ -29,7 +29,9 @@ class ProjectTask(models.Model):
         help="Technical field for computing allowed employee categories "
         "according categories at project level.",
     )
-    allowed_user_ids = fields.Many2many(
+    # This field could have been named allowed_user_ids but a field with
+    # that name already exists in the Odoo core 'project' module
+    allowed_assigned_user_ids = fields.Many2many(
         comodel_name="res.users",
         string="Allowed users",
         compute="_compute_allowed_user_ids",
```
pedrobaeza, line 37: Rename the compute method as well

pedrobaeza (approved):

project_hr/models/project_task.py
ernestotejeda: I didn't realize that, in the next chance I will change it if someone else has not changed it, luckily the name of the method of the field allowed_user_ids of odoo core is different (_compute_allowed_users) :)

pedrobaeza: Done myself in ad505ddc9bef7edf2217ee99356e52396b324a61

PR #912, opened 2022-04-11
[13.0][FIX] project_stock: Avoid permissions error if the user does not have access to stock.

victoralmau:
Avoid permissions error if the user does not have access to stock.

@Tecnativa TT35899

### DIFF
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -186,7 +186,8 @@ class Task(models.Model):
         if "stage_id" in vals:
             stage = self.env["project.task.type"].browse(vals.get("stage_id"))
             if stage.done_stock_moves:
-                self.action_confirm()
+                # Avoid permissions error if the user does not have access to stock.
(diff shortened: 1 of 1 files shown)


project_stock/models/project_task.py
```diff
@@ -186,7 +186,11 @@ def write(self, vals):
         if "stage_id" in vals:
             stage = self.env["project.task.type"].browse(vals.get("stage_id"))
             if stage.done_stock_moves:
-                self.action_confirm()
+                # Avoid permissions error if the user does not have access to stock.
+                _self = self
+                if not self.env.user.has_group("stock.group_stock_user"):
+                    _self = self.sudo()
```
pedrobaeza: Do sudo directly. This is not v12 where we have a trade-off of changing the user.

pedrobaeza (approved):

pedrobaeza:
Let's fast-track it for avoiding the permission problem:

/ocabot merge minor

PR #913, opened 2022-04-12
[13.0][FIX] project_stock: Update the onchange of the task so that it does not behave incorrectly when creating stock moves.

victoralmau:
Update the onchange of the task so that it does not behave incorrectly when creating stock moves.

@Tecnativa TT35602

### DIFF
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -100,18 +100,20 @@ class Task(models.Model):
             item.unreserve_visible = not any_quantity_done and already_reserved
 
     @api.onchange("picking_type_id")
-    def onchange_picking_type(self):
-        self.location_id = self.picking_type_id.default_location_src_id.id
-        self.location_dest_id = self.picking_type_id.default_location_dest_id.id
+    def _onchange_picking_type_id(self):
+        if self.picking_type_id:
+            self.location_id = self.picking_type_id.default_location_src_id.id
+            self.location_dest_id = self.picking_type_id.default_location_dest_id.id
 
-    @api.onchange("location_id", "move_ids")
-    def _onchange_location(self):
-        self.move_ids.update(
-            {
-                "warehouse_id": self.location_id.get_warehouse().id,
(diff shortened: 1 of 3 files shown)


pedrobaeza:
This if fixing the consequences, but not the real root cause. Why some of them are not filled? Are you doing things through onchanges and the fields don't exist? Maybe you need to give read permission to some master data like locations, and only hide elements, not totally remove them (group inside the arch, not in the view).

victoralmau:
> This if fixing the consequences, but not the real root cause. Why some of them are not filled? Are you doing things through onchanges and the fields don't exist? Maybe you need to give read permission to some master data like locations, and only hide elements, not totally remove them (group inside the arch, not in the view).

Ok sorry, you're right. The correct change is already applied.

project_stock/models/project_task.py
```diff
@@ -101,17 +101,19 @@ def _compute_unreserve_visible(self):
 
     @api.onchange("picking_type_id")
     def onchange_picking_type(self):
```
pedrobaeza: Naming guidelines:

```suggestion
    def _onchange_picking_type_id(self):
```

project_stock/models/project_task.py
```diff
@@ -101,17 +101,19 @@ def _compute_unreserve_visible(self):
...
-        self.location_id = self.picking_type_id.default_location_src_id.id
-        self.location_dest_id = self.picking_type_id.default_location_dest_id.id
+        if self.picking_type_id:
+            self.location_id = self.picking_type_id.default_location_src_id.id
+            self.location_dest_id = self.picking_type_id.default_location_dest_id.id
 
-    @api.onchange("location_id", "move_ids")
+    @api.onchange("location_id")
     def _onchange_location(self):
```
pedrobaeza:
```suggestion
    def _onchange_location_id(self):
```

pedrobaeza (approved):

victoralmau:
Included in https://github.com/OCA/project/pull/924

PR #914, opened 2022-04-20
[13.0][IMP] project_stock: Confirm the new stock movements in the "Check availability materials" button

victoralmau:
Confirm the new stock movements in the "**Check availability materials**" button (if movements have been added after confirming the materials).

@Tecnativa TT35602

### DIFF
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -118,10 +118,10 @@ class Task(models.Model):
         return {"name": values["name"]}
 
     def action_confirm(self):
-        self.mapped("move_ids").write({"state": "confirmed"})
-        self.action_assign()
+        self.mapped("move_ids")._action_confirm()
 
     def action_assign(self):
+        self.action_confirm()
         self.mapped("move_ids")._action_assign()
 
     def button_scrap(self):
@@ -187,7 +187,7 @@ class Task(models.Model):
             stage = self.env["project.task.type"].browse(vals.get("stage_id"))
             if stage.done_stock_moves:
                 # Avoid permissions error if the user does not have access to stock.
-                self.sudo().action_confirm()
+                self.sudo().action_assign()
         return res
 
     def unlink(self):
--- a/project_stock/tests/common.py
+++ b/project_stock/tests/common.py
@@ -14,6 +14,9 @@ class TestProjectStockBase(common.SavepointCase):
         cls.product_b = cls.env["product.product"].create(
             {"name": "Test product B", "standard_price": 20}
         )
+        cls.product_c = cls.env["product.product"].create(
+            {"name": "Test product C", "standard_price": 0}
+        )
         warehouse = cls.env["stock.warehouse"].search(
             [("company_id", "=", cls.env.company.id)], limit=1
         )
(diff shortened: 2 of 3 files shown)


project_stock/models/project_task.py
```diff
@@ -122,6 +122,9 @@ def action_confirm(self):
         self.action_assign()
 
     def action_assign(self):
+        self.move_ids.filtered(lambda x: x.state == "draft").write(
```
pedrobaeza: Why not calling the method to confirm instead?

project_stock/models/project_task.py
```diff
@@ -118,10 +118,11 @@ def _prepare_procurement_group_vals(self, values):
         return {"name": values["name"]}
 
     def action_confirm(self):
-        self.mapped("move_ids").write({"state": "confirmed"})
+        self.mapped("move_ids")._action_confirm()
         self.action_assign()
 
     def action_assign(self):
+        self.move_ids.filtered(lambda x: x.state == "draft")._action_confirm()
```
pedrobaeza: Call `self.action_confirm()`

project_stock/models/project_task.py
```diff
@@ -118,11 +118,11 @@ def _prepare_procurement_group_vals(self, values):
         return {"name": values["name"]}
 
     def action_confirm(self):
-        self.mapped("move_ids").write({"state": "confirmed"})
-        self.action_assign()
+        self.mapped("move_ids")._action_confirm()
+        self.mapped("move_ids")._action_assign()
```
pedrobaeza, line 125: Why is this here?

victoralmau: This change is related to https://github.com/OCA/project/pull/914#discussion_r854118954.

I think it's good to explain the process to understand everything.

**Before:**
- "Confirm Materials" button confirmed stock moves with the `action_confirm()` method.
- "Check availability" button called the `_action_assign()` method to reserve the stock.
- Additionally, "Confirm materials" button called the method of the "Check availability" button (to avoid clicking on confirm and always check availability).

**Actual use case:**
- The task is changed to "In Progress" stage.
- Stock moves are added and confirmed with "Confirm materials" button.
- The task is changed to "Completed" stage.
- The task is changed back to "In Progress" stage.
- Some extra stock moves is added.
- "Check availability" button does nothing for the new line because the status of line is draft (not confirmed).

**Changes done:**
- [x] Changed `self.mapped("move_ids").write({"state": "confirmed"})` to `self.mapped("move_ids")._action_confirm()`.
- [x] Changed `self.action_assign()` to `self.mapped("move_ids")._action_assign()`.
- [x] Changed the `action_assign()` event to call `action_confirm()` event and solve the previous use case.

pedrobaeza: My comment was not asking for that. There's no sense to have 2 methods that do the same

project_stock/models/project_task.py
```diff
@@ -118,10 +118,7 @@ def _prepare_procurement_group_vals(self, values):
         return {"name": values["name"]}
 
     def action_confirm(self):
-        self.mapped("move_ids").write({"state": "confirmed"})
-        self.action_assign()
-
-    def action_assign(self):
+        self.mapped("move_ids")._action_confirm()
         self.mapped("move_ids")._action_assign()
```
pedrobaeza: It seems I'm not able to transfer you what I mean. Let me put the code:

```suggestion
    def action_confirm(self):
        self.mapped("move_ids")._action_confirm()

    def action_assign(self):
        self.action_confirm()
        self.mapped("move_ids")._action_assign()
```

pedrobaeza (approved):

sergio-teruel (approved):

PR #922, opened 2022-05-12
[14.0][ADD] project_dev: initial commit

AshishHirapara:
This module adds a field "Devs" in project tasks with two goals: 

- Tracking which dev(s) are working on a specific task.
- Restricting devs access only to tasks where they are assigned as devs, 
  in order to reduce for them the noise of non-dev tasks on the kanban.


francesco-ooops (approved):
Functional review ok!

SicurSam (approved):
LGTM

antoniocanovas (approved):
Functional Ok

project_dev/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,4 @@
+For setting employee categories at project level:
```
francesco-ooops: To restrict project task visibility to records where "devs" field is set for a user:

project_dev/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+This module adds a field "Devs" in project tasks with two goals:
+
+- Tracking which dev(s) are working on a specific task.
+- Restricting devs access only to tasks where they are assigned as devs,
```
francesco-ooops: Restricting "devs" users visibility only to tasks where they are assigned as devs,

project_dev/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
...
+  in order to reduce for them the noise of non-dev tasks on the kanban.
```
francesco-ooops, line 5: Furthermore:
- subtasks inherit "Devs" values from their parent tasks
- a control is in place when setting an activity to a user with "only devs tasks" enabled in a task that doesn't have that user set as dev.

francesco-ooops (commented):
some updates to module description/usage

project_dev/readme/CONFIGURE.rst
AshishHirapara: done

project_dev/readme/DESCRIPTION.rst
AshishHirapara: done

project_dev/readme/DESCRIPTION.rst
AshishHirapara: done

francesco-ooops:
@OCA/project-service-maintainers can you take a look?

dreispt:
What if the project has no devs, but engineers or other kind of workers?
Would this feature also be useful in these other settings?

francesco-ooops:
> What if the project has no devs, but engineers or other kind of workers? Would this feature also be useful in these other settings?

Hi @dreispt, that's a very good question. I think yes, this module is useful in any case where: 

- project manager wants to track which worker is assigned to a task
- help worker display only tasks that are relevant for them 

I could not think of a solution to allow user to customize the label of the field "devs" in order to reflect their role (dev, engineer, etc.) - do you have any suggestion?

Otherwise we can just rename "devs" to "workers" or "assignees" (or any other term you suggest) to make it generic enough for any situation.

dreispt:
@francesco-ooops I would replace every reference to "Devs" with "Team" (or "team members").
Ideally that would also be for the module name, if not too much of a stretch for you (I know changing a module name can be pailful)

alexey-pelykh:
It seems that this module reimplements existing feature of project roles and assignments https://apps.odoo.com/apps/modules/12.0/project_role/

alexey-pelykh (approved):
I don't thing we should introduce new module that either duplicates or is very close to already existing module that needs porting, especially since existing module has features like synchronization with jira and advanced integration with billing

alexey-pelykh (changes requested):
.

francesco-ooops:
> I don't thing we should introduce new module that either duplicates or is very close to already existing module that needs porting

Hi @alexey-pelykh , maybe I did not fully understand the flow of project_role, but it looks to me they provide quite different features. AFAIU: 

1. scope
- project_role works at project level, as in: create role, assign user(s) to role, assign roles to project.

- project_dev (or the name we'll change it to to make it more generic) works at user+task level: set role in user, set user in task
eg: "this specific task is "assigned to" the project manager, but the following team will work on it too: worker1, worker2, worker3"

2. restrict features for workers

- project_dev: restrict access to relevant tasks

- project_role: such a feature is not contemplated since it's already possible to restrict access at project level

So overall, sure the existing module could be extended to provide features present in the other, but that would be justified if there was an overlapping of features, and it doesn't look the case to me; furthermore, that would take time and money, as the two modules have been created for quite different use cases

> especially since existing module has features like synchronization with jira and advanced integration with billing

These could be interesting features, but I haven't seen those mentioned in the module description?

AshishHirapara:
Based on the conversation above, the updated new PR is https://github.com/OCA/project/pull/930

PR #924, opened 2022-05-23
[13.0][FIX] project_stock: Update locations on stock moves + disallow update if there is any assigned move.

victoralmau:
Update locations on stock moves + disallow update if there is any assigned move.

@Tecnativa TT36812

### DIFF
--- a/project_stock/models/project_project.py
+++ b/project_stock/models/project_project.py
@@ -32,6 +32,14 @@ class ProjectProject(models.Model):
     )
 
     @api.onchange("picking_type_id")
-    def onchange_picking_type(self):
+    def _onchange_picking_type_id(self):
         self.location_id = self.picking_type_id.default_location_src_id.id
         self.location_dest_id = self.picking_type_id.default_location_dest_id.id
+
+    def write(self, vals):
+        """Update location information on pending moves when changed."""
+        res = super().write(vals)
+        field_names = ("location_id", "location_dest_id")
+        if any(vals.get(field) for field in field_names):
+            self.task_ids._update_moves_info()
+        return res
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -1,6 +1,7 @@
 # Copyright 2022 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 
 
 class Task(models.Model):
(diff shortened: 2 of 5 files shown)


sergio-teruel (approved):

project_stock/models/project_project.py
```diff
@@ -31,7 +32,28 @@ class ProjectProject(models.Model):
...
+
     @api.onchange("picking_type_id")
-    def onchange_picking_type(self):
+    def onchange_picking_type_id(self):
         self.location_id = self.picking_type_id.default_location_src_id.id
         self.location_dest_id = self.picking_type_id.default_location_dest_id.id
+
+    @api.onchange("location_id")
+    def onchange_location_id(self):
+        self._check_tasks_with_pending_moves()
```
pedrobaeza: This must be done on write as well

project_stock/models/project_project.py
```diff
@@ -31,7 +32,28 @@ class ProjectProject(models.Model):
         help="Default location to which materials are consumed.",
     )
 
+    def _check_tasks_with_pending_moves(self):
+        if "assigned" in self.task_ids.mapped("move_ids.state"):
```
pedrobaeza: Please don't repeat logic. Do `self.task_ids._check_tasks_with_pending_moves()`. And the same for the rest of the code.

pedrobaeza (changes requested):

project_stock/models/project_project.py
```diff
@@ -31,7 +31,25 @@ class ProjectProject(models.Model):
...
+        self.task_ids._check_tasks_with_pending_moves()
+
     @api.onchange("picking_type_id")
-    def onchange_picking_type(self):
+    def onchange_picking_type_id(self):
         self.location_id = self.picking_type_id.default_location_src_id.id
         self.location_dest_id = self.picking_type_id.default_location_dest_id.id
+
+    @api.onchange("location_id")
+    def onchange_location_id(self):
```
pedrobaeza: Use convention and avoid that this can be publicly called (the same with the rest):

```suggestion
    def _onchange_location_id(self):
```

Anyways, I'm thinking this should be done only on write, not on onchange, as anyways it should be checked again on that final step.

project_stock/models/project_task.py
```diff
@@ -104,14 +105,35 @@ def onchange_picking_type(self):
...
+    def onchange_location_id(self):
+        self._check_tasks_with_pending_moves()
+
+    @api.onchange("location_dest_id")
+    def onchange_location_dest_id(self):
+        self._check_tasks_with_pending_moves()
+
+    def _update_moves_info(self):
+        self._check_tasks_with_pending_moves()
+        picking_type = self.picking_type_id or self.project_id.picking_type_id
```
pedrobaeza: This is `self.ensure_one()` right now, but it's being called for several records.

pedrobaeza (changes requested):

project_stock/models/project_task.py
```diff
@@ -100,18 +101,32 @@ def _compute_unreserve_visible(self):
             item.unreserve_visible = not any_quantity_done and already_reserved
 
     @api.onchange("picking_type_id")
-    def onchange_picking_type(self):
+    def _onchange_picking_type(self):
```
pedrobaeza:
```suggestion
    def _onchange_picking_type_id(self):
```

project_stock/models/project_project.py
```diff
@@ -32,6 +32,13 @@ class ProjectProject(models.Model):
     )
 
     @api.onchange("picking_type_id")
-    def onchange_picking_type(self):
+    def _onchange_picking_type_id(self):
         self.location_id = self.picking_type_id.default_location_src_id.id
         self.location_dest_id = self.picking_type_id.default_location_dest_id.id
+
+    def write(self, vals):
+        res = super().write(vals)
```
pedrobaeza, line 41: Remember to always put proper docstrings:
```suggestion
        """Update location information on pending moves when changed."""
        res = super().write(vals)
```

project_stock/models/project_project.py
```diff
@@ -32,6 +32,13 @@ class ProjectProject(models.Model):
...
+        field_names = ("picking_type_id", "location_id", "location_dest_id")
```
pedrobaeza: Should `picking_type_id` include here? I don't think so, as the sequence is already gathered.

project_stock/models/project_task.py
```diff
@@ -100,18 +101,32 @@ def _compute_unreserve_visible(self):
...
+                _("It is not possible to change this with reserved movements in tasks.")
+            )
+
+    def _update_moves_info(self):
+        self._check_tasks_with_pending_moves()
+        for item in self:
+            picking_type = item.picking_type_id or item.project_id.picking_type_id
+            location = item.location_id or item.project_id.location_id
+            location_dest = item.location_dest_id or item.project_id.location_dest_id
+            moves = item.move_ids.filtered(lambda x: x.state not in ("cancel", "done"))
```
pedrobaeza: Filter also for those with different data (`location_id != location` and so on).

pedrobaeza (approved):

PR #927, opened 2022-05-27
[14.0] [IMP] Allow set duration on timesheet time control

angelmoya:
Previous this improvement when user launch wizard to "Start work", user can set "Start Time" and optionally "End Time".

If user set "End Time" an timesheet line is created with computed duration.

With this PR user can set "Start Time" and "Duration".

If user usually create thimesheet line with duration instead of start/stop, this way es quickly than edit task, create new line and save task.


flotho (approved):
thanks
tested on runboat,
LGTM

pegonzalezs:
LGTM @angelmoya 

Did validate on runboat

Just a minor thing: 
Would it make sense that if the "End Date" is introduced the "duration" field becomes invisible?

PR #510, opened 2019-04-26
[11.0][WIP] project_team

tarteo:
Create and manage project teams.

TODO:

- [ ] domain on team in project.task
- [ ] access rights for team: project user should be able to see the tasks where they are a member of the 
 assigned team
- [ ] access rights for multi company
- [ ] readme

PR #930, opened 2022-06-13
[14.0][ADD] project_task_team: initial commit

AshishHirapara:
This module adds a field "Task Team" in project tasks with two goals:

Tracking which team member(s) are working on a specific task.
Restricting team-member access only to tasks where they are assigned a team member, in order to reduce for them the noise of non-team-member tasks on the kanban and easily visualize only 'Task Team' project tasks.

update of https://github.com/OCA/project/pull/922


AshishHirapara:
Hello @OCA/project-service-maintainers, can any one help me with this PR's travis fail? I am running the tests in my local and there are no error is getting generated. So I want to know if there is anything that I can do to get more details about this fail? Or is there any way to debug the test issue? So I can fix it ASAP.

project_task_team/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,4 @@
+To restrict project task visibility to records where "Task Team" field is set for a user:
+
+#. Go to Settings > Users:
+#. Enable the extra rights for user :"Visualize only 'Task Team' project tasks"
```
francesco-ooops: To allow the user to be selected as Task Team member without restricting the task visibility for them, enable extra right "Allow to be selected as task team"

francesco-ooops (approved):
Functional review ok

francesco-ooops:
@dreispt since you commented on the original PR #922, could you take a look at this updated version? Thanks!

francesco-ooops:
> @dreispt since you commented on the original PR #922, could you take a look at this updated version? Thanks!

also @antoniocanovas and @alexey-pelykh who reviewed as well :)

SicurSam (approved):
Functional review ok!

project_task_team/readme/CONFIGURE.rst
AshishHirapara: done

francesco-ooops:
@OCA/project-service-maintainers could you take a look? :)

francesco-ooops:
@OCA/project-service-maintainers just a reminder :)

francesco-ooops:
@OCA/project-service-maintainers @dreispt @alexey-pelykh @rafaelbn weekly ping

antoniocanovas:
Hello @francesco-ooops , here is my opinion:

I can't validate this module, I would design in other way.

Now as designed, if admin or any project manager is assigned as "Visualize only task team" to take advance of the functionality, they can't admin because there is no way to see other tasks.
The boolean "Allow to be selected as task team" may be good if there are many users but I think that when time goes on, all task users will have this boolean activated, and it is confusing at first.

To avoid this, I think it could be in this way:
- New group with less permission, visible in Project permission selection: **"Team task"**, "User" or "Manager"
- This group can only see his "team task" by domain in one specific rule.
- Add a search filter to be used by any user (user, manager o admin) "My team tasks".
- New filter while writing on search bar "Search user in team users" 

In this way basic "Team task users" can take advance of the same original functionality, and managers will use too.

Thanks, and sorry for delay.

francesco-ooops:
Hi @antoniocanovas , thank you for your feedback! To address your doubts:

> Now as designed, if admin or any project manager is assigned as "Visualize only task team" to take advance of the functionality

There would be no reason to flag admins or PMs as "Visualize only task team", because there isn't any advantage to gain with that flag that can't be achieved with a filter by user.

The flag "Allow to be selected as task team" was created with the purpose of helping the user selecting task team members (in our use case: the PM) to find quickly the users that are supposed to be selected for that field (in our case: devs).

If we change "Allow to be selected as task team" from opt-in to opt-out (renaming it "Do not allow in task team"), would that work? In that case: every user can be selected as task team member by default, and if selection needs to be restricted, it's always possible to do it.

does that address your concerns?

francesco-ooops:
@antoniocanovas could you take a look at my message so we can proceed? thanks!

AaronHForgeFlow:
Are the task teams res.users objects? Why not creating a new model or using mail activity teams?

francesco-ooops:
> Are the task teams res.users objects? Why not creating a new model or using mail activity teams?

thanks for your feedback @AaronHForgeFlow! can you expand further on your proposal?

AaronHForgeFlow:
I mean, users should be actual users of the system, in enterprise version those users will involve another subscription for a user that is not an actual user. Is there benefit I am missing for using that?

Another option is to use a separate model, like this: https://github.com/OCA/project/pull/510/files#diff-80507e2757b3e068dc039af2b2383568a0f7597dd11fa31f893c34467702a057 Potentially you can combine this module with https://github.com/OCA/social/blob/14.0/mail_activity_team. So for example one project team will use an activity teams so each member can see the activities of the team.

This is just an idea that came out of the blue, i will not block this PR.

fufi07:
I am beginner 🔰 please guide me GitHub

francesco-ooops:
@AshishHirapara you can close this one

PR #934, opened 2022-06-17
[14.0][IMP] project_task_pull_request:   PR URI in main task view

ilyasProgrammer:


francesco-ooops (approved):
Functional review ok 

This PR allows to use PR URI field without developer mode, without changing current behavior

It also clarifies in description that PR URI field is only accessible through dev mode (lost quite some time looking for it)

@ilyasProgrammer can you mark as ready for review?

pedrobaeza:
Isn't better to just remove the group on existing field? FYI, we use `base_technical_features` OCA module and assign such group to everyone, as we haven't never understood the reason for hiding the fields in a "basic mode". I call it instead "limited mode".

francesco-ooops:
> Isn't better to just remove the group on existing field? FYI, we use `base_technical_features` OCA module and assign such group to everyone, as we haven't never understood the reason for hiding the fields in a "basic mode". I call it instead "limited mode".

I'll check with @ilyasProgrammer, but 
1) we didn't want to change anything from current behavior for other module users and 
2) wanted to (allow to) move the field to a more prominent position in the form, as it's easy to forget to fill a field that is in a secondary tab

(required field does not work in our case as not all tasks are development tasks)

pedrobaeza:
The module itself brings a mechanism for doing such field required when going to a certain stage. You should use such mechanism.

francesco-ooops:
> (required field does not work in our case as not all tasks are development tasks)

I find that mechanism too restrictive because it implies that all tasks must be related to a PR, while some tasks are pm related, eg: "import data" or "test with customer function x"

basically all I wanted was to have a PR field in main task form where I can quickly check if there's a PR related to the task :)

Do you think there's an easy way to achieve that?

pedrobaeza:
You select the restriction by project, so it's not restrictive at all, and you also select in which stages do you want to apply them. I have the stage "Testing" so tasks that don't require a pull request don't go through that stage.

francesco-ooops:
> You select the restriction by project, so it's not restrictive at all, and you also select in which stages do you want to apply them. I have the stage "Testing" so tasks that don't require a pull request don't go through that stage.

I understood how the module works, but it's not applicable to our flow as all projects have the same to do/processing/revision/acceptance/deploy/done stages. Some tasks have PRs from stage 1 to end, some never have them, and that's the same across projects. I appreciate the guidance anyway :)

But going back to the PR goal: all I'd like to have is a PR field in main task form where I can quickly click and move from project to github. Functionally, seemed to me this PR is a good way to proceed as it allows to maintain both current behavior and implement my request. What do you think?

pedrobaeza:
Well, it's to complicate a module and its maintenance adding parameters. If a good number of this module users want to move the field to top and to remove the technical group, you can directly propose that. I can accept such changes.

francesco-ooops:
> If a good number of this module users want to move the field to top and to remove the technical group, you can directly propose that.

I would like to do that, but it's not clear to me how such a poll should be conducted. AFAIK I don't think there's a discussion space available to OCA where I could intercept this specific module's users opinions. I could message on discord or open an issue on this repo but we know it would probably stay silent forever (understandably). So overall the solutions I see are: 

1) have a PR (this one) allowing both possibilities and get it merged. I don't think 78 lines will have much of an impact on OCA, although I understand small numbers adding up... but I don't think setting the field in a hidden tab was great design from the beginning.

2) have a PR moving the field to main form view and get it merged (those who will miss to see the PR and currently use the module and liked the field where it is... well, too bad for them)

3) keep the not-so-great design on OCA, make an internal module with the field where I want and move on :)

What would you do in my position?

francesco-ooops:
oh I forgot, would adding tests offset the maintenance technical debt? or be a good condition to take in charge mainteinance from this PR?

pedrobaeza:
I would go for 2, and those not happy can stay in a previous version of the module.

francesco-ooops:
> I would go for 2, and those not happy can stay in a previous version of the module.

alright thanks! :)

project_task_pull_request/__manifest__.py
```diff
@@ -12,5 +12,7 @@
     "application": False,
     "installable": True,
     "depends": ["project"],
-    "data": ["views/project_task_pull_request_view.xml"],
+    "data": [
+        "views/project_task_pull_request_view.xml",
```
pedrobaeza: This shouldn't be changed.

project_task_pull_request/readme/HISTORY.rst
```diff
@@ -0,0 +1,4 @@
+14.0.1.0.0
```
pedrobaeza: Remove this

project_task_pull_request/readme/USAGE.rst
```diff
@@ -6,4 +6,4 @@ To use this module, you need to:
 #. Go to Dashboard and select a project
 #. Attempt to move one of the project's task without a PR URI into one of
    the stages you selected to require a PR; you will receive a Validation Error
-#. To add a PR URI to a task, click on the task and go to the "Extra Info" tag
+#. To add a PR URI to a task, click on the task ans set it.
```
pedrobaeza:
```suggestion
#. To add a PR URI to a task, click on the task and set it in the corresponding field in the header.
```

pedrobaeza (changes requested):

pedrobaeza:
The file `HISTORY.rst` is still there, although empty.

pedrobaeza (approved):

pedrobaeza:
Let's give now some time to other people to review this.

francesco-ooops (approved):
Functional review ok!

Now field is always accessible and can be selected in search bar :)

pedrobaeza:
Maybe it's good to leave the conditional visibility on the "Other information" tab

francesco-ooops:
> Maybe it's good to leave the conditional visibility on the "Other information" tab

Can you clarify how should user choose whether field should be visible or not?

pedrobaeza:
Both are shown, for old and new habits

francesco-ooops:
@pedrobaeza is it ok? currently it is displayed twice (header and extra info tab), but I'm not sure if that's the conditional visibility you were referring to

project_task_pull_request/views/project_task_pull_request_view.xml
```diff
@@ -9,11 +9,10 @@
         <field name="arch" type="xml">
             <field name="sequence" position="before">
                 <field name="pr_required_states" attrs="{'invisible':True}" />
-                <field
-                    name="pr_uri"
-                    widget="url"
-                    attrs="{'invisible':[('pr_required_states', '=', [])]}"
```
pedrobaeza, line 15: I mean to not touch this field and leave it as it was. Just add the other.

pedrobaeza (approved):

francesco-ooops:
@pedrobaeza good for me too! can we merge?

pedrobaeza:
No, let's wait a bit for other reviewers if they have any complain.

francesco-ooops:
tagging @OCA/project-service-maintainers to check if anyone has an opinion on this, thanks! :)

leemannd (approved):

PR #936, opened 2022-06-24
[13.0][FIX+IMP] project_stock: Prevent merge moves + Show buttons when they are really needed..

victoralmau:
Changes done:
- [x] Auto-create procurement.group in task to use it in moves to prevent merge moves from another tasks.
- [x] Show buttons when they are really needed. 

@Tecnativa TT37708

### DIFF
--- a/project_stock/__manifest__.py
+++ b/project_stock/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Project Stock",
-    "version": "13.0.1.1.2",
+    "version": "13.0.2.0.0",
     "category": "Project Management",
     "website": "https://github.com/OCA/project",
     "author": "Tecnativa, Odoo Community Association (OCA)",
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -79,6 +79,7 @@ class Task(models.Model):
         inverse_name="stock_task_id",
         string="Analytic Lines",
     )
+    group_id = fields.Many2one(comodel_name="procurement.group",)
 
     def _compute_scrap_move_count(self):
         data = self.env["stock.scrap"].read_group(
@@ -151,8 +152,8 @@ class Task(models.Model):
         self.action_assign()
 
     @api.model
-    def _prepare_procurement_group_vals(self, values):
-        return {"name": values["name"]}
+    def _prepare_procurement_group_vals(self):
+        return {"name": "Task-ID: %s" % self.id}
 
     def action_confirm(self):
         self.mapped("move_ids")._action_confirm()
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -23,6 +23,12 @@ class Task(models.Model):
     use_stock_moves = fields.Boolean(related="stage_id.use_stock_moves")
     done_stock_moves = fields.Boolean(related="stage_id.done_stock_moves")
     stock_moves_is_locked = fields.Boolean(default=True)
+    allow_moves_action_confirm = fields.Boolean(
+        compute="_compute_allow_moves_action_confirm"
+    )
+    allow_moves_action_assign = fields.Boolean(
+        compute="_compute_allow_moves_action_assign"
+    )
     stock_state = fields.Selection(
         selection=[
             ("pending", "Pending"),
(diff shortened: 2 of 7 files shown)


pedrobaeza:
Why avoid to merge? I prefer to have them merged. And you want to ping @chienandalu

victoralmau:
> Why avoid to merge? I prefer to have them merged. And you want to ping @chienandalu

I think a use case will explain the problem:

- User add a stock move in a task (product A, quantity 10).
- Confirm material + check availability (using the buttons)
- Stock move has been removed.

Odoo core (and OCA addons)  uses it in some addons, for example: https://github.com/OCA/OCB/blob/13.0/addons/sale_stock/models/stock.py#L19-L23

chienandalu (approved):

pedrobaeza:
Why is it removed? It should be merged with another, but the quantity is put in that one

victoralmau:
Changes done to add `procurement.group` in tasks.

project_stock/models/project_task.py
```diff
@@ -82,6 +89,21 @@ def _compute_scrap_move_count(self):
...
+    def _compute_allow_moves_action_confirm(self):
+        for item in self:
+            item.allow_moves_action_confirm = bool(
+                any(move.state == "draft" for move in item.move_ids)
+            )
+
+    def _compute_allow_moves_action_assign(self):
+        for item in self:
+            item.allow_moves_action_assign = bool(
+                any(
```
pedrobaeza: `any` already returns a bool, so no need to cast.

project_stock/tests/common.py
```diff
@@ -88,6 +88,8 @@ def _create_task(self, products):
             self.env["project.task"].with_context(self._prepare_context_task(self))
         )
         task_form.name = "Test task"
+        # Save task to use default_get() correctlly in stock.moves
+        task_form.save()
```
pedrobaeza, line 92: I don't get why you need this.

victoralmau: Save is necessary to set `default_raw_material_task_id` (https://github.com/OCA/project/blob/13.0/project_stock/views/project_task_view.xml#L110) correctly and auto-create `procurement.group` record.

pedrobaeza: That sound weird, and more if no stock.move is still created. Isn't this a bug on `Form`?

victoralmau: Yes, it seems more like a bug in `Form()` but I think it's necessary for tests.

pedrobaeza:
Is this compatible with existing data?

victoralmau:
> Is this compatible with existing data?

IMO it's not a problem with existing data, although creating a migration script to set `procurement.group` in tasks and assign them to the stock moves would be much better.

pedrobaeza:
I think we can include it, but only for existing task moves.

project_stock/migrations/13.0.1.2.0/post-migration.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2022 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    env.cr.execute(
```
pedrobaeza: There's no need of SQL here. You can do a simply search over tasks with the o2m field.

project_stock/models/project_task.py
```diff
@@ -132,8 +152,10 @@ def _update_moves_info(self):
         self.action_assign()
 
     @api.model
-    def _prepare_procurement_group_vals(self, values):
-        return {"name": values["name"]}
+    def _prepare_procurement_group_vals(self):
+        """We use id in the name to avoid duplicate names if multiple tasks
+        have the same name."""
+        return {"name": "%s-%s" % (self.id, self.name)}
```
pedrobaeza: The procurement group name should be short, but also unique. I recommend:

```suggestion
        return {"name": "Task-ID: %s" % self.id}
```

project_stock/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 {
     "name": "Project Stock",
-    "version": "13.0.1.1.2",
+    "version": "13.0.1.2.0",
```
pedrobaeza: Change this also to 13.0.2.0.0, as it's a big change

pedrobaeza (approved):

PR #947, opened 2022-07-19
[14.0][UPD]project_milestone: new features added

AshishHirapara:
* New features added
    #. Subtask inherits the milestones from the parent task
    #. 'Milestone Required' field is added to project
    #. Milestone can be assigned directly from Task Kanban

### DIFF
--- a/project_milestone/models/project.py
+++ b/project_milestone/models/project.py
@@ -1,6 +1,6 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class Project(models.Model):
@@ -10,3 +10,10 @@ class Project(models.Model):
         "project.milestone", "project_id", string="Milestones", copy=True
     )
     use_milestones = fields.Boolean(help="Does this project use milestones?")
+
+    milestones_required = fields.Boolean()
+
+    @api.onchange("use_milestones")
+    def _onchange_use_milestones(self):
+        if not self.use_milestones and self.milestones_required:
+            self.milestones_required = False
--- a/project_milestone/models/project_task.py
+++ b/project_milestone/models/project_task.py
@@ -15,6 +15,9 @@ class ProjectTask(models.Model):
     use_milestones = fields.Boolean(
         related="project_id.use_milestones", help="Does this project use milestones?"
     )
+    milestones_required = fields.Boolean(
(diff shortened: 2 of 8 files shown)


francesco-ooops (approved):
Functional review ok!

elvise (approved):
LGTM!

project_milestone/models/project.py
```diff
@@ -10,3 +10,10 @@ class Project(models.Model):
         "project.milestone", "project_id", string="Milestones", copy=True
     )
     use_milestones = fields.Boolean(help="Does this project use milestones?")
+
+    milestones_required = fields.Boolean()
+
+    @api.onchange("use_milestones")
+    def use_milestones_onchange(self):
```
yajo: Onchange methods must be private, and unless there's a major reason, follow the standard of prefixing it with `_onchange_`. So it'd be something like:

```suggestion
    def _onchange_use_milestones(self):
```

project_milestone/models/project_task.py
```diff
@@ -23,3 +26,25 @@ def _read_group_milestone_ids(self, milestone_ids, domain, order):
...
+                    {
+                        "milestone_id": parent_task.milestone_id.id,
+                    }
+                )
+        res = super(ProjectTask, self).create(vals)
+
+        return res
+
+    @api.onchange("parent_id")
+    def parent_id_milestone_onchange(self):
```
yajo: Just like before:

```suggestion
    def _onchange_parent_id_milestone(self):
```

yajo (changes requested):
Quoting from [the docs](https://www.odoo.com/documentation/14.0/developer/howtos/rdtraining/09_compute_onchange.html):

> In many cases, both computed fields and onchanges may be used to achieve the same result. Always prefer computed fields since they are also triggered outside of the context of a form view.

Read those docs to understand the differences. I think that in this particular case, the computed writable field is a better option.

If you still prefer to keep going with onchanges, then at least please attend the 2 comments below.

Thanks!

project_milestone/models/project.py
AshishHirapara: @yajo I will go with changing the method names. Thanks for the suggestions. These are done.

project_milestone/models/project_task.py
AshishHirapara: done

francesco-ooops:
@yajo good now?

yajo (approved):

PR #951, opened 2022-07-26
[15.0][FIX] project_milestone: fixed migration scripts. 

Mantux11:
In migrate function there was cr left, changed that. Also in rename_fields was left cr, where was needed an env.


leemannd (approved):
I didn't use it. But it seems legit

gurneyalex (approved):

PR #954, opened 2022-08-05
[15.0][MIG] project leverage

Hardik-OSI:


project_leverage/README.rst
```diff
@@ -0,0 +1,28 @@
...
+============
+
+* No specific installation step required.
+
+Usage
+=====
+
+* Go to Project > Reporting > Timesheets
+* Select the measure Duration
+* Group by Project Manage
```
gurneyalex:
```suggestion
* Group by Project Manager
```

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,101 @@
...
+        month_list = []
+        month_name_list = []
+        for rec in raw_data:
+            month_name_list.append(rec.get("month_year"))
+            if (
+                str(rec.get("date_mm")).strip() + str(int(rec.get("year")))
+                not in month_list
+            ):
+                month_list.append(
+                    str(rec.get("date_mm")).strip() + str(int(rec.get("year")))
```
gurneyalex, line 32: would be so nice to have this string computed only once. Why do we call str()? concatenating 2 strings will return a string. 

You will get a crash if date_mm is not in rec (because rec.get() will return None and the call to strip() will fail. Same if "year" is not in rec. If you are sure the key is present, then don't use get, use a normal key access, otherwise use a default value which will not crash strip() or int() respectively.

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,101 @@
...
+                )
+            if project_data.get(rec.get("employee_id") or 0):
```
gurneyalex:
```suggestion
            if project_data.get(rec.get("employee_id", 0)):
```

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,101 @@
...
+                project_data[rec.get("employee_id") or 0].update(
```
gurneyalex:
```suggestion
                project_data[rec.get("employee_id", 0)].update(
```

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,101 @@
...
+                    {
+                        str(rec.get("date_mm")).strip()
+                        + str(int(rec.get("year"))): {
```
gurneyalex, line 38: again this concatenation -> please use an intermediate variable with a meaningful name.

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,101 @@
...
+                        }
+                    }
+                )
+            else:
+                project_data.update(
+                    {
+                        rec.get("employee_id")
+                        or 0: {
+                            str(rec.get("date_mm")).strip()
+                            + str(int(rec.get("year"))): {
```
gurneyalex, line 53: and again... please use an intermediate variable

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,101 @@
...
+                                "month_name": rec.get("month"),
+                                "hours": rec.get("sum"),
+                            },
+                            "name": rec.get("employee_id")
+                            and employee_obj.browse(int(rec.get("employee_id"))).name
+                            or "undefine",
```
gurneyalex:
```suggestion
                            "name": employee_obj.browse(int(rec["employee_id"])).name if 'employee_id' in rec else _("undefined"),
                        
```

project_leverage/models/account_analytic_line.py
```diff
@@ -0,0 +1,37 @@
...
+    )
+
+    @api.depends("employee_id", "project_user_id")
+    def _compute_is_manager(self):
+        for rec in self:
+            if rec.employee_id.user_id == rec.project_user_id:
+                rec.is_manager = True
+
+    @api.model
+    def create(self, vals):
```
gurneyalex, line 21: please update this method to use the model_create_multi decorator.

gurneyalex (changes requested):

project_leverage/controllers/main.py
```diff
@@ -0,0 +1,101 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo.http import request
+
+from odoo.addons.sale_timesheet.controllers.portal import SaleTimesheetCustomerPortal
+
+DEFAULT_MONTH_RANGE = 3
+
+
+class SaleTimesheetController(SaleTimesheetCustomerPortal):
+    def _plan_prepare_values(self, projects):
```
hailangvn, line 12: The project overview, which had used this method in 12.0, has been removed since 15.0 by [commit 2f77c862756](https://github.com/odoo/odoo/commit/2f77c862756). So this method should be refactored to another method. I am wondering what is the purpose of this method.

project_leverage/models/account_analytic_line.py
```diff
@@ -0,0 +1,37 @@
...
+        if "task_id" in vals:
+            Task = self.env["project.task"]
+            task_id = vals.get("task_id")
+            task = Task.browse(task_id) if task_id else Task
+            project_user_id = task.project_id.user_id and task.project_id.user_id.id
```
hailangvn: For better understanding and aligning with description, we can leverage existing [manager_id field](https://github.com/odoo/odoo/blob/b0e960bbe142f6d00b444f78fe11e63b83cecb29/addons/project/models/project.py#L1029).
```suggestion
            project_user_id = task.manager_id and task.manager_id.id
```

project_leverage/models/account_analytic_line.py
```diff
@@ -0,0 +1,37 @@
...
+            vals.update({"project_user_id": project_user_id or None})
+        return super(AccountAnalyticLine, self).create(vals)
+
+    def write(self, vals):
+        if "task_id" in vals:
+            Task = self.env["project.task"]
+            task_id = vals.get("task_id")
+            task = Task.browse(task_id) if task_id else Task
+            project_user_id = task.project_id.user_id and task.project_id.user_id.id
```
hailangvn: Same here.
```suggestion
            project_user_id = task.manager_id and task.manager_id.id
```

hailangvn (changes requested):
Thanks for the work. Please consider to update the module. Thanks.

project_leverage/models/account_analytic_line.py
hailangvn: It is better to move duplicated code from `create` and `write` methods into a method (`_set_manager_id` or something similar).

PR #956, opened 2022-08-08
[14.0][FIX] project_mail_chatter: We need to replace oe_chatter div to place it after the sheet and show as expected.

victoralmau:
We need to replace `oe_chatter` div to place it after the sheet and show as expected.

Before
project_mail_chatter-antes

After
project_mail_chatter-despues

@Tecnativa TT38498


chienandalu (approved):
Yep. Can you add to the roadmap that this module isn't needed in v15?

pedrobaeza (approved):

PR #961, opened 2022-08-15
[14.0][IMP] project_task_dependency: add open dep. task button

ilyasProgrammer:


astirpe (approved):

pedrobaeza:
As this module is deprecated in v15, not sure if adding too much things that won't exist in next version in core is a good idea.

elvise:
> As this module is deprecated in v15, not sure if adding too much things that won't exist in next version in core is a good idea.

Yup, just let's add the smart button "Blocking" and it's done, that's ok ?

elvise:
@pedrobaeza merge ?   :)

elvise (approved):
Functional: Perfect!

pedrobaeza:
Is these features available in core in 15.0? My fear is that transitioning from this enhanced version to core v15 option will provoke conflict ripping some features, and in that case, better to extract them apart for keeping them across versions.

elvise:
> Is these features available in core in 15.0? My fear is that transitioning from this enhanced version to core v15 option will provoke conflict ripping some features, and in that case, better to extract them apart for keeping them across versions.

@pedrobaeza Sure, we have implement only the same features

pedrobaeza (commented):
I don't see the smart-button of blocking count in v15, and please squash commits.

elvise:
> I don't see the smart-button of blocking count in v15, and please squash commits.

@pedrobaeza It appear when there is one or more dependencies

image

pedrobaeza:
What I see in v15 is `dependent_tasks_count` field and smart-button, and it seems to not be the same. Why don't you copy/paste such code for maximizing also the compatibility in extensions?

ilyasProgrammer:
> What I see in v15 is `dependent_tasks_count` field and smart-button, and it seems to not be the same. Why don't you copy/paste such code for maximizing also the compatibility in extensions?

_compute_dependent_tasks_count in v15 is different. And there is used different fields.
I renamed field to dependent_tasks_count and method to _compute_dependent_tasks.

pedrobaeza:
Yes, but then we go again to my comment in https://github.com/OCA/project/pull/961#issuecomment-1219566966 that I don't want something different in one version from the other, or the transition will be bad.

elvise:
> Yes, but then we go again to my comment in [#961 (comment)](https://github.com/OCA/project/pull/961#issuecomment-1219566966) that I don't want something different in one version from the other, or the transition will be bad.

@ilyasProgrammer can you check it ?

ilyasProgrammer:
@pedrobaeza But fields names already different. Field i've added has same name as in V15. I dont understand what kind of transition problems it can cause.

pedrobaeza:
A "user transition". They get used to some features on v14, and when moving to v15, they don't have a feature, but the contrary one, causing a lot of troubles. That's why I want "feature parity" between v14 OCA module and v15 core module, or if not, less features in v14 than in v15.

pedrobaeza (changes requested):
See my previous remark

francesco-ooops:
> 

Hi @pedrobaeza, I don't have very clear how to proceed with this PR. 

Do you think we should just do a backport of v.15 core features, regardless of the features of project_task_dependency?

If yes, should we push a new module in this repo? How should we name it? Thanks for your help, I don't have much experience with backportings.

pedrobaeza:
Yes, I recommend to backport v15 features in this module if you want to have them, and put any other improvement in an extra layer.

francesco-ooops:
@pedrobaeza I'm not sure it would make much sense to backport features into project_task_dependency module, as the features are almost the same (the differences being a tab with a different name and the smart buttons we're trying to backport). 

If that's ok we can just backport everything from v15 into another module and then users will decide whether to use project_task_dependency or the new module with the same features as core v15

this means in repo v14 there will be two modules with almost the same features, is that ok for you?

pedrobaeza:
But the problem is that current users of this module will transition to v15 core module, and adding features that are not going to be in v15 will be a bad migration path for them. I don't want an additional module with v15 features. What I don't want is that this module gets bigger from what it is, and on v15 forces us to maintain an additional module apart from core for not having them to lose some of the features that were added in v14.

francesco-ooops:
Ok, so we can't get this module bigger and we can't create a new one which
will be discarded after v15 (no maintenance needed I think?), what can we
do? 🥲

pedrobaeza:
But why the v15 feature doesn't fit you? It's the same feature, but reversed.

francesco-ooops:
v15 is great! Our goal is to adopt those features in v14

But what should we do to have those features in v14? This is what's unclear to me

pedrobaeza:
What I'm proposing is to adopt the same field name + view changes as of v15.

francesco-ooops:
ok, so we implement depend_on_ids and the view changes, but what do we do with the existing dependency_task_ids ?

pedrobaeza:
I was talking more about `dependent_tasks_count`, which is the missing feature here. `dependency_task_ids` can be kept and adapt the other to this one, and migrate on OpenUpgrade, or you totally switch to `depend_on_ids` + `dependent_tasks_count` here, and include already migration script for this adaptation.

ilyasProgrammer:
@pedrobaeza **dependent_tasks_count** is added in this module with the same name as in V15 and with the same compute method name. I dont understand why is it wrong. Could you explain please.

pedrobaeza (approved):
Wow, I completely missed this change: https://github.com/OCA/project/compare/619f35d92354531f5d3dd158f3207c50087b6119..bbbafffa5894cd1baf59103eec19d770d02dcc00 where you attended my remarks. Sorry for that missing.

PR #964, opened 2022-08-18
[IMP][12.0] project_task_material_stock: improve code for multi compa…

cubells:
…ny environment.

### DIFF
--- a/project_task_material_stock/models/project_task.py
+++ b/project_task_material_stock/models/project_task.py
@@ -88,6 +88,11 @@ class Task(models.Model):
         help='Keep this field empty to use the default value from'
         ' the project.'
     )
+    picking_type_id = fields.Many2one(
+        comodel_name="stock.picking.type",
+        string="Operation Type",
+        index=True,
+    )
 
     @api.multi
     def unlink_stock_move(self):
@@ -189,9 +194,9 @@ class ProjectTaskMaterial(models.Model):
 
     @api.multi
     def create_stock_move(self):
-        pick_type = self.env.ref(
-            'project_task_material_stock.project_task_material_picking_type')
         task = self[0].task_id
+        pick_type = task.picking_type_id or self.env.ref(
+            'project_task_material_stock.project_task_material_picking_type')
         picking_id = task.picking_id or self.env['stock.picking'].create({
             'origin': "{}/{}".format(task.project_id.name, task.name),
             'partner_id': task.partner_id.id,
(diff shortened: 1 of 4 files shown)


rlizana (approved):
LGTM :+1:

antonio-trey:
👍🏻

antonio-trey (approved):
👍🏻

pedrobaeza (commented):
I know that this is not your fault, but there should be a picking type per company, and this pull request is not improving that situation, just adding hooks. Can you solve the underlying problem?

cubells:
Yes, what's your idea in order to solve this @pedrobaeza ?

pedrobaeza:
You can hook on the company creation the specific creation of a picking type for each company, or have a field in the company to select which picking type to use.

cubells:
@pedrobaeza I think It's better do it at warehouse level instead of company level. Other modules do it in this manner. 

image

What do you think?

pedrobaeza:
But then you will need to specify the warehouse at project/task level. Going to that, it's preferable to configure directly the picking type at project/task instead. We do this in our new refactor `project_stock` module for v13.

cubells:
**But then you will need to specify the warehouse at project/task level.**

I must simply search for the first warehouse of current company and get the material picking type.

**Going to that, it's preferable to configure directly the picking type at project/task instead. We do this in our new refactor project_stock module for v13.**

Ok, I'm going to do it in this manner.

project_task_material_stock/models/project_task.py
```diff
@@ -189,15 +194,13 @@ def _prepare_stock_move(self):
 
     @api.multi
     def create_stock_move(self):
-        pick_type = self.env.ref(
```
pedrobaeza, line 192: Maybe you should leave this picking type as fallback.

cubells: The problem here is that in multi company environment, picking type is assigned even if company haven't a picking type because model stock.picking.type hasn't company_id field in version 12.0, so we have a picking type with locations of another company.

cubells: maybe we have to do some check in warehouse to avoid this problem.

pedrobaeza: Yeah, but you have now the option to avoid it setting explicitly the picking type here.

cubells: ok, I understood.

PR #965, opened 2022-08-22
[12.0][IMP] project_task_material_stock: add domain to picking type

cubells:
This changes avoid user to choose a picking type that not consume material.

### DIFF
--- a/project_task_material_stock/models/project_task.py
+++ b/project_task_material_stock/models/project_task.py
@@ -90,6 +90,7 @@ class Task(models.Model):
     )
     picking_type_id = fields.Many2one(
         comodel_name="stock.picking.type",
+        domain=[("code", "in", ("outgoing", "internal"))],
         string="Operation Type",
         index=True,
     )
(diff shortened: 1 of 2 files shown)


rlizana (approved):
LGTM

project_task_material_stock/models/project_task.py
```diff
@@ -90,6 +90,7 @@ def _compute_stock_state(self):
     )
     picking_type_id = fields.Many2one(
         comodel_name="stock.picking.type",
+        domain=[("code", "=", "outgoing")],
```
pedrobaeza:
```suggestion
        domain=[("code", "in", ("outgoing", "internal"))],
```

cubells: If I add internal type, material is not consumed, isn't It?

pedrobaeza: It's consumed from source location and move to destination location, so I think it can be correct.

cubells: @pedrobaeza done!

cubells:
thanks @pedrobaeza

Issue #966, opened 2022-08-22
Improvement of Unique Constrains on Project Key 


yweng8111:
**Module:**
project_key

**Problem and Solutions:**
I want to install the module project_key in a DB which contains some Projects with the same Name but different company_id. The Hook script in this module tries to generate the key for every project. But the function generate_project_key of project doesn't consider the company_id and show me the error of the unique constrains .

Can we just change the constrains like this?
```
    _sql_constraints = [
        ("project_key_unique", "UNIQUE(key, company_id)", "Project key must be unique")
    ]
```
If you think that the key must be unique, We can also improve the function generate_project_key with an extra parameter "company_id" and return the new key (old key + company_id).

What do you think?

marylla:
I think the current way to create project keys is very error-prone. 
We have a lot of projects that are very similiar and therefore have a similiar naming scheme.
We can not install the module after years of using Odoo, due to this hook and unique contraint.

At least we should consider the company in this unique-contraint as @yweng8111 suggested, but I think the hook should be totally different. What about a hash value or maybe just an ...easy numbering?

@pedrobaeza @dreispt What do you think? Do you see a better solution?

pedrobaeza:
I agree. I don't like this module at all. The uniqueness should be checked only for manual assigned keys. For automatic ones, there should be a loop until one key is free, using for example a counter.

yweng8111:
@pedrobaeza Thanks. I will try to improve the function like you suggested.

PR #967, opened 2022-08-25
[14.0] [IMP] add function to generate unique project key automatically.

yweng8111:
As we discussed in #966 , I have implemented a small extension for automatic generating of the unique project key.

### DIFF
--- a/project_key/models/project_project.py
+++ b/project_key/models/project_project.py
@@ -138,12 +138,24 @@ class Project(models.Model):
 
         data = text.split(" ")
         if len(data) == 1:
-            return data[0][:3].upper()
+            return self._generate_project_unique_key(data[0][:3].upper())
 
         key = []
         for item in data:
             key.append(item[:1].upper())
-        return "".join(key)
+        return self._generate_project_unique_key("".join(key))
+
+    def _generate_project_unique_key(self, text):
+        res = text
+        unique_key = False
+        counter = 0
+        while not unique_key:
+            if counter != 0:
+                res = "%s%s" % (text, counter)
+            unique_key = not bool(self.search([("key", "=", res)]))
+            counter += 1
+
+        return res
 
     def _update_task_keys(self):
         """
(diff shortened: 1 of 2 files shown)


marylla (approved):
I made a functional review.

I created 3 projects with similiar naming scheme, and tested it in 2 different runboat instances **with** and **without** this improvement and checked the key suggestions.

| Project Name              | Without Improvement | With Improvement |
| ---| ---| ---|
| "Internal Event"           | IE | IE |
| "International Event"   | IE | IE1 |
| "Intercompany Event" | IE | IE2 |

So, I guess, the problem of the unique contraint and the hook that fixes old projects (before installation of module project key) is fixed this way.

Thank you! :+1:

chandni299 (approved):
It looks good

project_key/models/project_project.py
```diff
@@ -138,12 +138,29 @@ def generate_project_key(self, text):
...
+        counter = 0
+        while not unique_key:
+            if counter != 0:
+                res = "%s%s" % (text, counter)
+            projects = self.search([("key", "=", res)])
+            if not projects:
+                unique_key = True
+            else:
+                counter += 1
+
```
pedrobaeza:
```suggestion
            unique_key = bool(self.search([("key", "=", res)]))
            counter += 1
```

project_key/models/project_project.py
```diff
@@ -138,12 +138,29 @@ def generate_project_key(self, text):
 
         data = text.split(" ")
         if len(data) == 1:
-            return data[0][:3].upper()
+            res = data[0][:3].upper()
+            return self._generate_project_unique_key(res)
```
pedrobaeza:
```suggestion
            return self._generate_project_unique_key(data[0][:3].upper())
```

pedrobaeza (commented):
Make code shorter

project_key/models/project_project.py
yweng8111: Thanks for the review.
I've changed it.

project_key/models/project_project.py
yweng8111: Done

marylla:
@pedrobaeza This was very fast. Awesome. :) Could you merge?

PR #968, opened 2022-08-30
[13.0][FIX] project_stock: Avoid task deletion error.

victoralmau:
We need to apply sudo to avoid error if the user does not have access to analytical accounting.

@Tecnativa TT38702

### DIFF
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -192,8 +192,12 @@ class Task(models.Model):
         return True
 
     def action_cancel(self):
+        """Cancel the stock moves and remove the analytic lines created from
+        stock moves when cancelling the task.
+        """
         self.move_ids.write({"state": "cancel"})
-        self.stock_analytic_line_ids.unlink()
+        # Use sudo to avoid error for users with no access to analytic
+        self.sudo().stock_analytic_line_ids.unlink()
         self.stock_moves_is_locked = True
         return True
 
@@ -206,7 +210,8 @@ class Task(models.Model):
         for move in self.mapped("move_ids"):
             move.quantity_done = move.reserved_availability
         self.mapped("move_ids")._action_done()
-        analytic_line_model = self.env["account.analytic.line"]
+        # Use sudo to avoid error for users with no access to analytic
+        analytic_line_model = self.env["account.analytic.line"].sudo()
         for move in self.move_ids.filtered(lambda x: x.state == "done"):
             vals = move._prepare_analytic_line_from_task()
             if vals:
@@ -233,7 +238,8 @@ class Task(models.Model):
         return res
 
     def unlink(self):
-        self.mapped("stock_analytic_line_ids").unlink()
+        # Use sudo to avoid error to users with no access to analytic
+        self.sudo().mapped("stock_analytic_line_ids").unlink()
         return super().unlink()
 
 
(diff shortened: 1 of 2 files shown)


chienandalu (approved):

project_stock/models/project_task.py
```diff
@@ -192,8 +192,10 @@ def button_unreserve(self):
         return True
 
     def action_cancel(self):
+        """We need to apply sudo to avoid error if the user does not have access to
```
pedrobaeza: This is not a docstring. The docstring is for commenting the goal of the method. In this case, it would be: `Remove analytic lines created with stock moves when cancelling the task`. This comment should go before the `self.sudo()...` line.

project_stock/models/project_task.py
```diff
@@ -203,10 +205,12 @@ def action_toggle_stock_moves_is_locked(self):
         return True
 
     def action_done(self):
+        """We need to apply sudo to avoid error if the user does not have access to
```
pedrobaeza: The same for this.

pedrobaeza (changes requested):

project_stock/models/project_task.py
victoralmau: The corresponding comment has already been applied in the correct place.

project_stock/models/project_task.py
```diff
@@ -193,7 +193,8 @@ def button_unreserve(self):
 
     def action_cancel(self):
         self.move_ids.write({"state": "cancel"})
-        self.stock_analytic_line_ids.unlink()
+        # Remove analytic lines created with stock moves when cancelling the task
```
pedrobaeza: This is not what I talked about. This text should go as docstring, and the other that you put before as comment in the same way as now you have put this one.

pedrobaeza: You can refine the docstring putting as well that moves are cancelled.

project_stock/models/project_task.py
```diff
@@ -193,7 +193,8 @@ def button_unreserve(self):
 
     def action_cancel(self):
         self.move_ids.write({"state": "cancel"})
```
pedrobaeza, line 198:
```suggestion
        """Cancel the stock moves and remove the analytic lines created from
        stock moves when cancelling the task.
        """
        self.move_ids.write({"state": "cancel"})
```

project_stock/models/project_task.py
```diff
@@ -193,7 +193,8 @@ def button_unreserve(self):
...
-        self.stock_analytic_line_ids.unlink()
+        # Remove analytic lines created with stock moves when cancelling the task
+        self.sudo().stock_analytic_line_ids.unlink()
```
pedrobaeza, line 200:
```suggestion
        # Use sudo to avoid error to users with no access to analytic
        self.sudo().stock_analytic_line_ids.unlink()
```

project_stock/models/project_task.py
```diff
@@ -233,7 +238,9 @@ def write(self, vals):
         return res
 
     def unlink(self):
-        self.mapped("stock_analytic_line_ids").unlink()
+        """We need to apply sudo to avoid error if the user does not have access to
```
pedrobaeza: This mustn't be a docstring...

pedrobaeza (approved):

PR #971, opened 2022-08-30
[15.0][FIX] project_forecast_line: take roles into account on calc and add setting to control consumption states

yankinmax:
This PR introduces two fixes:
1. Capacity Lines calculation does not take into accounts different roles. We need to adapt `_compute_employee_forecast_line_id` method
2. Add a setting to take control on states for which the consumption is confirmed.

### DIFF
--- a/project_forecast_line/models/forecast_line.py
+++ b/project_forecast_line/models/forecast_line.py
@@ -88,12 +88,15 @@ class ForecastLine(models.Model):
     @api.depends("employee_id", "date_from", "type", "res_model")
     def _compute_employee_forecast_line_id(self):
         employees = self.mapped("employee_id")
+        main_roles = employees.mapped("main_role_id")
         date_froms = self.mapped("date_from")
         date_tos = self.mapped("date_to")
+        forecast_roles = self.mapped("forecast_role_id") | main_roles
         if employees:
             lines = self.search(
                 [
                     ("employee_id", "in", employees.ids),
+                    ("forecast_role_id", "in", forecast_roles.ids),
                     ("res_model", "=", "hr.employee.forecast.role"),
                     ("date_from", ">=", min(date_froms)),
                     ("date_to", "<=", max(date_tos)),
@@ -104,12 +107,23 @@ class ForecastLine(models.Model):
             lines = self.env["forecast.line"]
         capacities = {}
         for line in lines:
-            capacities[(line.employee_id.id, line.date_from)] = line.id
+            capacities[
+                (line.employee_id.id, line.date_from, line.forecast_role_id.id)
+            ] = line.id
         for rec in self:
             if rec.type == "confirmed" and rec.res_model != "hr.employee.forecast.role":
-                rec.employee_resource_forecast_line_id = capacities.get(
-                    (rec.employee_id.id, rec.date_from), False
+                resource_forecast_line = capacities.get(
+                    (rec.employee_id.id, rec.date_from, rec.forecast_role_id.id), False
                 )
+                if resource_forecast_line:
+                    rec.employee_resource_forecast_line_id = resource_forecast_line
+                else:
+                    # if we didn't find a forecast line with a matching role
+                    # we get forecast line with the main role of the employee
+                    main_role_id = rec.employee_id.main_role_id
+                    rec.employee_resource_forecast_line_id = capacities.get(
+                        (rec.employee_id.id, rec.date_from, main_role_id.id), False
+                    )
             else:
                 rec.employee_resource_forecast_line_id = False
 
--- a/project_forecast_line/models/forecast_line.py
+++ b/project_forecast_line/models/forecast_line.py
@@ -85,8 +85,13 @@ class ForecastLine(models.Model):
         "forecast.line", "employee_resource_forecast_line_id"
     )
 
+    def _get_consumption_states(self):
+        consumption_states = self.env.company.forecast_consumption_states
+        return tuple(consumption_states.split("_"))
+
     @api.depends("employee_id", "date_from", "type", "res_model")
     def _compute_employee_forecast_line_id(self):
+        consumption_states = self._get_consumption_states()
         employees = self.mapped("employee_id")
         main_roles = employees.mapped("main_role_id")
         date_froms = self.mapped("date_from")
@@ -111,7 +116,10 @@ class ForecastLine(models.Model):
                 (line.employee_id.id, line.date_from, line.forecast_role_id.id)
             ] = line.id
         for rec in self:
-            if rec.type == "confirmed" and rec.res_model != "hr.employee.forecast.role":
+            if (
+                rec.type in consumption_states
+                and rec.res_model != "hr.employee.forecast.role"
+            ):
                 resource_forecast_line = capacities.get(
                     (rec.employee_id.id, rec.date_from, rec.forecast_role_id.id), False
                 )
--- a/project_forecast_line/models/project_task.py
+++ b/project_forecast_line/models/project_task.py
@@ -18,9 +18,9 @@ class ProjectTask(models.Model):
     def create(self, vals_list):
         # compatibility with fields from project_enterprise
         for vals in vals_list:
-            if "planned_date_begin" in vals:
+            if vals.get("planned_date_begin"):
                 vals["forecast_date_planned_start"] = vals["planned_date_begin"]
-            if "planned_date_end" in vals:
+            if vals.get("planned_date_end"):
                 vals["forecast_date_planned_end"] = vals["planned_date_end"]
         tasks = super().create(vals_list)
         tasks._update_forecast_lines()
@@ -94,6 +94,7 @@ class ProjectTask(models.Model):
                     # are not generating forecast lines from SO
                     _logger.info("skip task %s: draft sale")
                     continue
+
             if (
                 not task.forecast_date_planned_start
                 or not task.forecast_date_planned_end
--- a/project_forecast_line/tests/test_forecast_line.py
+++ b/project_forecast_line/tests/test_forecast_line.py
@@ -517,7 +517,7 @@ class TestForecastLineProject(BaseForecastLineTest):
             self.assertEqual(len(forecast), 1)
             # using assertEqual on purpose here
             self.assertEqual(forecast.forecast_hours, -6.0)
-            self.assertEqual(round(forecast.consolidated_forecast, 5), 0.75000)
+            self.assertAlmostEqual(forecast.consolidated_forecast, 0.75)
             self.assertEqual(
                 forecast.employee_resource_forecast_line_id.consolidated_forecast,
                 0.25,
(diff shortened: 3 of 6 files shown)


project_forecast_line/tests/test_forecast_line.py
```diff
@@ -610,3 +610,127 @@ def test_task_forecast_lines_consolidated_forecast_overallocation_multiple_tasks
...
+                [("employee_id", "=", self.employee_consultant.id)]
+            )
+            # we can take first line to check as forecast values are equal
+            employee_forecast_consultant = employee_forecast.filtered(
+                lambda l: l.res_model == "hr.employee.forecast.role"
+                and l.forecast_role_id == self.role_consultant
+            )[0]
+            self.assertEqual(employee_forecast_consultant.forecast_hours, 6.0)
+            self.assertEqual(
+                round(employee_forecast_consultant.consolidated_forecast, 5), 0.75000
```
yankinmax: @gurneyalex don't we expect here to have a negative value for the `consultant` role?

yankinmax: Though in UI it looks ok

UI

project_forecast_line/tests/test_forecast_line.py
```diff
@@ -610,3 +610,127 @@ def test_task_forecast_lines_consolidated_forecast_overallocation_multiple_tasks
...
+            project = self.env["project.project"].create(
+                {"name": "TestProjectDiffRoles"}
+            )
+            # set project in stage "in progress" to get confirmed forecast
+            project.stage_id = self.env.ref("project.project_project_stage_1")
+            task = self.env["project.task"].create(
+                {
+                    "name": "TaskDiffRoles",
+                    "project_id": project.id,
+                    "forecast_role_id": self.role_developer.id,
```
yankinmax: specify another role for forecast to be calculated using `main_role_id`

project_forecast_line/tests/test_forecast_line.py
```diff
@@ -610,3 +610,127 @@ def test_task_forecast_lines_consolidated_forecast_overallocation_multiple_tasks
...
+            employee_forecast = self.env["forecast.line"].search(
+                [("employee_id", "=", self.employee_consultant.id)]
+            )
+            # we can take first line to check as forecast values are equal
+            employee_forecast_consultant = employee_forecast.filtered(
+                lambda l: l.res_model == "hr.employee.forecast.role"
+                and l.forecast_role_id == self.role_consultant
+            )[0]
+            self.assertEqual(employee_forecast_consultant.forecast_hours, 6.0)
+            self.assertEqual(
```
gurneyalex: you should use self.assertAlmostEqual for float comparisons

project_forecast_line/tests/test_forecast_line.py
```diff
@@ -610,3 +610,127 @@ def test_task_forecast_lines_consolidated_forecast_overallocation_multiple_tasks
                 ),
                 -0.75000,
             )
+
+    def test_task_forecast_lines_employee_different_roles(self):
```
gurneyalex, line 611: can you add a docstring to explaining what you are testing here?

project_forecast_line/tests/test_forecast_line.py
```diff
@@ -610,3 +610,127 @@ def test_task_forecast_lines_consolidated_forecast_overallocation_multiple_tasks
...
+            employee_forecast_pm = employee_forecast.filtered(
+                lambda l: l.res_model == "hr.employee.forecast.role"
+                and l.forecast_role_id == self.role_pm
+            )[0]
+            self.assertEqual(employee_forecast_pm.forecast_hours, 2.0)
+            self.assertEqual(
+                round(employee_forecast_pm.consolidated_forecast, 5), 0.25000
+            )
+
+    def test_task_forecast_lines_employee_main_role(self):
```
gurneyalex, line 681: can you add a docstring to explaining what you are testing here?

project_forecast_line/tests/test_forecast_line.py
```diff
@@ -603,10 +603,150 @@ def test_task_forecast_lines_consolidated_forecast_overallocation_multiple_tasks
...
+        """
+        Test forecast lines when employee has different roles.
+
+        Employee has 2 forecast_role_id: consultant 75% and project manager 25%,
+        working 8h per day (standard calendar).
+        Create a task with forecast role consultant, with remaining time = 8h
+        and a scheduled period starting and ending on the same day (22/01/2022 for instance).
+        Assign this task to the user.
+
+        Expected: for the user, on 22/01/2022, 3 forecast lines.
```
gurneyalex: what matters are the values in these lines :)

gurneyalex (approved):

project_forecast_line/tests/test_forecast_line.py
yankinmax: I updated with values, thanks for the help with this fix!

gurneyalex (changes requested):
@yankinmax you need to change these lines to adapt the dict cache in these lines to accomodate for the role https://github.com/OCA/project/pull/971/files#diff-9c22705718731bd0bfc65398ad332d8562675e0a6592b660691f6ff9aa5907afR115-R122 otherwise the matching will not work properly

project_forecast_line/models/project_task.py
```diff
@@ -18,9 +18,9 @@ class ProjectTask(models.Model):
     def create(self, vals_list):
         # compatibility with fields from project_enterprise
         for vals in vals_list:
-            if "planned_date_begin" in vals:
+            if "planned_date_begin" in vals and vals.get("planned_date_begin"):
```
yankinmax: here if we have empty values of `planned_date_begin` or `planned_date_end` on create, then we won't compute forecast as `forecast_date_planned_start` and `forecast_date_planned_end` will be empty too

yankinmax:
@gurneyalex I found an issue why I didn't get negative consolidated forecast on creating, only on write.
But, I still can't solve the issue for the tests.

project_forecast_line/models/forecast_line.py
```diff
@@ -90,10 +90,12 @@ def _compute_employee_forecast_line_id(self):
         employees = self.mapped("employee_id")
         date_froms = self.mapped("date_from")
         date_tos = self.mapped("date_to")
+        forecast_roles = self.mapped("forecast_role_id")
```
```suggestion
        forecast_roles = self.mapped("forecast_role_id") | employees.mapped("main_role_id")
```
otherwise the lookup on line 121 may give False

project_forecast_line/models/forecast_line.py
```diff
@@ -104,12 +106,21 @@ def _compute_employee_forecast_line_id(self):
...
         for rec in self:
             if rec.type == "confirmed" and rec.res_model != "hr.employee.forecast.role":
-                rec.employee_resource_forecast_line_id = capacities.get(
-                    (rec.employee_id.id, rec.date_from), False
+                resource_forecast_line = capacities.get(
+                    (rec.employee_id.id, rec.date_from, rec.forecast_role_id.id), False
                 )
+                if resource_forecast_line:
+                    rec.employee_resource_forecast_line_id = resource_forecast_line
+                else:
```
gurneyalex, line 128:
```suggestion
                else:
                # we did not find a forecast line with a matching role -> we default on a forecast line with the main role of the employee
```

project_forecast_line/models/project_task.py
gurneyalex: side note: planned_date_begin and planned_date_end are fields introduced in a enterprise module. We cannot depend on them being set on the CE version, this is a hack to avoid overloading the form view in a complex way or introducing a glue module depending on Odoo EE

your fix is correct

gurneyalex:
```suggestion
            if vals.get("planned_date_begin"):
```
is shorter and does the same

project_forecast_line/models/project_task.py
```diff
@@ -18,9 +18,9 @@ class ProjectTask(models.Model):
...
                 vals["forecast_date_planned_start"] = vals["planned_date_begin"]
-            if "planned_date_end" in vals:
+            if "planned_date_end" in vals and vals.get("planned_date_end"):
```
gurneyalex:
```suggestion
            if vals.get("planned_date_end"):
```

project_forecast_line/tests/test_forecast_line.py
```diff
@@ -603,10 +603,155 @@ def test_task_forecast_lines_consolidated_forecast_overallocation_multiple_tasks
...
+
+    def test_task_forecast_lines_employee_main_role(self):
+        """
+        Test forecast lines when employee has different roles
+        and different from employee's role is assigned to the task.
+
+        Employee has 2 forecast_role_id: consultant 75% and project manager 25%,
+        working 8h per day (standard calendar).
+        Create a task with forecast role developer, with remaining time = 8h
+        and a scheduled period starting and ending on the same day (22/01/2022 for instance).
```
gurneyalex:
```suggestion
        and a scheduled period starting and ending on the same day (14/01/2022 for instance).
```

project_forecast_line/tests/test_forecast_line.py
```diff
@@ -603,10 +603,155 @@ def test_task_forecast_lines_consolidated_forecast_overallocation_multiple_tasks
...
+        Assign this task to the user.
+
+        Expected: for the user, on 22/01/2022, 3 forecast lines.
```
gurneyalex:
```suggestion
        Expected: for the user, on 14/01/2022, 3 forecast lines.
```

and add the values

project_forecast_line/tests/test_forecast_line.py
```diff
@@ -603,10 +603,155 @@ def test_task_forecast_lines_consolidated_forecast_overallocation_multiple_tasks
...
+                "sequence": 1,
+            }
+        )
+        consultant_role = self.env["hr.employee.forecast.role"].search(
+            [
+                ("employee_id", "=", self.employee_consultant.id),
+                ("role_id", "=", self.role_consultant.id),
+            ]
+        )
+        consultant_role.rate = 75
```
gurneyalex, line 715: I would move this inside the with freeze_time() block below. If you don't do this you will have the forecast lines for the employee created relative to the test run date, and likely there will be no line for the 14/01/2022 which is officially in the past as of now.

project_forecast_line/tests/test_forecast_line.py
```diff
@@ -603,10 +603,155 @@ def test_task_forecast_lines_consolidated_forecast_overallocation_multiple_tasks
...
```
gurneyalex, line 644: I would move this inside the with freeze_time() block below. If you don't do this you will have the forecast lines for the employee created relative to the test run date, and likely there will be no line for the 14/01/2022 which is officially in the past as of now.

project_forecast_line/models/project_task.py
yankinmax: thanks for explanation

project_forecast_line/models/forecast_line.py
```diff
@@ -85,8 +85,22 @@ class ForecastLine(models.Model):
...
+        Hook here to take control on states
+        for which the consumption is confirmed or not.
+
+        For instance, holidays requests and sales quotation lines
+        create lines of type "forecast" and won't be taken into account
+        during consolidated forecast computation, whereas tasks for project
+        which are in a running state create lines with type "confirmed"
+        and will be used to compute consolidated forecast.
+        """
+        return ("confirmed",)
```
yankinmax: @gurneyalex can you pls take a look?

gurneyalex: I would have made this a configuration sett_hg rather than a hook that needs extending.

yankinmax: @gurneyalex I've updated PR, can you pls take a look?

gurneyalex (approved):

PR #972, opened 2022-08-31
[IMP] project_forecast_line_bokeh_chart

gurneyalex:
Enhance the display in the bokeh chart to better display negative
consolidated capacity.

Without the fix, the stacked vbar would display the negative capacity
bar starting from the top of the stack and going down, superposed on
positive capacity bars, which was confusing.

This fix gathers overload capacity (negative consolidated capacities)
and plots them in a separacte vbar with only negative values, which
produces an easier to understand chart.

See https://github.com/bokeh/bokeh/issues/9011 for a discussion of this
stacked bar chart behavior in bokeh.


gurneyalex:
image
example plot with the fix

yankinmax (approved):
Thanks for the work done

PR #974, opened 2022-09-07
[15.0][FIX] project_forecast_line: add missing access rights

yankinmax:
These are basic concepts about access rights in `project_forecast_line`:
1. Internal users can read `forecast.line`. Noone can delete them as there are foreign keys and they will be removed when either project, task, employee etc. are removed.
2. Internal users can read `forecast.role`. HR Officer can do all operations.
3. Internal users can read `hr.employee.forecast.role`. HR Officer can do all operations.


gurneyalex (changes requested):
if you go this way, you must set ondelete="restrict" to all foreign keys referencing the roles.

yankinmax:
didn't know about this, thanks for review

gurneyalex (approved):

jcoux (approved):

PR #977, opened 2022-09-09
[15.0][IMP] project_forecast_line_holidays_public: Add a setting `Recompute forecast lines immediately`

yankinmax:
The issue is that on databases containing lots of employees, projects and tasks, public holidays creation can take huge amount of time or even fail, because forecast lines are recomputed.
We propose to add a setting to choose if we want immediate forecast lines recomputation on public holidays creation. When the setting is not checked, the creation of a public holiday does nothing in the forecast lines.


gurneyalex (approved):
Why are you deleting the pot file?

yankinmax:
@jcoux told me they are not recognized by odoo, so useless

gurneyalex (approved):

PR #979, opened 2022-09-13
[FIX] project_forecast_line: error in consolidated capacity

gurneyalex:
During some operations, forecast lines are deleted and recreated. 

This can lead to some hr.employee.forecast.role lines being deleted which leaves the forecast lines on the same period without a related document to store the consolidated capacity. Normally that line is recreated shortly afterwards but the creation does not recompute the link between the parent-less lines and the new one. This is typically the case when a leave request is confirmed: the forecast line of the leave disapears and the capacity forecast of the employee based on the updated work calendar is regenerated.

This patches forces the recomputation when new lines are created.

It also fixes the missing link to the employee for leave requests forecast lines.

### DIFF
--- a/project_forecast_line/models/forecast_line.py
+++ b/project_forecast_line/models/forecast_line.py
@@ -348,3 +348,26 @@ class ForecastLine(models.Model):
         # we routinely unlink forecast lines, let"s not fill the logs with this
         with mute_logger("odoo.models.unlink"):
             return super().unlink()
+
+    @api.model_create_multi
+    @api.returns("self", lambda value: value.id)
+    def create(self, vals_list):
+        records = super().create(vals_list)
+        employee_role_lines = records.filtered(
+            lambda r: r.res_model == "hr.employee.forecast.role"
+        )
+        if employee_role_lines:
+            # check for existing records which could have the new lines as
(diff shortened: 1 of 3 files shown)


project_forecast_line/models/forecast_line.py
```diff
@@ -348,3 +348,26 @@ def unlink(self):
...
+        records = super().create(vals_list)
+        employee_role_lines = records.filtered(
+            lambda r: r.res_model == "hr.employee.forecast.role"
+        )
+        if employee_role_lines:
+            # check for existing records which could have the new lines as
+            # employee_resource_forecast_line_id
+            other_lines = self.search(
+                [
+                    ("employee_resource_forecast_line_id", "=", False),
```
yankinmax, line 364: if we're checking new `employee_resource_forecast_line_id` why we are filtering on empty?

gurneyalex: we want to find lines that could be linked to the created (we are in the create() method, we know which are the created lines) through the employee_resource_forecast_line_id, so we look for lines which don't already have a value in that field.

yankinmax (approved):

ntsirintanis (approved):

PR #980, opened 2022-09-14
[15.0][FIX] project_timeline: missing action

jguenat:
Hello,
On my database, the timeline view was not available for projects.
As explained [there](https://github.com/odoo/odoo/blob/1736d5a95825c5fe74c7d0103351efb8565da6ee/addons/project/views/project_views.xml#L743), it is required to modify both actions as sometimes one or the other is used.


pedrobaeza (approved):

jguenat:
Hello @chrisandrewmann , can you review this? Thanks

chrisandrewmann (approved):

chrisandrewmann:
@jguenat No problem, hope i've done it ok as i've never approved one before!

PR #981, opened 2022-09-15
[FIX] project_forecast_line: Remove setting and add new consolidated field

yankinmax:
The idea is to have two consolidated forecast fields, instead of setting.

### DIFF
--- a/project_forecast_line/models/forecast_line.py
+++ b/project_forecast_line/models/forecast_line.py
@@ -90,13 +90,8 @@ class ForecastLine(models.Model):
         "forecast.line", "employee_resource_forecast_line_id"
     )
 
-    def _get_consumption_states(self):
-        consumption_states = self.env.company.forecast_consumption_states
-        return tuple(consumption_states.split("_"))
-
     @api.depends("employee_id", "date_from", "type", "res_model")
     def _compute_employee_forecast_line_id(self):
-        consumption_states = self._get_consumption_states()
         employees = self.mapped("employee_id")
         main_roles = employees.mapped("main_role_id")
         date_froms = self.mapped("date_from")
@@ -121,10 +116,7 @@ class ForecastLine(models.Model):
                 (line.employee_id.id, line.date_from, line.forecast_role_id.id)
             ] = line.id
         for rec in self:
-            if (
-                rec.type in consumption_states
-                and rec.res_model != "hr.employee.forecast.role"
-            ):
+            if rec.type == "confirmed" and rec.res_model != "hr.employee.forecast.role":
                 resource_forecast_line = capacities.get(
                     (rec.employee_id.id, rec.date_from, rec.forecast_role_id.id), False
                 )
--- a/project_forecast_line/models/forecast_line.py
+++ b/project_forecast_line/models/forecast_line.py
@@ -61,11 +61,18 @@ class ForecastLine(models.Model):
         "holidays, sales, or tasks. ",
     )
     consolidated_forecast = fields.Float(
+        help="Consolidated forecast for lines of all types consumed",
         digits=(12, 5),
         store=True,
         compute="_compute_consolidated_forecast",
     )
-
+    confirmed_consolidated_forecast = fields.Float(
+        string="Confirmed lines consolidated forecast",
+        help="Consolidated forecast for lines of type confirmed",
+        digits=(12, 5),
+        store=True,
+        compute="_compute_consolidated_forecast",
+    )
     currency_id = fields.Many2one(related="company_id.currency_id", store=True)
     company_id = fields.Many2one(
         "res.company", required=True, default=lambda s: s.env.company
@@ -112,11 +119,15 @@ class ForecastLine(models.Model):
             lines = self.env["forecast.line"]
         capacities = {}
         for line in lines:
-            capacities[
-                (line.employee_id.id, line.date_from, line.forecast_role_id.id)
-            ] = line.id
+            employee_id = line.employee_id
+            date_from = line.date_from
+            forecast_role_id = line.forecast_role_id
+            capacities[(employee_id.id, date_from, forecast_role_id.id)] = line.id
         for rec in self:
-            if rec.type == "confirmed" and rec.res_model != "hr.employee.forecast.role":
+            if (
+                rec.type in ("forecast", "confirmed")
+                and rec.res_model != "hr.employee.forecast.role"
+            ):
                 resource_forecast_line = capacities.get(
                     (rec.employee_id.id, rec.date_from, rec.forecast_role_id.id), False
                 )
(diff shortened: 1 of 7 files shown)


project_forecast_line/models/forecast_line.py
```diff
@@ -91,12 +98,22 @@ class ForecastLine(models.Model):
...
+        """
+        Hook here to take control on states
+        for which the consumption is confirmed or not.
+        For instance, holidays requests and sales quotation lines
+        create lines of type "forecast" and won't be taken into account
+        during consolidated forecast computation, whereas tasks for project
+        which are in a running state create lines with type "confirmed"
+        and will be used to compute consolidated forecast.
+        """
+        return ("confirmed",)
```
gurneyalex: this method should be removed

project_forecast_line/models/forecast_line.py
```diff
@@ -140,6 +154,16 @@ def _compute_employee_forecast_line_id(self):
             else:
                 rec.employee_resource_forecast_line_id = False
 
+    def _get_resource_forecast_line_values(self):
+        data = {}
+        for d in self.env["forecast.line"].read_group(
+            [("employee_resource_forecast_line_id", "in", self.ids)],
+            fields=["forecast_hours"],
+            groupby=["employee_resource_forecast_line_id"],
```
```suggestion
            groupby=["employee_resource_forecast_line_id", "type"], lazy=False
```

project_forecast_line/models/forecast_line.py
```diff
@@ -140,6 +154,16 @@ def _compute_employee_forecast_line_id(self):
...
+        ):
+            data[d["employee_resource_forecast_line_id"][0]] = d["forecast_hours"]
```
gurneyalex: Build a dictionary with both groups

```suggestion
             line_id = d["employee_resource_forecast_line_id"][0]]
             if line_id not in data:
                 data[line_id] = {}
             data[line_id][d['type']] = d["forecast_hours"]
```

project_forecast_line/models/forecast_line.py
```diff
@@ -159,15 +183,24 @@ def _convert_forecast(self, data):
...
+        for possible forecast lines type extending.
+        Here we compute consolidated forecast only for confirmed forecast lines
+        and both forecast and confirmed forecast line types
+        """
+
+        confirmed_lines_values = self._get_resource_forecast_line_values()
+        self._compute_employee_forecast_line_id(
+            consumption_states=lambda self: self._get_consumption_states()
+            + ("forecast",)
+        )
```
```suggestion

```

project_forecast_line/models/forecast_line.py
```diff
@@ -159,15 +183,24 @@ def _convert_forecast(self, data):
...
+        consumed_lines_values = self._get_resource_forecast_line_values()
```
```suggestion
```

project_forecast_line/models/forecast_line.py
```diff
@@ -159,15 +183,24 @@ def _convert_forecast(self, data):
...
-        ):
-            data[d["employee_resource_forecast_line_id"][0]] = d["forecast_hours"]
+        """
+        The idea is to use same computed method for consolidated forecast fields,
+        for possible forecast lines type extending.
+        Here we compute consolidated forecast only for confirmed forecast lines
+        and both forecast and confirmed forecast line types
+        """
+
+        confirmed_lines_values = self._get_resource_forecast_line_values()
```
```suggestion
        grouped_lines_values = self._get_resource_forecast_line_values()
```

project_forecast_line/models/forecast_line.py
```diff
@@ -159,15 +183,24 @@ def _convert_forecast(self, data):
...
+            consumption_states=lambda self: self._get_consumption_states()
+            + ("forecast",)
+        )
+        consumed_lines_values = self._get_resource_forecast_line_values()
         for rec in self:
-            rec.consolidated_forecast = rec._convert_forecast(data)
+            rec.confirmed_consolidated_forecast = rec._convert_forecast(
+                confirmed_lines_values
+            )
+            rec.consolidated_forecast = rec._convert_forecast(consumed_lines_values)
```
gurneyalex: This needs rewriting:

1. change the _convert_forecast method to a method converting hours to days, taking a single argument and returning a single value
2. move back the logic that has been moved in that method to get the positive value for consumption lines and the sum for employee.forecast.role related lines
3. you have in grouped_line_values a dictionary which maps an employee.forecast.role line to a dictionary which has {'confirmed': total_hours, 'forecast': total_hours} and possibly one of the key is missing -> use this to compute both fields for the forecast lines. For the other lines the value is still the opposite of the consumption.

yankinmax:
@gurneyalex I suppose this is what we need. Give it a look, please.

project_forecast_line/models/forecast_line.py
```diff
@@ -140,34 +143,55 @@ def _compute_employee_forecast_line_id(self):
...
+            [("employee_resource_forecast_line_id", "in", self.ids)],
+            fields=["forecast_hours"],
+            groupby=["employee_resource_forecast_line_id", "type"],
+            lazy=False,
+        )
+        for d in grouped_line_result:
+            line_id = d["employee_resource_forecast_line_id"][0]
+            if line_id not in data:
+                data[line_id] = {}
+            data[line_id][d["type"]] = d["forecast_hours"]
```
```suggestion
                data[line_id] = {'confirmed': 0, 'forecast': 0}
            data[line_id][d["type"]] += d["forecast_hours"]
```

gurneyalex: I think you need the sum, please double check what you get if you have 2 confirmed tasks on the same day.

gurneyalex (changes requested):

project_forecast_line/models/forecast_line.py
yankinmax: @gurneyalex I've updated the code and tests. Thanks, for the review.
However Travis is failing because of the `project_task_code` unique sequence.
Use of `@tagged("-at_install", "post_install")` doesn't help. Any ideas?

gurneyalex (approved):

project_forecast_line/tests/test_forecast_line.py
```diff
@@ -489,39 +490,91 @@ def setUpClass(cls):
...
+            sum(f.confirmed_consolidated_forecast for f in forecast), 4.0
+        )
+        consol_employee_forecast = sum(
+            f.employee_resource_forecast_line_id.consolidated_forecast for f in forecast
+        )
+        confir_employee_forecast = sum(
+            f.employee_resource_forecast_line_id.confirmed_consolidated_forecast
+            for f in forecast
+        )
+        self.assertAlmostEqual(consol_employee_forecast, -13.0)
```
gurneyalex, line 567: running the test on rebase after the fix shows -16.0 here. But I don't understand what you are checking: on the sum on line 557 and 563, you are summing the same value since all the forecast lines have the same employee_resource_forecast_line_id. 

This test deserves some explanation (and likely a rewrite or a fix)

gurneyalex: for the record, I'm taking the opportunity of the OCA code sprint to work on this.

PR #985, opened 2022-09-23
[13.0][FIX] project_stock: Change cancel process (set as quantity_done = 0)

victoralmau:
Only cancel the stock moves not done

@Tecnativa TT39138

### DIFF
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -90,12 +90,14 @@ class Task(models.Model):
         for item in self:
             item.scrap_count = count_data.get(item.id, 0)
 
+    @api.depends("move_ids", "move_ids.state")
     def _compute_allow_moves_action_confirm(self):
         for item in self:
             item.allow_moves_action_confirm = any(
                 move.state == "draft" for move in item.move_ids
             )
 
+    @api.depends("move_ids", "move_ids.state")
     def _compute_allow_moves_action_assign(self):
         for item in self:
             item.allow_moves_action_assign = any(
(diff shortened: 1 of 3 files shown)


project_stock/models/project_task.py
```diff
@@ -90,12 +90,14 @@ def _compute_scrap_move_count(self):
         for item in self:
             item.scrap_count = count_data.get(item.id, 0)
 
+    @api.depends("move_ids.state")
```
pedrobaeza: For additions and removals

```suggestion
    @api.depends("move_ids", "move_ids.state")
```

project_stock/models/project_task.py
```diff
@@ -196,7 +198,7 @@ def action_cancel(self):
         """Cancel the stock moves and remove the analytic lines created from
         stock moves when cancelling the task.
         """
-        self.move_ids.write({"state": "cancel"})
+        self.move_ids.write({"quantity_done": 0})
```
pedrobaeza: Don't you cancel the move afterwards?

victoralmau: Cancelling a stock move is not possible with UX and not allowed by odoo https://github.com/odoo/odoo/blob/13.0/addons/stock/models/stock_move.py#L1324

Defining as quantity done 0 is the most optimal way to do this "_cancellation_" process. Also, this change does re-generate correctly `stock.quant` records.

pedrobaeza (approved):

victoralmau:
Ping @sergio-teruel

sergio-teruel (approved):

PR #994, opened 2022-10-11
Fix project forecast line remove setting

gurneyalex:
Replace the setting for the computation of the non confirmed consolidated forecast with a second field


gurneyalex:
superseeds  #981

yankinmax (approved):
thx @gurneyalex for the work done

leemannd (approved):

PR #996, opened 2022-10-12
[15.0][IMP] Automatic migration from project_category to project_type on Upgrade 14->15

CRogos:
project_category was renamed to project_type with the v15 migration. 
Adding project_category with a dependency on project_type  will trigger the automated installation after upgrade.


pedrobaeza:
This is not done this way, but adding the renamed/merged module in OpenUpgrade

CRogos:
OK, thx. Is OpenUpgrade also used by the Odoo Enterprise upgrade path?

PR #999, opened 2022-10-18
[15.0][IMP] project_forecast_line: Some improvements

yankinmax:
1. Add computed fields to written fields to trigger forecast lines update
To reproduce the case:
  - Create forecast on project task for example.
  - Change initially planned hours.
  - Forecast lines are not recomputed.
What is expected:
when initially planned hours are changed remaining hours are being recomputed. And it should trigger forecast line update.
2. Add hook for several UoM hours/days conversion
3. Reorder forecast search filters.

### DIFF
--- a/project_forecast_line/models/forecast_line.py
+++ b/project_forecast_line/models/forecast_line.py
@@ -158,8 +158,16 @@ class ForecastLine(models.Model):
             data[line_id][d["type"]] += d["forecast_hours"]
         return data
 
+    @api.model
+    def _get_consolidation_uom(self):
+        """
+        Returns the unit of measure used for the consolidated forecast.
+        The default is days.
+        """
+        return self.env.ref("uom.product_uom_day")
+
     def _convert_hours_to_days(self, hours):
-        to_convert_uom = self.env.ref("uom.product_uom_day")
+        to_convert_uom = self._get_consolidation_uom()
         project_time_mode_id = self.company_id.project_time_mode_id
         return project_time_mode_id._compute_quantity(
             hours, to_convert_uom, round=False
--- a/project_forecast_line/models/project_project.py
+++ b/project_forecast_line/models/project_project.py
@@ -2,6 +2,8 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import models
 
+from ..utils import get_written_computed_fields
+
 
 class ProjectProject(models.Model):
     _inherit = "project.project"
@@ -9,10 +11,15 @@ class ProjectProject(models.Model):
     def _update_forecast_lines_trigger_fields(self):
         return ["stage_id"]
 
+    # TODO: Stored computed fields updates don't go through write,
+    # but they do go through _write. Check if possible to override
+    # that method instead, to avoid maintaing get_written_computed_fields
     def write(self, values):
         res = super().write(values)
-        written_fields = list(values.keys())
+        written_computed_fields = get_written_computed_fields(
+            self, tuple(sorted(values))
+        )
         trigger_fields = self._update_forecast_lines_trigger_fields()
-        if any(field in written_fields for field in trigger_fields):
+        if any(field in written_computed_fields for field in trigger_fields):
             self.task_ids._update_forecast_lines()
         return res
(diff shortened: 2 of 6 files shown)


project_forecast_line/models/project_project.py
```diff
@@ -9,10 +9,24 @@ class ProjectProject(models.Model):
     def _update_forecast_lines_trigger_fields(self):
         return ["stage_id"]
 
+    def _get_written_computed_fields(self, values):
+        # get written and computed field names
+        task_computed = []
+        written_fields = list(values.keys())
+        tree = self.pool.field_triggers.get(self._fields[next(iter(written_fields))])
```
gurneyalex: Are you sure about this? It looks like you are only considering the first written field, and not any of the others.

project_forecast_line/models/project_project.py
```diff
@@ -9,10 +9,24 @@ class ProjectProject(models.Model):
...
+        if tree:
+            tocompute = (
+                self.sudo().with_context(active_test=False)._modified_triggers(tree)
+            )
+            for field, _records, _create in tocompute:
+                if field.model_name == "project.project":
+                    task_computed.append(field.name)
+        return written_fields + task_computed
```
gurneyalex: This whole implementation looks complex to me, and I'm concerned of the speed impact it may have as you recompute this on all create and write calls, whereas the list of values that can trigger a write on another field is fixed. 

I suggest changing the `values` parameter to a sorted tuple of field names (you would call `self._get_written_computed_fields(tuple(sorted(values)))` ) so you can used a cached decorator to prevent recomputing this all the time.

gurneyalex (changes requested):

project_forecast_line/models/project_project.py
yankinmax: I see the problem, I'll correct.

yankinmax: basically, I use implementation, but didn't pay attention to the details here:
https://github.com/odoo/odoo/blob/73de38e0e35f4ffb8069441cf11ebd48ff5aadb1/odoo/models.py#L5984

and maybe it would be better to use v16 variant
https://github.com/odoo/odoo/blob/99d7d0a7b875ffb1e5ce00cd7f3fc2c9e0359796/odoo/models.py#L5980

yankinmax:
@gurneyalex I've updated PR with changes you've requested (I suppose :smile:).
For some reason, Odoo core test is failing. I've checked it locally after source update and don't have this issue

project_forecast_line/models/forecast_line.py
```diff
@@ -158,8 +158,12 @@ def _get_grouped_line_values(self):
             data[line_id][d["type"]] += d["forecast_hours"]
         return data
 
+    @api.model
+    def _get_convert_uom(self):
```
gurneyalex: I would rename this
```suggestion
    def _get_consolidation_uom(self):
        """return the unit of measure used for the consolidated forecast. The default is days"""
```

project_forecast_line/models/forecast_line.py
```diff
@@ -158,8 +158,12 @@ def _get_grouped_line_values(self):
...
+        return self.env.ref("uom.product_uom_day")
+
     def _convert_hours_to_days(self, hours):
-        to_convert_uom = self.env.ref("uom.product_uom_day")
+        to_convert_uom = self._get_convert_uom()
```
```suggestion
        to_convert_uom = self._get_consolidation_uom()
```

gurneyalex (approved):

gurneyalex (approved):

gurneyalex (approved):
tests are red but this seems to be related to another module.

project_forecast_line/models/project_project.py
```diff
@@ -11,8 +13,10 @@ def _update_forecast_lines_trigger_fields(self):
 
     def write(self, values):
         res = super().write(values)
-        written_fields = list(values.keys())
+        written_computed_fields = get_written_computed_fields(
+            self, tuple(sorted(values))
+        )
```
ivantodorovich, line 21: Stored computed fields updates don't go through `write`, but they do go through `_write`.
You could override that method instead, and thus you avoid maintaing this complex utility method

ivantodorovich (approved):
non blocking comment

project_forecast_line/models/project_project.py
yankinmax: thanks, @ivantodorovich for the suggestion. For the moment I've added it in TODO, I'll check when have time.

PR #1003, opened 2022-10-25
[14.0][FIX] project_timesheet_time_control: Avoid texts on buttons

pedrobaeza:
Since v14, `string` attribute of <button> elements are directly shown in the views as text instead of being a tooltip, and thus, it takes a lot of space in screen.

And it was not the intended visualization when the module was conceived, just a side effect.

Thus, we restore what we expect, moving `string` to `title` for having the tooltips again.

**Before**

imagen

**After**

Selección_004

@Tecnativa TT40169

### DIFF
--- a/project_timesheet_time_control/views/account_analytic_line_view.xml
+++ b/project_timesheet_time_control/views/account_analytic_line_view.xml
@@ -15,23 +15,21 @@
                 <field name="show_time_control" invisible="1" />
                 <button
                     name="button_resume_work"
-                    string="Resume work"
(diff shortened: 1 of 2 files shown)


sergio-teruel (approved):

rvalyi (approved):

pedrobaeza:
Thanks for the quick reaction.

Proceeding then:

/ocabot merge minor

pedrobaeza:
And the related one:

https://github.com/OCA/timesheet/pull/535

PR #916, opened 2022-04-25
[15.0][MIG][project_task_stage_state] Standard migration to project_task_stage_state

manuelregidor:
Standard migration.

PR #1009, opened 2022-11-01
16.0 mig project task stage state

aiendry-aktivsoftware:


misern2 (approved):
LGTM :+1:

Kata-2022:
Works Correctly.

yostashiro:
Just quoting comment from https://github.com/OCA/project/pull/1010#issuecomment-1302895584.

> Looks like the commit history prior to the 15.0 migration is missing. The reason seems to be that 15.0 migration was somehow merged without history. https://github.com/OCA/project/commits/15.0/project_task_stage_state
> 
> Do we need to take some remedial actions here (I suppose yes)?
> 
> Tagging some people involved in the 15.0 migration #916 here: @rafaelbn @dreispt @manuelregidor

AungKoKoLin1997 (commented):
Can you please update USAGE.rst and regenerate README.rst. I don't think you generated correctly. This is still pointing out to 15.0 branch.

project_task_stage_state/models/project_task_type.py
```diff
@@ -0,0 +1,20 @@
...
+    ("cancelled", "Cancelled"),
+]
+
+
+class ProjectTaskType(models.Model):
+    """Added state in the Project Task Type."""
+
+    _inherit = "project.task.type"
+
+    state = fields.Selection(_TASK_STATE)
```
Shide, line 20: Maybe the best approach to allow reusability of the module will be:
```python
@api.model
def _get_task_states(self):
    return [
        ("draft", "New"),
        ("open", "In Progress"),
        ("pending", "Pending"),
        ("done", "Done"),
        ("cancelled", "Cancelled"),
    ]

state = fields.Selection(selection="_get_task_states")
```

It will increase reusability because you can add more (without using selection_add) or just replace all the suggested task states and replace by your preference.

I have a module that with this change will be viable to use this module.

Shide (commented):
I would like to have this change, but I can make a PR after the migration.

If anyone think this could be a better approach (including you), you could consider migrating the module with this change incorporated.

rafaelbn:
Hello @dreispt , to adopt this module and be contributor and maintainers we need to improve it. Could you give us you opinion about https://github.com/OCA/project/pull/1009#pullrequestreview-1173768006 please? Thank you!

PR #1010, opened 2022-11-02
[16.0][MIG] project_task_stage_state: Migration to 16.0

AungKoKoLin1997:


yostashiro:
Looks like the commit history prior to the 15.0 migration is missing. The reason seems to be that 15.0 migration was somehow merged without history. https://github.com/OCA/project/commits/15.0/project_task_stage_state

Do we need to take some remedial actions here (I suppose yes)?

Tagging some people involved in the 15.0 migration https://github.com/OCA/project/pull/916 here: @rafaelbn @dreispt @manuelregidor

yostashiro:
@AungKoKoLin1997 I think you can close this one. There is a preceding migration PR https://github.com/OCA/project/pull/1009.

PR #1014, opened 2022-11-02
16.0 mig project task default stage

uncannycs:


AungKoKoLin1997 (commented):
Function review: It worked as expected.

leemannd (approved):

emagdalenaC2i (approved):
LGTM

project_task_default_stage/data/project_data.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <!-- Task Stages -->
+    <record id="project_tt_analysis" model="project.task.type">
```
legalsylvain, line 4: This look demo data. Why it is in the data folder ? 
Analysis / Specification / Merge are IT stages, it's not universal.

Don't you think ?

pedrobaeza: This has been this way for ages, following Odoo's inheritance that included this on old versions. They can be removed, but I find them useful as initial stage set. They can be put as not default.

legalsylvain: > This has been this way for ages, following Odoo's inheritance that included this on old versions.

Yes, you're right, the code comes from Odoo code since V9.

> They can be removed

Well, the problem I see is that when we install modules, we expect that there are no "functional data". Only technical one that should not be dropped. the question I asked myself was, "if I delete this data, will it make the application crash? like ``base.main_company``, ``product.product_category_all``.
>  but I find them useful as initial stage set

well, I'm in favor to remove that data from the "data". 2 options : 
- we move data into demo folder.
- we create a trivial `` project_task_it_task_data``that just load data. after all, this data can be useful, even outside the context of the current module? 

I can make a PR if you are agree with one of the solution.

pedrobaeza: I don't consider them demo data, and also an extra module is overkilled. Let's remove them if others agree.

PR #1020, opened 2022-11-04
[15.0][MIG] project_stock: Migration to 15.0

victoralmau:
Migration to 15.0

@Tecnativa TT40363


project_stock/models/stock_move.py
```diff
@@ -45,6 +46,7 @@ def _prepare_analytic_line_from_task(self):
             "product_uom_id": self.product_uom.id,
             "company_id": analytic_account.company_id.id or self.env.company.id,
             "partner_id": task.partner_id.id or task.project_id.partner_id.id or False,
+            "project_id": task.project_id.id,
```
pedrobaeza: This was removed on purpose, as only timesheets have the project filled.

project_stock/models/stock_move.py
```diff
@@ -60,10 +62,10 @@ def _prepare_analytic_line_from_task(self):
         if "product_id" in analytic_line_fields:
             vals["product_id"] = product.id
         # Extra field added in hr_timesheet addon
-        if "employee_id" in analytic_line_fields:
+        if "employee_id" in analytic_line_fields and task.user_ids:
```
pedrobaeza: This shouldn't be filled as well. These are not timesheets.

project_stock/tests/test_project_stock.py
```diff
@@ -51,121 +58,130 @@ def test_project_task_misc(self):
...
+            task.stock_analytic_tag_ids,
+            task.stock_analytic_line_ids.mapped("tag_ids"),
         )
         self.assertIn(
             self.analytic_account,
-            self.task.stock_analytic_line_ids.mapped("account_id"),
+            task.stock_analytic_line_ids.mapped("account_id"),
         )
 
+    @users("manager-user")
```
pedrobaeza: Do both with admin and this user

project_stock/tests/test_project_stock.py
```diff
@@ -51,121 +58,130 @@ def test_project_task_misc(self):
...
 
     def _test_task_analytic_lines_from_task(self, amount):
-        self.assertEqual(len(self.task.stock_analytic_line_ids), 2)
-        self.assertEqual(
-            sum(self.task.stock_analytic_line_ids.mapped("unit_amount")), 3
-        )
-        self.assertEqual(
-            sum(self.task.stock_analytic_line_ids.mapped("amount")), amount
-        )
+        task = self.env["project.task"].browse(self.task.id)
```
pedrobaeza: For doing the diff less bigger, just do:

```suggestion
        self.task = self.env["project.task"].browse(self.task.id)
```

pedrobaeza (changes requested):

project_stock/models/stock_move.py
victoralmau: Reviewing v14, I see that it auto-defined `project_id` with `hr_timesheet` (not sure how yet).

pedrobaeza: But these are not timesheets. That's why you shouldn't set it.

project_stock/tests/test_project_stock.py
victoralmau: Ok, I have added admin user in tests, but currently it is not possible to add multiple users with the `@users()` decorator without having side effect (no roolback of changes per user).

pedrobaeza: What about this?: Create another method `test_..._admin` with the other @users decorator and calling the first method

victoralmau:
Changes done and tests fixed.

project_stock/models/project_task.py
```diff
@@ -74,6 +74,9 @@ class ProjectTask(models.Model):
     )
     stock_analytic_tag_ids = fields.Many2many(
         comodel_name="account.analytic.tag",
+        relation="account_analytic_tag_project_task_stock_rel",
```
pedrobaeza, line 77: This will require migration scripts?

project_stock/security/project_stock_security.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="account_analytic_line_project_stock_rule" model="ir.rule">
+        <field name="name">account.analytic.line.project_stock</field>
+        <field name="model_id" ref="analytic.model_account_analytic_line" />
+        <field name="domain_force">[('stock_task_id', '!=', False)]</field>
+        <field
+            name="groups"
+            eval="[(4, ref('analytic.group_analytic_accounting')), (4, ref('project.group_project_manager'))]"
```
pedrobaeza: Why is this needed? This should be bypassed other way (with `sudo` for example).

project_stock/tests/test_project_stock.py
```diff
@@ -76,7 +93,16 @@ def test_project_task_without_analytic_account(self):
         self.task.action_done()
         self.assertFalse(self.task.stock_analytic_line_ids)
 
-    def test_project_task_analytic_lines_without_tags(self):
+    @users("admin")
```
pedrobaeza: No need of this decorator if you keep the other method without renaming nor users decorator (it already run with admin), and only add the other method.

project_stock/models/project_task.py
victoralmau: Yes, it requires renaming the table in migration script (already done).

project_stock/security/project_stock_security.xml
victoralmau: This is necessary because if `hr_timesheet` addon is installed, the analytic lines generated according to the rules cannot be accessed: https://github.com/odoo/odoo/blob/15.0/addons/hr_timesheet/security/hr_timesheet_security.xml

pedrobaeza: And that's the idea... HR timesheet users only see timesheet entries, not these ones. Only accounting managers with full access will see all the analytic entries.

victoralmau: Ok, removed

pedrobaeza (approved):
Code review

project_stock/README.rst
```diff
@@ -0,0 +1,120 @@
...
+    * `Type of operation`: Delivery
+    * `Default Source Location`: WH/Stock
+    * `Default Destination Location`: WH/Stock/Shelf 1
+#. Go to *Project -> Configuration -> Projects*.
+#. Create a new project with the following options:
+    * `Name`: Task material
+    * `Operation type`: Task material
+#. Go to *Project -> Configuration -> Task Stages* and edit some records.
+    * `In progress`: Check Use Stock Moves option and add the created project.
+    * `Done`: Check Use Stock Moves option + Lock Stock Moves and add the created project.
```
chienandalu:
```suggestion
    * `Done`: Check Use Stock Moves option + Done Stock Moves and add the created project.
```

project_stock/README.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Usage
+=====
+
+#. Go to *Projects -> Task material (project)* and create a task and edit it.
+#. *Stock Info* is displayed in the *Extra info* tab with the same project information, but it can be modified.
+#. Add some product to *Stock Info* tab and set some initial demand (1 for example).
+#. Change the stage to Done.
+#. *Stock Info* tab is readonly and some buttons show in header:
+    * `Confirm material`: Confirm all moves.
```
chienandalu: This one show up only in the "In progress" stage

chienandalu (changes requested):
I detected an issue in my functional test:

- When I try to add more than one product line to the task and save it the following error is raised:

```
UncaughtPromiseError > TypeError
Uncaught Promise > Cannot read properties of undefined (reading 'type')

TypeError: Cannot read properties of undefined (reading 'type')
    at compareRecords (http://localhost:15069/web/assets/debug/web.assets_backend.js:69576:45)
    at stableCompare (http://localhost:15069/web/assets/debug/web.assets_common.js:66639:25)
    at Array.sort (<anonymous>) (/web/static/src/legacy/js/views/basic/basic_model.js:5170)
    at Object.stableSort (http://localhost:15069/web/assets/debug/web.assets_common.js:66638:22) (/web/static/src/legacy/js/core/utils.js:904)
    at Class._sortList (http://localhost:15069/web/assets/debug/web.assets_backend.js:69588:19) (/web/static/src/legacy/js/core/utils.js:903)
    at http://localhost:15069/web/assets/debug/web.assets_backend.js:69269:26 (/web/static/src/legacy/js/views/basic/basic_model.js:5182)
    at async Promise.all (index 4) (/web/static/src/legacy/js/views/basic/basic_model.js:4863)
    at async Promise.all (index 0)

victoralmau:
> I detected an issue in my functional test:
> 
>     * When I try to add more than one product line to the task and save it the following error is raised:

Changes done and and problem solved.

chienandalu (approved):
Now the test went perfect :+1: Good job!

PR #1023, opened 2022-11-09
[15.0][MIG] project_stock_product_set: Migration to 15.0

victoralmau:
Migration to 15.0

@Tecnativa TT40364


chienandalu (commented):
When I set my task as done, I can still add sets to the task. Shouldn't be the button hidden in this case?

project_stock_product_set/tests/test_project_stock_product_set.py
```diff
@@ -21,7 +22,9 @@ def setUpClass(cls):
             }
         )
 
+    @users("basic-user")
```
pedrobaeza: Do the test also with admin. This way, if both tests fail in a new version, we know that it's not permission issue.

project_stock_product_set/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,3 @@
+It is important to clarify some concepts at the level of project stages:
+#. Use Stock Moves: If you mark this check, when a task goes to this state, it will use stock moves.
```
pedrobaeza: You need an empty line, plus use `* ` instead of `#. `

project_stock_product_set/wizard/product_set_add.py
```diff
@@ -0,0 +1,47 @@
...
+    def _prepare_stock_move_lines(self):
+        move_lines = []
+        for _seq, set_line in enumerate(self._get_lines(), start=1):
+            values = self.prepare_stock_move_data(set_line)
+            move_lines.append((0, 0, values))
+        return move_lines
+
+    def prepare_stock_move_data(self, set_line):
+        self.ensure_one()
+        line_values = set_line.prepare_stock_move_values(self.task_id, self.quantity)
```
pedrobaeza: Why having 2 prepare methods per line?

victoralmau: I have changed this so as not to have another separate method.

victoralmau:
> When I set my task as done, I can still add sets to the task. Shouldn't be the button hidden in this case?

Ok, good suggestion, it is already changed to hide button.

project_stock_product_set/wizard/product_set_add.py
```diff
@@ -0,0 +1,42 @@
...
+        else None,
+        ondelete="cascade",
+    )
+
+    def _prepare_stock_move_lines(self):
+        move_lines = []
+        sm_model = self.env["stock.move"]
+        for _seq, set_line in enumerate(self._get_lines(), start=1):
+            values = set_line.prepare_stock_move_values(self.task_id, self.quantity)
+            values.update(sm_model.play_onchanges(values, values.keys()))
```
pedrobaeza: But these lines should be done in the `prepare_stock_move_values` (and it should be `_prepare_stock_move_values`).

pedrobaeza (approved):

chienandalu (approved):
:+1: Thanks!

PR #1029, opened 2022-11-16
[15.0][UPD] project_forecast_line: update forecast lines even if employee is not active

yankinmax:
We want to update forecast lines even if the employee is not active.
Hook can be overridden in custom modules to update only active employees with simple filtered.
```
def _get_task_employees(self):
    employees = super()._get_task_employees()
    return employees.filtered("active")
```

### DIFF
--- a/project_forecast_line/models/project_task.py
+++ b/project_forecast_line/models/project_task.py
@@ -71,6 +71,9 @@ class ProjectTask(models.Model):
                     task.forecast_role_id = employee.main_role_id
                     break
 
+    def _get_task_employees(self):
+        return self.with_context(active_test=False).mapped("user_ids.employee_id")
+
     def _update_forecast_lines(self):
         today = fields.Date.context_today(self)
         forecast_vals = []
@@ -116,7 +119,7 @@ class ProjectTask(models.Model):
                 continue
             date_start = max(today, task.forecast_date_planned_start)
             date_end = max(today, task.forecast_date_planned_end)
-            employee_ids = task.mapped("user_ids.employee_id").ids
+            employee_ids = task._get_task_employees().ids
             if not employee_ids:
                 employee_ids = [False]
             _logger.debug(


damdam-s (approved):

ivantodorovich:
The hook seems a bit too specific and unnecessary to me

Can't you achieve the same in your custom module by overriding `_update_forecast_lines` like this?

```py
    def _update_forecast_lines(self):
        # OVERRIDE to update only lines with active employees
        self = self.filtered("employee_id.active")
        return super(HrLeave, self)._update_forecast_lines()
```

yankinmax:
@ivantodorovich for Leaves and Employee forecast role objects yes, we can override as you suggest. But not for the Tasks.
Employees on tasks are selected here:
https://github.com/OCA/project/blob/7fb2830300e9ee0a8229d612c25c73158d1547be/project_forecast_line/models/project_task.py#L119
And we can have several users, so employees related to a task. Then if we filter on tasks:
```
self = self.filtered("user_ids.employee_id.active")
```
we skip tasks we don't want to skip. Because some employee are active and some not.
I'm not sure even we can have preferred result with such filter.
So, I would prefer to have specific hook for this. And this is going to be actually important change, because there can be lots of employees and tons fo forecast line. This will actually improve performance also, not to create useless (on some projects) records.

ivantodorovich:
Then I recommend a different hook to be added on `project.task`: one called `_get_forecast_line_employees`

```
def _get_forecast_line_employees(self):
    return self.user_ids.employee_id
```

Then you can override this method on your custom module to filter out inactive employees.

project_forecast_line/models/project_task.py
```diff
@@ -116,18 +123,18 @@ def _update_forecast_lines(self):
...
                 task,
                 date_start,
                 date_end,
                 task.remaining_hours,
             )
-            forecast_hours = task.remaining_hours / len(employee_ids)
-            for employee_id in employee_ids:
+            forecast_hours = task.remaining_hours / len(employees)
+            for employee_id in employees:
+                if self._update_only_active_employee() and not employee_id.active:
```
ivantodorovich: If you end up keeping this. be careful there's a bug here.
On line 126, `employees` can be `[False]`, so you'd get attr error by trying to read `employee_id.active`

yankinmax:
**UPDATE**
FIX for `project_task_code`: null value in column code, that violates not-null constraint is included as `test-requirements.txt`
- [ ] https://github.com/OCA/project/pull/1032

yankinmax:
@ivantodorovich finally, while issue investigation I found that in the current module version, forecast lines aren't created for employees during task creation, if the employee is inactive.
So, I've changed the PR to fix that. And explained how this could be overridden in inherited modules.

ivantodorovich (approved):

leemannd:
@yankinmax The test dependency is failing, could you have a look at it?

PR #825, opened 2021-06-23
[14.0] mig project_gtd

fredzamoabg:
**project_gtd** module migration from version 10.0

PR #1038, opened 2022-12-01
[14.0] mig project_gtd

leemannd:
project_gtd module migration from version 10.0


leemannd:
I took over https://github.com/OCA/project/pull/825 with this change appliyed

Issue #1042, opened 2022-12-04
[15.0][project_stock] Error selecting type operation when any task is already created


Bilbonet:
## Module
project_stock

## Describe the bug
Error selecting type operation when any task is already created

## To Reproduce
**Affected versions**: 14.0

1. Create a new project
2. Create a new task for the project
3. Go to the project. In the stock section, select the picking operation type and save the change.

you get next error:

```
The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/bilbonet/odoo15/server/odoo/http.py", line 643, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/home/bilbonet/odoo15/server/odoo/http.py", line 301, in _handle_exception
    raise exception.with_traceback(None) from new_cause
AttributeError: 'stock.location' object has no attribute 'get_warehouse'
```

victoralmau:
The problem occurs in 15.0: https://github.com/OCA/project/pull/1046

Bilbonet:
Yes, sorry.
The problem was in 15.0

Thanks !!!!

PR #1046, opened 2022-12-09
[15.0][FIX] project_stock: Set the warehouse_id field correctly.

victoralmau:
Related to https://github.com/OCA/project/issues/1042

Set the `warehouse_id` field correctly.

@Tecnativa

### DIFF
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -152,7 +152,7 @@ class ProjectTask(models.Model):
             )
             moves.update(
                 {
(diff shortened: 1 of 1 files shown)


chienandalu (approved):
:+1:

pedrobaeza (approved):

pedrobaeza:
Are you sure this one is not needed in 14.0?

victoralmau:
Yes, in v14 we use https://github.com/odoo/odoo/blob/14.0/addons/stock/models/stock_location.py#L156-L160

PR #1052, opened 2023-01-04
[16.0][MIG] project_stock

RabbitJon-S73:
Standard migration to 16.0.


RabbitJon-S73:
@victoralmau

project_stock/models/project_task.py
```diff
@@ -72,12 +73,10 @@ class ProjectTask(models.Model):
...
-        comodel_name="account.analytic.tag",
-        relation="account_analytic_tag_project_task_stock_rel",
-        column1="project_task_id",
-        column2="account_analytic_tag_id",
-        string="Move Analytic Tags",
+    stock_analytic_distribution = fields.Json(
+        "Move Analytic Tags",
+        copy=True,
+        readonly=False,
     )
```
victoralmau: I'm not completely sure, but it is necessary to add a migration script to transform and keep the old data, right?

project_stock/tests/common.py
```diff
@@ -22,26 +22,15 @@ def setUpClass(cls):
...
-        cls.analytic_tag_1 = cls.env["account.analytic.tag"].create(
             {
-                "name": "Test tag 1",
-                "active_analytic_distribution": True,
-                "analytic_distribution_ids": [
-                    (0, 0, {"account_id": cls.analytic_account.id, "percentage": 100}),
-                ],
-            }
+                "name": "Test account",
+                "plan_id": cls.env.ref("analytic.analytic_plan_projects").id,
```
victoralmau: I believe that this plan should not be used, the new defined plans should be created with the corresponding distribution so that the value of the tests is the same.

victoralmau (changes requested):
Good job, thanks for your work

Ping @pedrobaeza

project_stock/models/project_task.py
pedrobaeza: Yes, this will require a migration script, although it's still not defined how to do it on core fields, and this will probably generates an openupgradelib method, so it can be left as is for now.

project_stock/tests/common.py
RabbitJon-S73: Plan is now a required field, that's why I used it https://github.com/odoo/odoo/blob/6b101d5ff0ad78b9ae2ad312f362c2224e7af457/addons/analytic/models/analytic_account.py#L39

victoralmau: Ok sorry for the misunderstanding, I meant to create a new plan (`account.analytic.plan`) in test and not re-use an existing one.

victoralmau (approved):
Code review OK

PR #1053, opened 2023-01-13
[FIX] project_task_code: Make sequence available for all companies

rolandojduartem:
Currently, the sequence that generates task codes is only available for the
main company [1], which could cause issues at the moment of the installation
process with another modules in multi-companies cases, for example
project_timesheet_holiday [2] because in the other companies, this line [3]
will return False or "/" and will trigger the UNIQUE constraint. With
this change, the sequence can be used by any company [4] using the False
value in the field company_id.

References:
[1] https://github.com/odoo/odoo/blob/7b6c25e/odoo/addons/base/models/ir_sequence.py#L148
[2] https://github.com/odoo/odoo/blob/7b6c25e/addons/project_timesheet_holidays/models/res_company.py#L14
[3] https://github.com/OCA/project/blob/9d08f2d/project_task_code/models/project_task.py#L29
[4] https://github.com/odoo/odoo/blob/7b6c25e/odoo/addons/base/models/ir_sequence.py#L282

### DIFF
--- a/project_task_code/data/task_sequence.xml
+++ b/project_task_code/data/task_sequence.xml
@@ -8,5 +8,6 @@
         <field name="code">project.task</field>
         <field eval="4" name="padding" />
         <field name="prefix">T</field>
+        <field name="company_id" eval="False" />
     </record>
 </odoo>


rolandojduartem:
Hi, @bizzappdev and @luisg123v, please, could you review this PR?

cc @desdelinux

luisg123v (changes requested):
Code LGTM, just regarding commit description:
- NULL -> False
- it could happen installation issues -> there may be installation issues
- for another modules -> on other modules
- different of the company -> different from

I'd say:

```
[FIX] project_task_code: make sequence available for all companies

Currently, the sequence that generates task codes is only available for the
main company, which could cause issues....
```

I mean, technical explanation is OK, but I think it would be more clear if you start with the what and why before explaining the how.

rolandojduartem:
Hi, @luisg123v, I applied the changes, also, I added the migration script because the data is no update.

rolandojduartem:
Hi, @luisg123v, please review ths MR

luisg123v (approved):
LGTM :+1: 

@moylop260 could you review, please?

luisg123v (changes requested):
@rolandojduartem just perform a write in the migration, instead for reloading the file. This so that, if any other customizations has been done functionally (e.g. prefix or step) are not overwritten

rolandojduartem:
@luisg123v ready!

luisg123v (approved):
LGTM :+1: 

@moylop260 this time is for real :smile:

rolandojduartem:
Hi, @moylop260, could you review this PR?

moylop260:
> which could cause issues at the moment of the installation process 

If it is raising errors when the module is installed so it doesn't need a migration

Consider it could be a module already installed in production stable instances and the migration could make new issues

So, no migration needed here x2

rolandojduartem:
Hi, @moylop260, @yankinmax, @gurneyalex and @bizzappdev, please, could you review this PR?

project_task_code/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "Sequential Code for Tasks",
-    "version": "15.0.1.0.3",
+    "version": "15.0.1.0.4",
```
moylop260: The bot will auto-bump the version

So, it is not even needed here

rolandojduartem: Done!

rolandojduartem:
@moylop260 done!

moylop260 (approved):
👍

PR #1055, opened 2023-01-19
[15.0][MIG] project_key

vincent-hatakeyama:
Standard migration of the project_key module.


project_key/models/project_project.py
```diff
@@ -13,7 +13,7 @@ class Project(models.Model):
         comodel_name="ir.sequence", string="Key Sequence", ondelete="restrict"
     )
 
-    key = fields.Char(string="Key", size=10, required=False, index=True, copy=False)
+    key = fields.Char(size=10, required=False, index=True, copy=False)
```
zamberjo, line 16: Tanks for this @vincent-hatakeyama 

I only see a small error from a previous migration:
`Remove size=X attribute in Char fields, as it's no longer valid for restricting the size of the strings.`

see: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0

Functionally it is Ok for me

zamberjo (changes requested):

project_key/models/project_project.py
vincent-hatakeyama: That’s strange because in Odoo 14, size is used on the attachment’s checksum : https://github.com/odoo/odoo/blob/14.0/odoo/addons/base/models/ir_attachment.py#L393
In the database, the type of the column is `character varying(40)`.
(the url is also defined as size 1024 and the information is found on the database too)

That’s still the case with Odoo 15 and 16.

vincent-hatakeyama:
I’ve added the fix from #1072 as suggested.

project_key/models/project_project.py
zamberjo: You can see it here:  https://github.com/odoo/odoo/blob/14.0/odoo/fields.py#L1624

You can leave it for me, it can be removed in the next version.

zamberjo (approved):

jcfernandez-890825:
@vincent-hatakeyama @pedrobaeza 
Is this module gonna be migrated to v15 and v16 ?

zamberjo:
> @vincent-hatakeyama @pedrobaeza Is this module gonna be migrated to v15 and v16 ?

See https://github.com/OCA/project/pull/1169

jcfernandez-890825:
@zamberjo 

Thanks for getting me the code for v16.
So this module is just pending to be merged on v15 and v16?

vincent-hatakeyama:
> Is this module gonna be migrated to v15 and v16 ?

This is the migration for Odoo 15; it just hasn’t been merged in OCA.

vincent-hatakeyama:
I’ve just rebase this.

@OCA/project-service-maintainers It would be nice if this was merged since the 16.0 version is also merged.

dreispt:
This PR can't be reopened.
Can you create a new one please?

PR #1061, opened 2023-02-08
[MIG] project_purchase_link: Migration to 16.

AnizR:
Migration of module project_purchase_link

Some modifications had to be done due to the fact that we now have analytic_distribution instead of just an anlaytic_account.


pedrobaeza:
Please squash together these commits:

Selección_071

AnizR:
> Please squash together these commits:
> 
> Selección_071

Done

project_purchase_link/models/project_project.py
```diff
@@ -23,84 +21,205 @@ class ProjectProject(models.Model):
     )
 
     def _compute_purchase_info(self):
+        # read group
+        query = self.env["purchase.order.line"]._search(
```
pedrobaeza: Why removing the `read_group`? I don't think the complexity of the code deserves the few juice on the performance (if any).

AnizR: In fact, since now on a PO line we have an analytic_distribution (and an analytic distribution is a json such as:
{analytic_account_id1:rate1, analytic_account_id2:rate2...}). 
I think that it is not possible to use a read group as it was done in previous versions because I am trying to find PO lines that have the analytic account of the project within their analytic distribution:
`query.add_where(
            "purchase_order_line.analytic_distribution ?| array[%s]",
            [str(project.analytic_account_id.id) for project in self],
        )`
        
 But correct me if I am wrong

pedrobaeza: OK, I see... Everything gets complicated with the new analytic system. I let users of this module to judge, as I don't use it and only doing some administrative collaboration.

acsonefho (approved):

acsonefho:
@pedrobaeza ~~can you retry your merge please?~~ :pray: 
Sorry I miss read your previous command.

Do you have super-power to merge this please?

acsonefho (approved):

pedrobaeza:
There's a missing review, but well, I think it's enough with the arguments given and your review:

/ocabot merge nobump

PR #1062, opened 2023-02-09
[13.0][FIX] project_stock_product_set: Check wizard with user in tests

victoralmau:
Check wizard with user in tests.

@Tecnativa TT41461

### DIFF
--- a/project_stock_product_set/models/product_set_line.py
+++ b/project_stock_product_set/models/product_set_line.py
@@ -9,6 +9,7 @@ class ProductSetLine(models.Model):
     def prepare_stock_move_values(self, task, quantity):
         self.ensure_one()
         return {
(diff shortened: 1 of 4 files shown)


CarlosRoca13 (approved):

project_stock_product_set/models/product_set_line.py
```diff
@@ -9,6 +9,7 @@ class ProductSetLine(models.Model):
     def prepare_stock_move_values(self, task, quantity):
         self.ensure_one()
         return {
+            "name": self.product_id.name,
```
pedrobaeza: I would say it's:

```suggestion
            "name": self.product_id.display_name,
```

pedrobaeza (approved):
/ocabot merge patch

PR #1069, opened 2023-02-14
[14.0][MIG] project_wbs

JasminSForgeFlow:
Standard Migration

@ForgeFlow


AaronHForgeFlow (commented):
Looks good overall, can you remove the strings from the buttons in the WBS list? 

image

JasminSForgeFlow:
> Looks good overall, can you remove the strings from the buttons in the WBS list?
> 
> image

Done, Thanks

AaronHForgeFlow (commented):
Creating subprojects from the list is good. However, When using the button there is no way to change the code, resulting in this error. There must be something else, because the subproject created, even if it has the same code, it should not have the same complete_wbs_code. 

project_wbs

JasminSForgeFlow:
> Creating subprojects from the list is good. However, When using the button there is no way to change the code, resulting in this error. There must be something else, because the subproject created, even if it has the same code, it should not have the same complete_wbs_code.
> 
> project_wbs [ project_wbs ](https://user-images.githubusercontent.com/19620251/218730847-8bf779e2-3a83-4d6f-ba46-bef09b350056.gif) [ ](https://user-images.githubusercontent.com/19620251/218730847-8bf779e2-3a83-4d6f-ba46-bef09b350056.gif)

Fixed.

AaronHForgeFlow (approved):
Functional review LG!

JuanyDForgeflow (approved):
LGTM

PR #1080, opened 2023-03-05
[WIP][15.0][MIG]project_gtd: Migration to 15.0

bizzappdev:


pedrobaeza:
/ocabot migration project_gtd

@bizzappdev thanks for this massive migration pull requests wave. Unfortunately, the pull request is only the first step of the OCA workflow. As you may know, there should be 2 reviews and one for a PSC/maintainer for merging them. Maybe you can do cross reviews for asking other contributors to review your PRs, or inside your team, do as well cross-reviews. The problem is that if you put everything under the same GitHub login, it's not possible technically nor trustable. Do you think there can be any solution?

bizzappdev:
> 
@pedrobaeza thank you for taking care of updating all PRS.
We are also planning to do the Review by the team soon.

I like to differ for if everything is from one account "it's not possible technically nor trustable" 
technically we did an internal code sprint in December and after that every now and then whoever was free they were checking and fixing things or testing the migrated code. once it is ready we do a code review internally and check the quality for the same and then testing of the functionality after that only we create PR on the OCA.
We have chosen this because this will give flexibility to us and we can also have a watch over the quality.
Also, not everyone is checking the progress or do communicate on the PR. so from one account I can at least do the communication and keep watch over the changes or updates.

let me know if this does not fall into the technical or trustable part. we will try to limit the PR creation from one account.

pedrobaeza:
We can do the same using an organization (Tecnativa), but pushing from individual commits.

The technical part is that from the same account that you do the PR, you can't make a review (try it and tell me). The trustable part is that if anybody sees an approval from the same account, how do they know that is a second person who is reviewing it instead the same one?

PR #1082, opened 2023-03-06
[14.0][FIX] project_timeline_hr_timesheet: Set the correct class in all cases

victoralmau:
Set the correct class in all cases.
ejemplo

Please @pedrobaeza and  @CarlosRoca13 can you review it?

@Tecnativa TT40299

### DIFF
--- a/project_timeline_hr_timesheet/views/project_task_view.xml
+++ b/project_timeline_hr_timesheet/views/project_task_view.xml
@@ -14,7 +14,7 @@
             <xpath expr="//small[@name='planned_hours']" position="after">
                 <div class="o_project_timeline_hr_timesheet_progress">
                     <div
-                        t-att-class="record.remaining_hours &lt; 0 and 'danger' or ''"
+                        t-att-class="record.remaining_hours &lt; 0 ? 'danger ' : ' '"
                         t-attf-style="width: #{record.progress}%"
                     />
                 </div>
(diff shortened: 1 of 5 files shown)


project_timeline_hr_timesheet/views/project_task_view.xml
```diff
@@ -14,7 +14,13 @@
             <xpath expr="//small[@name='planned_hours']" position="after">
                 <div class="o_project_timeline_hr_timesheet_progress">
                     <div
-                        t-att-class="record.remaining_hours &lt; 0 and 'danger' or ''"
+                        t-if="record.remaining_hours &lt; 0"
```
pedrobaeza: I don't think you need this, just put the other as `t-attf-class`

victoralmau: I have tried to make that change but it does not work in this case.
ejemplo-1

pedrobaeza: It's weird, as the Qweb should do it:

https://www.odoo.com/documentation/14.0/es/developer/reference/javascript/qweb.html

Maybe it's not being rendered properly? Anyway, if not possible, just drop a comment here to know why it's done that way. Please always remember to document each weird thing for not questioning each time why is there.

victoralmau: I finally found out how to do it in one line.

CarlosRoca13 (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1083, opened 2023-03-08
[13.0][FIX] project_stock_product_set: Set group_id allways in new stock moves from wizard

victoralmau:
Set group_id allways in new stock moves from wizard

This problem exists since https://github.com/OCA/project/commit/1b8a63e75ab168460e28679b585d3f0f586f0cbb

@Tecnativa TT42000

### DIFF
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -1,4 +1,4 @@
-# Copyright 2022 Tecnativa - Víctor Martínez
+# Copyright 2022-2023 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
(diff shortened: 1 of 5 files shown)


project_stock/models/project_task.py
```diff
@@ -123,6 +123,14 @@ def _compute_unreserve_visible(self):
             any_quantity_done = any([m.quantity_done > 0 for m in item.move_ids])
             item.unreserve_visible = not any_quantity_done and already_reserved
 
+    def _set_group_id(self):
```
pedrobaeza: Remember that you are adding this in a general model, so the naming of the methods should be self-explanatory as possible:

```suggestion
    def _set_procurement_group_id(self):
```

project_stock_product_set/models/product_set_line.py
```diff
@@ -7,7 +7,7 @@ class ProductSetLine(models.Model):
     _inherit = "product.set.line"
 
     def prepare_stock_move_values(self, task, quantity):
-        self.ensure_one()
+        task._set_group_id()  # We make sure it is defined
```
pedrobaeza: Please do this only once, not one per line.

pedrobaeza (changes requested):

pedrobaeza (approved):

victoralmau:
Ping @CarlosRoca13

CarlosRoca13 (approved):

PR #1013, opened 2022-11-02
16.0 mig project task code

uncannycs:

PR #1084, opened 2023-03-09
[16.0][MIG] project_task_code

Abranes:
Supersedes #1013


rafaelbn:
Hello @Abranes 

Please this module in 15 (at least) is creating constraint that breaks runboat with other modules. could you FIX it here in migration?

@Shide  
> Yes, project_task_code module is creating a constraint on project.task and other modules who creates tasks (by strange ways or on a different order) breaks the init of the ruboat with demo data.

@sbidoul 
> Ah, yes. IMO modules that add constraints on standard models should enable them by configuration, otherwise we reach dead ends like this one.

https://github.com/OCA/project/pull/1096#issuecomment-1484977104

Thank you!

Abranes:
> Hello @Abranes
> 
> Please this module in 15 (at least) is creating constraint that breaks runboat with other modules. could you FIX it here in migration?
> 
> @Shide
> 
> > Yes, project_task_code module is creating a constraint on project.task and other modules who creates tasks (by strange ways or on a different order) breaks the init of the ruboat with demo data.
> 
> @sbidoul
> 
> > Ah, yes. IMO modules that add constraints on standard models should enable them by configuration, otherwise we reach dead ends like this one.
> 
> [#1096 (comment)](https://github.com/OCA/project/pull/1096#issuecomment-1484977104)
> 
> Thank you!

Done! You can review it.

project_task_code/hooks.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def pre_init_hook(cr):
+    """
+    With this pre-init-hook we want to avoid error when creating the UNIQUE
+    code constraint when the module is installed and before the post-init-hook
+    is launched.
+    """
+    cr.execute("ALTER TABLE project_task ADD COLUMN code character varying;")
+    cr.execute("UPDATE project_task SET code = id;")
```
Shide: Should be done only when the task code is activated? It's deactivated by default.
Same on `post_init_hook`

project_task_code/models/project_task.py
```diff
@@ -0,0 +1,51 @@
...
+        if not project_task_unique_code:
+            return
+
+        for task in self:
+            if self.search([("code", "=", task.code), ("id", "!=", task.id)], limit=1):
+                raise ValidationError(_("The code must be unique!"))
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
```
Shide, line 39: Should be done if user has activated the corresponding parameter in `res.config.settings` and sequence exists in `self.env.company`.

project_task_code/models/res_config_settings.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    project_task_unique_code = fields.Boolean(
+        string="Unique Task Code",
+        config_parameter="project_task_code.project_task_unique_code",
```
Shide: I think could be a `ir.sequence` record (for multi company purposes)

project_task_code/models/project_task.py
```diff
@@ -0,0 +1,51 @@
...
+        string="Task Number",
+        required=True,
+        default="/",
+        readonly=True,
+        copy=False,
+    )
+
+    @api.constrains("code")
+    def _check_project_task_unique_code(self):
+        project_task_unique_code = (
```
Shide: No need to check for parameter. This should pass even if the parameter is not set. Only check it if task has code.

Shide (changes requested):
Code is being set even if in the settings, the Task Code Unique is not checked.
Also check the `pre_init_hook` and `post_init_hook`

rousseldenis:
@Abranes

project_task_code/views/project_view.xml
```diff
@@ -0,0 +1,46 @@
...
+                <field name="code" />
+            </field>
+        </field>
+    </record>
+    <record id="project_task_code_kanban_view" model="ir.ui.view">
+        <field name="name">project.task.code.kanban</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.view_task_kanban" />
+        <field name="arch" type="xml">
+            <field name="name" position="before">
```
jort-aardug: Field "name" exists twice in the kanban view, once for non-active tasks (shows with strikethrough), once for active tasks. This field selector only targets the first one, so you only see the code field on archived tasks.

yaazkal:
Is there something pending here still?

Shide:
> Is there something pending here still?

Check @jort-aardug and @Shide suggestions

project_task_code/models/project_task.py
robinkeunen: I think there is a confusion here in the use of the parameter : 

- the task code is generated when the module is installed
- the parameter adds the check for unicity

robinkeunen (approved):

yaazkal:
@Abranes can you check the suggestions for this to be merged? Thanks

project_task_code/views/project_view.xml
robinkeunen: I tried fixing it but it does it actually does not interact well with the `name_with_subtask_count` widget : 

- there is no space between code and name fields
- only the code in stroke trough when the task is archived.

Screenshot 2023-10-26 at 16 38 52
Screenshot 2023-10-26 at 16 39 17

robinkeunen: This
```xml
<xpath expr="//field[@name='name']/.." position="before">
    <field name="code" class="me-2"/>
</xpath>
```
produces

Screenshot 2023-10-26 at 17 17 39


For both archived and unarchived tasks. The `s` tag is overiden by the `name_with_subtask_count` widget.

robinkeunen:
I made a suggestion for the _name tag_ comment. If I understand correctly, the `project_task_unique_code` parameter adds the check for unicity, it does not activate the feature. If so, @Shide 's comment do not apply.

I can do a superseeding PR if @Abranes or maintainers give me the go.

Abranes:
Sorry guys, I forgot this PR.

I'll try to address all points next week.

project_task_code/hooks.py
Abranes: Removed, not needed because it's deactivated by default.

project_task_code/models/res_config_settings.py
Abranes: Done

project_task_code/models/project_task.py
Abranes: Changed because now the code is only set if the sequence is set in configuration, so now the constrains check if the task has code and if any other task in the company has the same code.

project_task_code/views/project_view.xml
Abranes: Done

Abranes:
@jort-aardug @Shide @robinkeunen @yaazkal could you check it one more time? :)

Shide (approved):
**praise:** Good job with this man :)

Code LGTM

robinkeunen (changes requested):
I think there is a confusion here, the module should

- add the task code upon installation
- force the unicity constraint on configuration

The last fixes does the opposite:

- force the unicity constraint upon installation
- create the task code on configuration

Abranes:
> I think there is a confusion here, the module should
> 
> * add the task code upon installation
> * force the unicity constraint on configuration
> 
> The last fixes does the opposite:
> 
> * force the unicity constraint upon installation
> * create the task code on configuration

It's what @Shide requests and I think it's not a bad approach.

robinkeunen:
But that changes the purpose of the module, migration is not the good time to do that. Plus a better module name would then me `project_task_code_unique`.

Plus it does not follow @sbidoul and @rafaelbn 's recommendation. Let's see what others think.

robinkeunen:
Hi @rousseldenis  @dreispt , we're disagreeing on the way to go with this PR. Can you help us move forward ?

Shide:
> But that changes the purpose of the module, migration is not the good time to do that. Plus a better module name would then me `project_task_code_unique`.
> 
> Plus it does not follow @sbidoul and @rafaelbn 's recommendation. Let's see what others think.

Maybe we should add a migration script?
 
The module without this fix/feature will break tests for sure at any point in the future... If someone has the bad luck of having this module installed in another way.

rafaelbn:
Hello,

I don't know in this point which is the disagree. I pinged @Abranes because this module is breaking other modules test, etc... and we said to FIX it here.

> Hello @Abranes
> 
> Please this module in 15 (at least) is creating constraint that breaks runboat with other modules. could you FIX it here in migration?
> 
> @Shide
> 
>     Yes, project_task_code module is creating a constraint on project.task and other modules who creates tasks (by strange ways or on a different order) breaks the init of the ruboat with demo data.
> 
> @sbidoul
> 
>     Ah, yes. IMO modules that add constraints on standard models should enable them by configuration, otherwise we reach dead ends like this one.
> 
> https://github.com/OCA/project/pull/1096#issuecomment-1484977104
> 
> Thank you!

- [ ] The request was to address a constraint issue in the project_task_code module for version 15. The module's constraint on project.task is causing conflicts with other modules in runboat environments, especially those creating tasks in unconventional ways. @sbidoul suggests that modules adding constraints to standard models should have the option to enable these constraints through configuration to avoid such conflicts. This suggestion aims to prevent dead ends and improve module compatibility.

We are not using this module at all, just we are affected in other modules un runboat and test because this one.

@robinkeunen > But that changes the purpose of the module
@Shide > The module without this fix/feature will break tests for sure at any point in the future... If someone has the bad luck of having this module installed in another way.

- The purpose of the module shouln't be change but FIXED

Maybe @oihane or @pedrobaeza as original contributors what to put an opinion. I don't know exacly what's the problem now. @dreispt Do you have a solution here o some past use case?

Thank you all! 😄 ❤️

pedrobaeza:
Can you summarize instead of copying a lot of texts? If I understand correctly, there was a constraint and it has been removed in the migration? Or the contrary? Which is the problem in runboat?

robinkeunen:
In 15.0 : `project_task_code`

- adds `code` field on tasks
- adds sql contraint `project_task_unique_code`
- a hook initializes code from `project.task` sequence
- in `create`, code is assigned from `project.task` sequence

In his migration review (2023-03-27), @rafaelbn pointed out that the a configuration should be added for the constraint. Otherwise, it breaks the tests from other modules.

I understand from the suggestion that the module should

- add the task code upon installation
- force the unicity constraint on configuration

The fix from @abranes, based on @shide's request, does the opposite, it 

- forces the unicity constraint upon installation
- creates the task code on configuration

pedrobaeza:
Thanks for the summary, Robin.

It's true that enforcing a uniqueness can be conflicting for prior version databases not containing this constraint migrating to this version. If implemented via SQL constraint, Postgres will try to put the constraint and fail, not having the constraint at all, but only a warning each time you update. Doing that by Python constraint allows to be enabled by configuration, and this would be a saner feature.

Another conflicting change is to fill existing taxes with a code, as you may not want to have one at all. I would go for having a button on configuration to fill the code on "empty code" taxes on demand.

Do you see it very complicated to be done?

huguesdk:
i don’t understand why so many changes are introduced in a simple version migration. this sql constraint is present since version 8. the hooks that run upon module initialization ensure that this constraint is satisfied, and that tasks created subsequently (even by other modules) will have a unique code (coming from a sequence). it shouldn’t cause an issue.

i don’t think it’s a good idea to introduce such significant changes as part of a version migration, unless something that changed in that version prevents the module from working as it did in the previous version.

i think that the changes in commit https://github.com/OCA/project/commit/90103ce1cdf1520131997f6e6de11396cf2351ac are enough.

robinkeunen:
Rather than arguing, I propose this PR #1209. I leave it to the PSC to choose what is more appropriate :-)

huguesdk:
could someone give a more detailed explanation about what the problem that we’re trying to solve exactly is (apart from porting this module to 16)? i’ve read that it caused errors during the runboat build. could someone copy the exact error here to see if it can be solved in a simpler way?

note that the original sql constraint is not multi-company compatible (it requires uniqueness of the code while it should probably be unique per company) while sequences are per company by default. maybe this could be the issue?

legalsylvain:
Hi @rafaelbn 

> Please this module in 15 (at least) is creating constraint that breaks runboat with other modules. could you FIX it here in migration?

could you mention the error generated ? I don't understand how it can break things !

Thanks !

Abranes:
Hi all! I can make the changes you want but first let's all be clear, because in the first instance I perform a normal migration, then modify the module so that the constraint was not by SQL and was by python and thus be able to activate or deactivate it in the configuration to avoid problems with other modules, and finally carry out the current implementation, all at the request of different people.

How do you prefer it to proceed?

huguesdk:
i personally think that the sql constraint should stay. it is the only way to ensure uniqueness in a reliable way. checking it in python won’t prevent the same code to be used if records are created at the same time. we need to check what problem the sql constraint actually caused, and fix it without removing the constraint.

rafaelbn:
Thinking in an OCA Repo with more modules involved:

IMO modules that add constraints on standard models should enable them by configuration, otherwise we reach dead ends. So not by SQL but by python and thus be able to activate or deactivate it in the configuration to avoid problems with other modules

Not blocking anything, just my opinion.

Thanks all by efforts in the best way! 😄 ❤️

legalsylvain:
> IMO modules that add constraints on standard models should enable them by configuration, otherwise we reach dead ends. So not by SQL but by python and thus be able to activate or deactivate it in the configuration to avoid problems with other modules

Hi. my opinion is "Keep It simple". An SQL constraint is way more simple as constraint enabled by configuration. So, if there is no existing problem, I propose to keep the V15 code. 

https://github.com/OCA/project/blob/15.0/project_task_code/models/project_task.py#L18-L20 is simpler (**2 lines**) than : 
- https://github.com/OCA/project/pull/1084/files#diff-c6bf6da19dd508004fa0507f6c93818e4d02c9cc2ccf5a8ef3b86482959e59d2R37-R63 (**37 lines**)
- https://github.com/OCA/project/pull/1084/files#diff-5cd651cf3b3f58da9ee3f1c78edea6215d0464137e4e03300db36cddb4edf6faR1-R26 (**26 lines**)
- https://github.com/OCA/project/pull/1084/files#diff-b66987146c83cf675c185c5bf63cdf252472583bd56acd42a61bc8701c5b26afR1-R14 (**14 lines**).

@rafaelbn : you was saying that V15 runboat was broken, but I see it green. Is it fixed ? https://runboat.odoo-community.org/webui/builds.html?repo=OCA/project&target_branch=15.0

rafaelbn:
Hello @legalsylvain! Happy new year 2024! 😄 

@Shide will explain after holidays the problem that this modules causes when living in the same repo with other ones. 

We are not using this module at all, we just found problems when working in others in the same repo.

Thank you!

Shide:
@legalsylvain the case which i'm saying this needs a change is because this happens to me in this PR: https://github.com/OCA/project/pull/1096#issuecomment-1484977104

Check out the next 4 comments where i'd invoked you because the runboat breaks

legalsylvain:
Hi @Shide. Thanks a lot for pointing the issue. @robinkeunen, could you take a look at this issue ? 
thanks ! 
and happy new years !

rafaelbn:
This PR is continued in https://github.com/OCA/project/pull/1209

PR #1092, opened 2023-03-18
[15.0][FIX][project_timesheet_time_control] Error: AttributeError: type object 'expression' has no attribute AND

ignaciord:
It seems that "expression" is not correctly imported from odoo.osv. 

This causes the error:

"AttributeError: type object 'expression' has no attribute 'AND'" when this code block is executed:

`            return expression.AND(
                [
                    domain,
                    [
                        "|",
                        ("privacy_visibility", "!=", "followers"),
                        ("message_partner_ids", "in", [self.env.user.partner_id.id]),
                    ],
                ]
            )`

### DIFF
--- a/project_timesheet_time_control/wizards/hr_timesheet_switch.py
+++ b/project_timesheet_time_control/wizards/hr_timesheet_switch.py
@@ -3,7 +3,7 @@
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
-from odoo.osv.expression import expression
+from odoo.osv import expression
 
 
 class HrTimesheetSwitch(models.TransientModel):


pedrobaeza:
Is this problem reproducible in runboat (accessible from "Runbot | Try me" button in the readme)?

If not, the problem is in your deployment method, mangling import paths.

ignaciord:
> Is this problem reproducible in runboat (accessible from "Runbot | Try me" button in the readme)?
> 
> If not, the problem is in your deployment method, mangling import paths.

Yes, the problem is reproducible in runboat.

These are the steps to reach the error:

1) Log in with user with no "group_timesheet_manager" privileges (for example user "Marc Demo" with login name "demo").

2) Go to Project Module and try to start a task, for example in project named "Office Design" (the first one).

3) The wizard does not launch and and "Odoo server error" window shows this error:

`  File "/opt/odoo/odoo/fields.py", line 2659, in _description_domain
    return self.domain(env[self.model_name]) if callable(self.domain) else self.domain
  File "/mnt/data/odoo-addons-dir/project_timesheet_time_control/wizards/hr_timesheet_switch.py", line 16, in _domain_project_id
    return expression.AND(
Exception

The above exception was the direct cause of the following exception:

  File "/opt/odoo/odoo/http.py", line 643, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/odoo/http.py", line 301, in _handle_exception
    raise exception.with_traceback(None) from new_cause
AttributeError: type object 'expression' has no attribute 'AND'`

pedrobaeza (approved):
Curious that this works for timesheet managers, but not timesheet users.

Let's fast-track it to fix the problem:

/ocabot merge patch

ignaciord:
> Curious that this works for timesheet managers, but not timesheet users.
> 
> Let's fast-track it to fix the problem:
> 
> /ocabot merge patch

I think that works for timesheet managers but not timesheet users becouse the expression `return expression.AND(` is only evaluated when this condition is true: `if not self.user_has_groups("hr_timesheet.group_timesheet_manager"):`

Thanks a lot Pedro for the fast merge.

PR #1096, opened 2023-03-24
[15.0][ADD] project_fields_sequence

Shide:
Add a new module to be able to fill project fields based on sequences when creating a new one.

Define a sequence starting with `project.field.` and concatenate a (technical) field name to it.
Eg: `project.field.name`, `project.field.alias_name`

Field names must be **Character** types.

This will return the next value of the sequence when the project is created (and `default_get` is called)

MT-1506 @moduon  @rafaelbn @yajo @anddago78 please review if you want :)


Shide:
@sbidoul Hey! Do you know how to fix the runboat build? I've tried to mark `project_task_code` module as **rebel**, but it's failing when loading demo data (i suppose).

sbidoul:
I don't know. It does not look like a runboat issue. It's either an Odoo bug, or an incompatibility with a module of this repo?

Shide:
Yes, `project_task_code` module is creating a constraint on `project.task` and other modules (not this one) who creates tasks (by strange ways or on a different order) breaks the init of the ruboat with demo data.

sbidoul:
Ah, yes. IMO modules that add constraints on standard models should enable them by configuration, otherwise we reach dead ends like this one.

yajo:
I asked @Shide declared the rebel module groups, so tests go ✅ now. The only part that's missing is runboat.

I'm thinking that maybe runboat should be able to boot always, no matter if the build failed. After all, if installation fails because of these kind of problems, one could always go to the runboat and create an empty database, or uninstall the rebel module by hand. Does that make sense to you @sbidoul?

project_fields_sequence/models/project_project.py
```diff
@@ -0,0 +1,31 @@
...
+        res = super().default_get(fields_list)
+        is_model = self.env["ir.sequence"].sudo()
+        fields_with_seq = [
+            s.code[len(PROJECT_FIELDS_SEQUENCE_CODE_BASE) :]
+            for s in is_model.search(
+                [
+                    ("code", "ilike", PROJECT_FIELDS_SEQUENCE_CODE_BASE + "%"),
+                ]
+            )
+        ]
```
yajo: This will improve performance, readability and accuracy:

```suggestion
        fields_with_seq = {
            seq.code[len(PROJECT_FIELDS_SEQUENCE_CODE_BASE) :]
            for seq in is_model.search(
                [
                    ("code", "=like", PROJECT_FIELDS_SEQUENCE_CODE_BASE + "%"),
                ]
            )
        }
```

project_fields_sequence/static/description/icon.svg
yajo, line 1: It is better to remove the SVG if it's gonna be the default one for OCA.

The purpose of the SVG version is to be able to edit the icon later, but this SVG is already available elsewhere.

yajo (approved):
Code review.

yajo:
Superseded by https://github.com/OCA/server-tools/pull/2598

PR #1106, opened 2023-04-17
[FIX] project_forecast_line issues

gurneyalex:
- [FIX] forecast_line: issue in multicompany leading to wrong forecast
- [FIX] wrong condition preventing project.task quick update path to be used
- [FIX] wrong result when project.task quick update path is used (inversion of the sign of the task forecast)
- [FIX] crash when a task is not related to a project


yankinmax (approved):

thomaspaulb (approved):
Time to merge this?

PR #1108, opened 2023-04-28
[15.0][IMP] project_timesheet_time_control: Add base.group_multi_company group to company_id field from wizard

victoralmau:
Add `base.group_multi_company` group to `company_id` field from wizard (similar to 14.0: https://github.com/odoo/odoo/blob/14.0/addons/hr_timesheet/views/hr_timesheet_views.xml#L93)

@Tecnativa


pedrobaeza (approved):
Fast-tracking as straight-forward:

/ocabot merge minor

PR #1109, opened 2023-05-09
[16.0][mig] project_status: Migration to 16.0

dsolanki-initos:


pedrobaeza:
In theory Odoo already includes project stages in the core in this version, which is very similar to this module. Is there any specific reason to keep this one?

dsolanki-initos:
> In theory Odoo already includes project stages in the core in this version, which is very similar to this module. Is there any specific reason to keep this one?

@pedrobaeza, It's almost similar to project stages but It has two more fields named "is_closed" which used to specify this is a closing status and "description".

pedrobaeza:
But then it's convenient to just add such fields to existing model instead of adding the whole stack, isn't it?

dsolanki-initos:
> But then it's convenient to just add such fields to existing model instead of adding the whole stack, isn't it?

Yes, Agreed
Shall I add those two fields to the existing model and do the changes in this PR?

pedrobaeza:
I would create a new module `project_stage_extra_info` or similar for both fields (or more to come).

dsolanki-initos:
@pedrobaeza, Okay, I'll create a new module named `project_stage_extra_info` and create a new PR to refer to this PR.

dsolanki-initos:
Closing this PR and as per our discussion, I've created another PR that adds extra fields to the existing model. Ref. https://github.com/OCA/project/pull/1113

PR #1111, opened 2023-05-09
[16.0][mig] project_task_pull_request: Migration to 16.0

dsolanki-initos:


flotho (changes requested):
tested on runbaot, functionnal OK
small UI imporvements :
image

PR #1112, opened 2023-05-12
[IMP] project_sequence: configure display name pattern

yajo:
To change the project display name pattern, follow these steps:

1. Go to *Project > Configuration > Settings*.
2. Edit the *Project display name pattern* field.

   The default format is ``%(sequence_code)s - %(name)s``. You can use those same placeholders to customize the pattern.

@moduon MT-1506


fcvalgar (approved):
LGTM, Thank you @yajo

PR #1118, opened 2023-05-22
[IMP] project_role: inherit create in batch

len-foss:


SkiBY (approved):

astirpe (approved):
Thank you for this PR! The fix looks good.

I just want to propose two minor improvements, but they are not blocking, so I approve already.

1. I would rename `_fill_create_assignment_values()` to something more specific for this module, something like: `_project_role_create_assignment_values()`
2. For improving performance, I would move the `browse(values["company_id"])` from inside the method to outside, so that the browse will retrieve all the companies (eg.: `companies = ... .browse(values["company_ids"])`) and then inside the method just filter to get the correct company (eg.: `company = companies.filtered(lambda c: c.id == values["company_id"])`)

len-foss:
@astirpe Thank you, I updated it in that sense since both remarks make good sense. I don't think the method is much more difficult to override this way so it should be fine.

astirpe:
Can this one be merged?

auroraeosrose:
I'm also waiting on this - looks good to me!

ramiadavid:
@dreispt Looks like it hasn't been merged

PR #1119, opened 2023-05-22
[ADD] [16.0] Project Portal Custom Field Access

ivantodorovich:
Odoo has a feature called "Project Sharing" that allows portal users to view a project's kanban, which has functionalities similar to the backend system.

There's a preset list of fields that portal users can access, which is harcoded in the module's source code.

That's nice but inconvinient if a user introduces new fields to the form and wants to share these with the portal users. They can't currently do this without using a Python override.

To resolve this, the module lets users whitelist these custom fields directly through the ``ir.model.field`` model.

### DIFF
--- a/project_portal_custom_field_access/README.rst
+++ b/project_portal_custom_field_access/README.rst
@@ -1,0 +1,1 @@
+# TO BE GENERATED
--- a/project_portal_custom_field_access/__init__.py
+++ b/project_portal_custom_field_access/__init__.py
@@ -1,0 +1,1 @@
+from . import models
--- a/project_portal_custom_field_access/__manifest__.py
+++ b/project_portal_custom_field_access/__manifest__.py
@@ -1,0 +1,15 @@
+# Copyright 2023 Camptocamp SA (https://www.camptocamp.com).
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Project Portal Custom Field Access",
+    "summary": "Grant the Portal User access to custom Project Task fields",
+    "version": "16.0.1.0.0",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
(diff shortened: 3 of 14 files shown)


Aldeigja:
Hi @ivantodorovich ,

We encountered the same issue while troubleshooting Odoo 16 portal with some modules from OCA/project installed and projects shared via 'Project Sharing.'

I'm curious to understand why the work on this PR was stopped. Do you have any objections if we proceed with continuing the work you started in this PR?

ivantodorovich:
Hello @Aldeigja 

In fact the work on this PR was not stopped. The module is finished and working, we're using it in one of our projects.

The PR was tagged as stale and later closed automatically by github actions because it didn't get any reviews in several months. Sometimes it happens that contributions like this don't raise enough interest in the community, and so they're never merged

AungKoKoLin1997:
@ivantodorovich 
Thanks for the module. It will be very useful when we have custom fields for project task. 
I tried to check your module and I found this feature is only for manual state field and can't use for the custom field that is added by custom module. If I am not wrong, the custom fields that are added by custom module define as a base for state.
So, this feature is currently available only when we created new field from frontend UI.
Is there any background or reason regarding this restriction?

ivantodorovich:
Hello @AungKoKoLin1997 

For fields added by custom modules you don't need this.

You just need to have your custom module override `SELF_WRITABLE_FIELDS` and hardcode your module field names there.

The only point of this module is to handle the case where fields are created from the UI. Since we can't override this method from the UI, we need a configurable solution like this.

AungKoKoLin1997:
@ivantodorovich Thanks for the explanation. I got your point. 
Can you recreate new PR? I will review it.

ivantodorovich:
Thanks for the interest! 
Here's the new PR 😉 
- https://github.com/OCA/project/pull/1229

PR #1134, opened 2023-06-06
[15.0][FIX] project_timesheet_time_control: Apply sudo to prevent acl errors with ir.model

victoralmau:
Apply sudo to prevent acl errors with `ir.model`

@Tecnativa


pedrobaeza (approved):
Fast-tracking this straight-forward patch:

/ocabot merge patch

PR #1140, opened 2023-06-07
[MIG] project_forecast_line: migration to 16

martiqubiq:


project_forecast_line/tests/test_forecast_line.py
```diff
@@ -538,7 +569,7 @@ def setUpClass(cls):
             # flush needed here to trigger the recomputation with the correct
             # frozen time (otherwise it is called by the test runner before the
             # tests, outside of the context manager.
-            cls.task.flush()
```
rousseldenis: IMHO, flush_recordset() is more correct (on all calls).

PR #1145, opened 2023-06-28
[FIX] [15.0] project_forecast_line: Fixed bug by adding hr.group_hr_user group 

thomaspaulb:
Fixed bug by adding `hr.group_hr_user ` group to field not available in hr.employee.public.

Basically same as [this](https://github.com/OCA/hr/pull/1261/files) and Odoo explains as [this](https://github.com/odoo/odoo/blob/c55badb8d70bdccfeb82bffb330d3f0b1b162ed1/addons/hr/models/hr_employee.py#L23).

If this is not done, it can lead to `ValueError: Invalid field 'role_ids' on model 'hr.employee.public'` for users with no `hr.employee` rights.

### DIFF
--- a/project_forecast_line/models/hr_employee.py
+++ b/project_forecast_line/models/hr_employee.py
@@ -15,9 +15,21 @@ class HrJob(models.Model):
 class HrEmployee(models.Model):
     _inherit = "hr.employee"
 
-    role_ids = fields.One2many("hr.employee.forecast.role", "employee_id")
+    # NB: these fields need to be marked as groups="hr.groups_hr_user",
+    # because otherwise if a user with less privileges tries to read
+    # hr.employee.public, they'll get an AccessError on these fields.
+    # More info at: https://github.com/odoo/odoo/blob/d21887008b1ac7
+    # 69bd91d24e972323ffe936391a/addons/hr/models/hr_employee.py#L22
+    role_ids = fields.One2many(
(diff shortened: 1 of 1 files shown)


gjotten:
@gurneyalex we ran into an ACL issue here.

gjotten (approved):

moitabenfdz (approved):

project_forecast_line/models/hr_employee.py
```diff
@@ -15,9 +15,16 @@ class HrJob(models.Model):
 class HrEmployee(models.Model):
     _inherit = "hr.employee"
 
-    role_ids = fields.One2many("hr.employee.forecast.role", "employee_id")
+    role_ids = fields.One2many(
```
yankinmax, line 23: thanks @thomaspaulb for the fix. Would be great if you can simply add a comment before these fields to explain why we need `groups="hr.group_hr_user`
Some link to the https://github.com/odoo/odoo/blob/d21887008b1ac769bd91d24e972323ffe936391a/addons/hr/models/hr_employee.py#L22

yankinmax (approved):
Small comment, thanks

project_forecast_line/models/hr_employee.py
thomaspaulb: Done

yankinmax:
@leemannd can you pls merge this PR?

PR #1149, opened 2023-07-10
[IMP] project_task_dependency: Search by recursive dependency

SirTakobi:
This is just the beginning for allowing to search on the many new _dependency/depending_ fields that are not stored.

Without this, when searching by `recursive_dependency_task_ids` I see the log:
> ERROR <db_name> odoo.osv.expression: Non-stored field project.task.recursive_dependency_task_ids cannot be searched.

This might be extended to the other fields and I hope someone else is maybe able to cover the other operators too.

I know this is not a complete solution but is probably better than an error in the log.

This should be also rather easy to forward-port in active versions, I'm sorry to propose this for `10.0` but I need it for this version.
If/When approved, I'll happily consider to forward-port :)


SirTakobi:
CI should be happy after https://github.com/OCA/project/pull/1150

PR #1150, opened 2023-07-11
Use Copier

SirTakobi:
See https://github.com/OCA/oca-addons-repo-template.

Currently failing until https://github.com/OCA/oca-addons-repo-template/pull/199, there are also things to be decided in https://github.com/OCA/oca-addons-repo-template/issues/200


rousseldenis:
@SirTakobi Could you update this ?

PR #1158, opened 2023-07-24
[16.0][ADD] project_task_name_with_id

AungKoKoLin1997:
This module introduces improvements to our project task management. Now, users can locate tasks using not only the task name, but also the unique task ID. Furthermore, to make tasks easily distinguishable, each task's display name will now include its unique ID.

@qrtl


yostashiro (approved):
Code and functional review. 👍 

image

yostashiro:
The usage overlaps with that of `project_task_code` https://github.com/OCA/project/pull/1084, but this module is simpler for those who do not need a specific sequence rule.

francesco-ooops:
@AungKoKoLin1997 could you add an USAGE file?

AungKoKoLin1997:
> could you add an USAGE file?

Done!

dreispt:
I believe we already have a module for this need: `project_task_code`.
Would you consider migrating it to 16.0?

yostashiro:
@dreispt Please see this comment: https://github.com/OCA/project/pull/1158#issuecomment-1652799761

One reason I prefer this module over `project_task_code` is that `id` shows as the reference number in portal, therefore using `id` naturally makes the design more coherent with standard Odoo, IMO.

GauthierWillaerts (approved):

ivs-cetmix:
> @dreispt Please see this comment: [#1158 (comment)](https://github.com/OCA/project/pull/1158#issuecomment-1652799761)
> 
> One reason I prefer this module over `project_task_code` is that `id` shows as the reference number in portal, therefore using `id` naturally makes the design more coherent with standard Odoo, IMO.

@yostashiro thank you for a hint!
I think one of the things TODO in the project_task_code module will be to redefine portal URLs using code instead of id.

project_task_name_with_id/views/project_task_views.xml
```diff
@@ -0,0 +1,92 @@
...
+                    </s>
+                    <t t-else="">
+                        <field name="display_name" widget="name_with_subtask_count" />
+                    </t>
+                </strong>
+            </xpath>
+        </field>
+    </record>
+
+    <!-- Sharing views -->
```
yostashiro, line 38: @ivs-cetmix You may also want to pay attention to the sharing views in `project_task_code`, in case you invite external partners to collaborate on your projects.

yostashiro:
@AungKoKoLin1997 Can you please check if you can make `project_task_code` and `project_task_name_with_id` coexist without conflicts by making parts of the functions configurable (or just by tweaking tests)? I will have to seggregate these modules from each other otherwise.

yostashiro (commented):
Looks good. Can you please squash commits.

AungKoKoLin1997:
@dreispt Can you please try to merge again?

PR #1159, opened 2023-07-24
[ADD] project_task_tag

ajaniszewska-dev:


project_task_tags/views/project_task.xml
```diff
@@ -0,0 +1,18 @@
...
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.view_task_form2" />
+        <field name="arch" type="xml">
+            <field name="tag_ids" position="before">
+                <field name="allowed_tag_ids" widget="many2many_tags" invisible="1" />
+            </field>
+            <field name="tag_ids" position="attributes">
+                <attribute name="domain">[('id','in',allowed_tag_ids)]</attribute>
+                <!-- make sure there is no project key in context -->
+                <attribute name="context">{}</attribute>
```
TDu: Why, could this break some functionality ?

TDu (commented):
Good start, but it will needs some unit test. And I also think the module name could be improved to reflect better what the module does.

simahawk:
> And I also think the module name could be improved to reflect better what the module does.

and it should not end w/ "s" :wink:

project_task_tags/models/project.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import api, fields, models
+
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    allowed_tag_ids = fields.Many2many("project.tags", compute="_compute_task_tags")
```
simahawk:
```suggestion
    allowed_tag_ids = fields.Many2many("project.tags", compute="_compute_allowed_tag_ids")
```

project_task_tags/views/project_task.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_task_form2" model="ir.ui.view">
+        <field name="name">project.task.allowed_tags.form.inherit</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.view_task_form2" />
+        <field name="arch" type="xml">
+            <field name="tag_ids" position="before">
+                <field name="allowed_tag_ids" widget="many2many_tags" invisible="1" />
```
simahawk:
```suggestion
                <field name="allowed_tag_ids" invisible="1" />
```
I don't think the widget is needed

project_task_tags/views/project_task.xml
```diff
@@ -0,0 +1,18 @@
...
+            </field>
+            <field name="tag_ids" position="attributes">
+                <attribute name="domain">[('id','in',allowed_tag_ids)]</attribute>
```
simahawk: Can you try this (assuming `project_id` is in the view)?
```suggestion
                <attribute name="domain">['|', (not project_id and 1 or 0, '=', 1), ('project_ids', 'in', project_id)</attribute>
```
if it works you can drop the rest of the changes :wink:

simahawk (changes requested):

project_task_tags/views/project_task.xml
ajaniszewska-dev: it is used in name_search and seems returns all tags linked to proj: https://github.com/odoo/odoo/blob/16.0/addons/project/models/project.py#L2755

project_task_tag/tests/test_project_tags.py
```diff
@@ -0,0 +1,36 @@
...
+
+from odoo.tests.common import TransactionCase
+
+class TestProjectTags(TransactionCase):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.tag_1 = cls.env['project.tags'].create({'name':'tag1'})
+        cls.tag_2 = cls.env['project.tags'].create({'name':'tag2'})
```
ajaniszewska-dev: KeyError?

project_task_tags/views/project_task.xml
ajaniszewska-dev: yep, that seems to work fine on the view

project_task_tag/tests/test_project_tags.py
```diff
@@ -0,0 +1,33 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo.tests.common import TransactionCase
+
+
+class TestProjectTags(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.tag_1 = cls.env["project.tags"].create({"name": "tag1"})
```
simahawk: I'm not sure to understand what you are testing here :thinking: 
As the change now is only in the view, I would use `odoo.tests.Form` to make sure the modifiers of the tag_ids field are properly set.

ajaniszewska-dev: sorry, it was after yesterdays suggestion but it had more sense to have tests when field was computed, i don't think it is crucial to test domain on the field. Removed these tests.

project_task_tags/views/project_task.xml
simahawk: hmm so what's the point of this module? :thinking: 
Besides, that method is one of the ugliest I've seen in terms of clarity :sweat_smile:

ajaniszewska-dev: the point is to show only tags which are already chosen on parent project. OR all tags available if there is no project.
if we keep project_id in ctxt, doesnt matter what kind of domain is on the field - it always shows more/all tags as a possibility to choose from. So we need to remove proj_id key from context and set domain on field - to show only tags from linked project or all tags (either by computed field, which was here before, or domain proposed by you). Domain alone wont work.

leemannd:
From what I see the PR seems correct. @simahawk Do you have time to update your review?
@ajaniszewska-dev when the review is updated, could you sqash the commits together?

grindtildeath:
@simahawk Can you please update your review?

leemannd (approved):

PR #1164, opened 2023-07-28
[16.0][FIX] project_stock: Filter valid stock moves (avoiding those done and cancelled) in action_done() to avoid side effects.

victoralmau:
FWP from 15.0: https://github.com/OCA/project/pull/1163

Filter valid stock moves (avoiding those done and cancelled) in action_done() to avoid side effects.

Use case:
- Product A with 1 qty
- Confirm materials
- Edit Product A line to 0 qty
- Transfer Materials
- New line: Product B with 1 qty
- Confirm materials
- Transfer Materials (previously, the error https://github.com/odoo/odoo/blob/15.0/addons/stock/models/stock_move.py#L585 would be displayed.)

@Tecnativa TT44572

### DIFF
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -215,7 +215,10 @@ class ProjectTask(models.Model):
         return True
 
     def action_done(self):
-        for move in self.mapped("move_ids"):
+        # Filter valid stock moves (avoiding those done and cancelled).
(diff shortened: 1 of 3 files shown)


chienandalu (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1171, opened 2023-09-22
[16.0][ADD] project_stock_analytic_tag: New module

victoralmau:
New module (restore analytical tags in project tasks).

@Tecnativa


pedrobaeza (approved):
One doubt

pedrobaeza:
On `project_stock`, please add a migration script that checks if any analytic tag is set on the task moves, and if so, look for this module and put it with state = `to install`.

chienandalu (approved):

PR #1190, opened 2023-11-07
[15.0][MIG] project_stage_state: Migration to 15.0

nwinkelstraeter:


mathiasfrancke:
#916 Duplicate Migration. New name is "project_task_stage_state".  See: https://github.com/OCA/project/pull/916#issuecomment-1109144839

PR #1197, opened 2023-11-16
[15.0][IMP] Allow search by task code

iTecan:
This modification allows to search task using the task code that Odoo creates


dmitriypaulov (approved):
LGTM

project_task_code/models/project_task.py
```diff
@@ -42,3 +42,15 @@ def name_get(self):
             name = "[{}] {}".format(rec.code, task[1])
             new_result.append((rec.id, name))
         return new_result
+
+    @api.model
+    def _name_search(
```
yajo, line 47: **issue:** override the public method if possible (and it is).

```suggestion
    def name_search(
```

This is a good practice because the code is more readable and predictable. Public methods are documented. Thus, more developers know about them.

Besides FWIW, the private method was created [to allow overriding the user that issues the search](https://github.com/odoo/odoo/blob/8dddaa61c86d0942fc5a7487707feb6d0da0e1dc/odoo/models.py#L1913-L1914). This is not your case.

project_task_code/models/project_task.py
```diff
@@ -42,3 +42,15 @@ def name_get(self):
...
+        self, name, args=None, operator="ilike", limit=100, name_get_uid=None
+    ):
+        if not args:
+            args = []
+        if name:
+            args += ["|", ["code", "ilike", name]]
```
yajo, line 53: **issue:** only add new domain if operator is ilike.

Otherwise you could be getting to weird domains when call uses a different operator.

yajo (changes requested):
Code review. Some concerns that should be simple to fix. Thanks!

project_task_code/models/project_task.py
ivs-cetmix, line 53: I would suggest using `expression.AND` whenever you need to aggregate two domains. You can  an example [here](https://search.odooism.com/?q=expression.AND&i=nope&files=&excludeFiles=po%24%7Cpot%24%7Cyml%24%7Cyaml%24%7Ccss%24%7C%2Fstatic%2Flib%2F%7C%2Ftests%2F&repos=15.0)

ivs-cetmix (changes requested):

project_task_code/models/project_task.py
dmitriypaulov: @ivs-cetmix did you mean `expression.OR`? It seems like in our case [this one](https://search.odooism.com/?q=expression.OR&i=nope&files=&excludeFiles=po%24%7Cpot%24%7Cyml%24%7Cyaml%24%7Ccss%24%7C%2Fstatic%2Flib%2F%7C%2Ftests%2F&repos=15.0) is more suitable

ivs-cetmix: > @ivs-cetmix did you mean `expression.OR`? It seems like in our case [this one](https://search.odooism.com/?q=expression.OR&i=nope&files=&excludeFiles=po%24%7Cpot%24%7Cyml%24%7Cyaml%24%7Ccss%24%7C%2Fstatic%2Flib%2F%7C%2Ftests%2F&repos=15.0) is more suitable

Yes, you are right. My mistake)

ivs-cetmix:
Hi @iTecan thank you for your contribution! Could you please check the comments and update the code so we could have it merged?

PR #1200, opened 2023-11-22
[16.0][MIG] project_stock_product_set: Migration to 16.0

victoralmau:
Migration to 16.0

@Tecnativa TT45873


project_stock_product_set/models/product_set_line.py
```diff
@@ -8,19 +8,21 @@ class ProductSetLine(models.Model):
...
             "raw_material_task_id": task.id,
-            "picking_type_id": task.picking_type_id.id
-            or task.project_id.picking_type_id.id,
         }
+        # According to default values set from stock_move field in task form
+        stock_move_model = self.env["stock.move"].with_context(
+            default_raw_material_task_id=task.id
+        )
+        values.update(
+            stock_move_model.default_get(
```
pedrobaeza, line 24: This doesn't seem correct, as it's not respecting task definition.

victoralmau: Ok sorry, I think it is necessary to explain how it was in v15 and how it is now in v16.

In v15 there was the `_set_procurement_group_id()` method (https://github.com/OCA/project/blob/15.0/project_stock/models/project_task.py) in `project_stock` that was used in https://github.com/OCA/project/blob/15.0/project_stock/models/stock_move.py#L90 . In `project_stock_product_set` the method call to set the `group_id` https://github.com/OCA/project/blob/15.0/project_stock_product_set/wizard/product_set_add.py#L33 was used in the wizard.

In v16 there is no `_set_procurement_group_id()` method in `project_stock` and this logic has been defined directly in `default_get()` of `stock.move`: https://github.com/OCA/project/blob/16.0/project_stock/models/stock_move.py#L83 For this reason in `project_stock_product_set` `default_get()` of `stock.move` is called with the `default_raw_material_task_id` context of the corresponding task and so the correct values of: `group_id`, `location_id`, `location_dest_id` and `picking_type_id`.

What do you think would be the appropriate change in this case?

pedrobaeza: As said, locations defined in the task or the project are not respected.

victoralmau: According to https://github.com/OCA/project/blob/16.0/project_stock/models/stock_move.py#L89 the definition of the task or the project is respected  for the fields `location_id`, `location_dest_id` and `picking_type_id`, right? What is the reason why this technique does not work in this case?

pedrobaeza: OK, got it. My fault.

pedrobaeza (approved):

chienandalu (commented):
Testing in runbot, when adding a set to the task, the following error happens:

image

victoralmau:
> Testing in runbot, when adding a set to the task, the following error happens:
> 
The error that you indicate is due to the fact that the necessary information is not set (neither in the task or in the project)
task-error
project-error

This is something that could happen also in previous versions, do you think we should take this into account and show a custom error?
The demo project "_Task material_" already has this information set to do some testing without having to configure anything..

chienandalu (approved):
Thanks, I see :smile: 

Yes, it can be a future improvement for `project_stock`. For the moment, let's approve this :+1:

PR #1203, opened 2023-12-11
[FIX] project_duplicate_subtask: access to ir.actions.act_window

mdurepos:
Fixes a bug where regular users (project users and managers) could not duplicate tasks with subtasks unless they had full admin rights. Prior to this fix, they would see the following error:

```
You are not allowed to access 'Action Window' (ir.actions.act_window) records.

This operation is allowed for the following groups:
	- Administration/Settings

Contact your administrator to request access if necessary.
```

### DIFF
--- a/project_duplicate_subtask/models/project_task.py
+++ b/project_duplicate_subtask/models/project_task.py
@@ -1,15 +1,27 @@
 # Copyright (C) 2021 ForgeFlow S.L.
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html)
 
-from odoo import models
+from odoo import api, models
 
 
 class ProjectTask(models.Model):
     _inherit = "project.task"
 
+    @api.model
+    def _extract_action_dict(self, action):
+        action = action.sudo()
+        return {
+            "name": action.name,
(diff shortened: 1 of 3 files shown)


project_duplicate_subtask/models/project_task.py
```diff
@@ -1,15 +1,27 @@
...
+            "context": action.context,
+            "domain": action.domain,
+            "search_view_id": action.search_view_id,
+            "help": action.help,
+        }
+
     def action_duplicate_subtasks(self):
-        action = self.env.ref("project.action_view_task")
-        result = action.read()[0]
+        action = self._extract_action_dict(self.env.ref("project.action_view_task"))
```
AungKoKoLin1997, line 24: You can just make self as sudo() and it will allow to duplicate for all users. I don't think you need to fix in other parts of module.
```suggestion
        self = self.sudo()
        action = self.env.ref("project.action_view_task")
        result = action.read()[0]
```

rousseldenis: @AungKoKoLin1997 Nope. You should use _for_xml_id() function. Search in code to have example.

PR #1209, opened 2023-12-20
[16.0][MIG] project_task_code : alternative without functionality change

robinkeunen:
This PR is an alternative to #1084.

In 15.0 : project_task_code

- adds code field on tasks
- adds sql contraint project_task_unique_code
- a hook initializes code from project.task sequence
- in create, code is assigned from project.task sequence

In his migration review (2023-03-27), @rafaelbn pointed out that the a configuration should be added for the constraint. Otherwise, it breaks the tests from other modules.

I understand from the suggestion that the module should

- add the task code upon installation
- force the unicity constraint on configuration

The fix in #1084 does the opposite, it

- forces the unicity constraint upon installation
- creates the task code on configuration

It has the advantage of making the module multi-company compliant but at the coast of - according to me - changing the functionality of the module.

I leave it to the PSC to choose the most appropriate PR :-)


project_task_code/models/project_task.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    code = fields.Char(
+        string="Task Number",
+        required=True,
+        default="/",
+        readonly=True,
```
ivs-cetmix, line 14: Not for migration: What is the reason behind keeping this field readonly?

ivs-cetmix (approved):
Functional review LGTM. Using this module in prod already)

Shide (approved):
I'm ok with this, but If in the future happens something that breaks runboat, will be good to modify the module.

The runboat break happens to me in this PR: https://github.com/OCA/project/pull/1096#issuecomment-1484977104

Abranes (approved):

project_task_code/models/project_task.py
robinkeunen: Because it is generated through a sequence :-)

robinkeunen:
I'm sorry this is going back and forth that way, not trying to be a troll ... :/

I actually was convinced by @huguesdk and @legalsylvain 's comment on https://github.com/OCA/project/pull/1084 . I see no reason why the sql constraint would conflict w/ other module since the field code is added by this module.

I wonder if the error you had in v15 was due to the constraint checking unicity on the code but not on the company ? We'll see and fix if it rises again.

robinkeunen:
Last 3 commits should be smashed.

robinkeunen:
Hi @dreispt , can you take a look at this PR ? It raised many concerns in https://github.com/OCA/project/pull/1084 and https://github.com/OCA/project/pull/1013 . It would be great to finally settle it :-)

huguesdk (approved):
@robinkeunen thanks for this. would you please squash the last 3 commits?

Shide (approved):

rafaelbn:
@legalsylvain , are you PSC here to merge?

legalsylvain:
Nop ! 
Here is the list : https://github.com/OCA/repo-maintainer-conf/blob/master/conf/psc/project.yml

rafaelbn (approved):
/ocabot merge patch

Merging as long discussed in other PR, but if any issue I will ping you 😄

robinkeunen:
🙏 that was a long one, thanks @Shide and @Abranes

PR #1213, opened 2024-01-05
[15.0][FIX] project_wbs: fix code, indent consistency and list view

AaronHForgeFlow:
2 fixes:

- the view:

image

- wbs complete code and indent consistency:

make computed fields in analytic account not editable from the project That can cause data corruption between the project and the analytic account, in case a change is needed in the project the master data in the analytic account should change instead, that is, the code and the parent_id fields. For example, I have a project P1 and a child 01. The complete wbs code of the child is P1 / 01. If I change the complete wbs code in the child to be XXX the complete wbs code of the parent is still P1 but the complete wbs code of the child would be XXX and the code in the child analytic account will still be 01. That is not consistent.

@ForgeFlow

### DIFF
--- a/project_wbs/models/project_project.py
+++ b/project_wbs/models/project_project.py
@@ -191,13 +191,11 @@ class Project(models.Model):
     )
     wbs_indent = fields.Char(
         related="analytic_account_id.wbs_indent",
-        readonly=False,
     )
     complete_wbs_code = fields.Char(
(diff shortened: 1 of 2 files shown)


GuillemCForgeFlow (approved):
Code and functional review 👍🏽

LoisRForgeFlow (approved):
LGTM

PR #1210, opened 2023-12-22
[16.0][MIG] project_key

vincent-hatakeyama:
Supersedes #1169 with multi create.

PR #1221, opened 2024-01-23
[16.0][IMP] project_key: improve display

vincent-hatakeyama:
Goes after #1210


hspl-vegra (approved):

dreispt:
This should be rebased, so that it shows only the commit with the improvements.

dreispt (approved):

PR #1222, opened 2024-01-24
[16.0][MIG] project_administrator_restricted_visibility: Migration to 16.0

edlopen:
Standard migration of the project_administrator_restricted_visibility.

MT-4844

@moduon


project_administrator_restricted_visibility/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "Project Administrator Restricted Visibility",
+    "version": "16.0.1.0.0",
+    "summary": "Adds a 'Project Administrator' access group "
+    "with restricted visibility to 'Projects'",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/project",
+    "license": "AGPL-3",
+    "category": "Project",
```
yajo, line 9: Please add yourself and @rafaelbn as [maintainers](https://doc.moduon.team/doc/estandares-de-contribucion-gxxTTBnNnb#h-mantenedores).

yajo (changes requested):
Code looks good!

One minor guideline missing to follow.

Also, rebase fixing your committer email as we discussed privately.

To let the bot auto-add this as the migration PR for the module, we issue this command:

/ocabot migration project_administrator_restricted_visibility

EmilioPascual (approved):
LGTM

project_administrator_restricted_visibility/__manifest__.py
edlopen: Done!

NachoAlesLopez:
Code looks good and the module works perfectly fine. I did find a small issue. When trying to uninstall the module, it throws this error:

image

It seems to be related to the uninstall hook. In older versions of Odoo (13 and lower), the field `groups` of `ir.rule` didn't have the `ondelete` constraint (https://github.com/odoo/odoo/blob/13.0/odoo/addons/base/models/ir_rule.py#L22) but newer version have added it (https://github.com/odoo/odoo/blob/16.0/odoo/addons/base/models/ir_rule.py#L23). Can you check this error out?

edlopen:
Thanks @NachoAlesLopez . It seems that this bug has existed since version 14. I just fixed the uninstall hook so that it first removes the group from the rule and then deletes it.

NachoAlesLopez (approved):
Thanks and great job, @edlopen! Just tested it and the module uninstalled successfully.

project_administrator_restricted_visibility/tests/test_uninstall_hook.py
```diff
@@ -0,0 +1,32 @@
...
+from odoo.tests.common import TransactionCase
+
+from ..hooks import uninstall_hook
+
+
+class TestUninstallHook(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
```
yajo: **issue:** this is unnecessary

```suggestion
```

yajo (changes requested):
Great job with that test! 🎉 

One minor fix to be ready

project_administrator_restricted_visibility/tests/test_uninstall_hook.py
edlopen: Done!

yajo (approved):
/ocabot merge patch

PR #1223, opened 2024-01-24
[15.0][FIX] project_wbs: calculation of root analytic account

AaronHForgeFlow:
In case there is no parent project with account_class project we should not take the parent, but use current project. Steps to reproduce:

1. Create a master project account_class = null
2. Create a child project account_class = 'project'
3. The root analytic account is missing, it should be the child project as it is the root project (only project with account class 'project' are considered a project)

cc @ForgeFlow


JordiMForgeFlow (approved):
LGTM

MiquelRForgeFlow (approved):

JordiBForgeFlow (approved):

PR #1229, opened 2024-01-30
[16.0] [ADD] Project Portal Custom Field Access

ivantodorovich:
Odoo has a feature called "Project Sharing" that allows portal users to view a project's kanban, which has functionalities similar to the backend system.

There's a preset list of fields that portal users can access, which is harcoded in the module's source code.

That's nice but inconvinient if a user introduces new fields to the form and wants to share these with the portal users. They can't currently do this without using a Python override.

To resolve this, the module lets users whitelist these custom fields directly through the ``ir.model.field`` model.

----

Re-created an old closed PR:
- https://github.com/OCA/project/pull/1119

ping @AungKoKoLin1997


project_portal_custom_field_access/models/project_task.py
```diff
@@ -0,0 +1,30 @@
...
+                    ("project_portal_access", "=", True),
+                ],
+                ["name"],
+            )
+        )
+        return {r["name"] for r in res}
+
+    @property
+    def SELF_WRITABLE_FIELDS(self):
+        return super().SELF_WRITABLE_FIELDS | self._get_portal_custom_field_access()
```
AungKoKoLin1997, line 30: How do you think if we add an option to choose `READABLE` or `WRITEABLE`? It may be useful in case of user wants to show only information for the collaborators like `Milestone` field.

ivantodorovich: The `milestone_id` is a core field, and it's already exposed: https://github.com/odoo/odoo/blob/492169bf562768f661727d6387c5e47bee700d9e/addons/project/models/project.py#L49

The purpose of this module is to be able to expose custom fields (those created through the UI, those that start with `x_`). All other code-defined fields can be exposed through custom modules :) this module is not useful there

I do take your point to be able to choose between read-only and read-and-write. It would be a nice improvement, though it's not required by our project at the moment

AungKoKoLin1997: I think you misunderstood me about `milestone_id`. I wanted to mention was in case of  user wants to create custom fields through UI and he wants to make only readable for collaborators(I mentioned `milestone_id` as an example of readonly for collaborator). Whatever you got my point.

PR #1231, opened 2024-01-31
[ADD] project_internal_collaborator: add module

dessanhemrayev:
This module implements "Project Collaborators". These users have readonly access to projects where they are added as collaborators and tasks they are following in such projects.
Only users who have "Project/User" group can be added as collaborators.
Collaborators can be added to and removed from projects both manually and automatically.


project_collaborator/models/project_project.py
```diff
@@ -0,0 +1,88 @@
+# Copyright (C) 2024 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, fields, models
+
+
+class Project(models.Model):
+    _inherit = "project.project"
+
+    collaborator_ids = fields.Many2many(
```
AungKoKoLin1997: Odoo standard have this field name, isn't? Does this module override the standard project collaborator features?
https://github.com/odoo/odoo/blob/059b19c59a9222b676be5246eb94d680cf6bf27b/addons/project/models/project.py#L373

project_collaborator/security/rules.xml
```diff
@@ -0,0 +1,13 @@
+<odoo noupdate="1">
+    <!-- Project: Collaborators can see assigned projects -->
```
AungKoKoLin1997: Is that enough to see projects for the collaborators with `base.group_user`? Project app is only visible for project users. Or am I miss something?
https://github.com/odoo/odoo/blob/059b19c59a9222b676be5246eb94d680cf6bf27b/addons/project/views/project_views.xml#L5-L9

project_collaborator/models/project_project.py
dessanhemrayev: Thanks, for review.
I'll fix it

PR #1237, opened 2024-02-14
[15.0][ADD] project_task_note: Add notes in project task

carolinafernandez-tecnativa:
TT47764

@Tecnativa

@pedrobaeza @pilarvargas-tecnativa


project_task_note/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2024 Tecnativa Carolina Fernandez
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
+    "name": "Add notes in Project Task",
```
pedrobaeza: This text is shorter from the summary (or just put the name, not both).

```suggestion
    "name": "Project task notes",
```

project_task_note/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "version": "15.0.1.0.0",
+    "category": "Project Management",
+    "summary": "Add notes in Project Task",
```
pedrobaeza:
```suggestion
    "summary": "Add notes in project tasks",
```

project_task_note/models/project_task.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Tecnativa Carolina Fernandez
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import fields, models
+
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    notes = fields.Html()
```
pedrobaeza:
```suggestion
    notes = fields.Html(string="Internal notes")
```

project_task_note/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module add notes in project task only visible for internal users
```
pedrobaeza:
```suggestion
This module add notes in project tasks, only visible for internal users.
```

Add also a USAGE section for indicating where to see that notes, and maybe a CONTEXT file for indicating why we are creating this (because there's no field where storing general comments and notes not visible to the customers in the portal).

project_task_note/views/project_task_views.xml
```diff
@@ -0,0 +1,20 @@
...
+        Copyright 2024 Tecnativa Carolina Fernandez
+        License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+    -->
+    <!-- Project Task form -->
+    <record id="view_task_form2" model="ir.ui.view">
+        <field name="name">project_task.form.note</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.view_task_form2" />
+        <field name="arch" type="xml">
+            <page name="description_page" position="after">
```
pedrobaeza: I'm thinking that is better to put the notes in the same tab of the description, as being in read mode, you don't have to switch between pages to see both texts, and they are shortened to the minimum occupied height. Anyway, check how this will see in v16 (always in edit mode).

pedrobaeza (changes requested):

pilarvargas-tecnativa (approved):
LGTM :)

pedrobaeza:
The way the field widget is hooked seems to be different:

image


Please use the same as the other field. Have you tested how it looks on v16?

carolinafernandez-tecnativa:
> The way the field widget is hooked seems to be different:
> 
> image
> 
> Please use the same as the other field. Have you tested how it looks on v16?

I added the same as the other field

https://github.com/odoo/odoo/blob/15.0/addons/project/views/project_views.xml#L898

but stills appear as in your picture, also without group.

On v16 works fine, instead of using group, I added a label because otherwise the field appears as readonly I do not know why..

image

project_task_note/views/project_task_views.xml
```diff
@@ -0,0 +1,20 @@
...
+    -->
+    <!-- Project Task form -->
+    <record id="view_task_form2" model="ir.ui.view">
+        <field name="name">project_task.form.note</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.view_task_form2" />
+        <field name="arch" type="xml">
+            <field name="description" position="after">
+                <group name="internal_notes_group" string="Internal notes">
+                    <field name="notes" type="html" nolabel="1" />
```
pedrobaeza: You can put the collaborative flag also here.

pedrobaeza (approved):
/ocabot merge nobump

Migrate it to v16, please

PR #1243, opened 2024-03-05
[17.0][MIG] project_list: Migration to 17.0

MohamedOsman7:


MohamedOsman7:
Not needed, now part of the modul: project

PR #1246, opened 2024-03-06
[IMP] project_task_code: enable name search by Task Number

CLaurelB:
The search functionality has been improved to include the Task Number.
This enhancement is achieved by inheriting the _name_search method to
include the Task Number in the search domain.

The parent method adds the Name to the search domain (1). As a
consequence of this action, when the args parameter is empty in the
inherited method, an incomplete OR term must be added. Without this
addition, the search domain would default to using an AND operator.

Additionally, tests have been added for this new inheritance.

(1) https://github.com/odoo/odoo/blob/4625cb2f/odoo/models.py#L1928

### DIFF
--- a/project_task_code/models/project_task.py
+++ b/project_task_code/models/project_task.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
+from odoo.osv import expression
 
 
 class ProjectTask(models.Model):
@@ -42,3 +43,20 @@ class ProjectTask(models.Model):
             name = "[{}] {}".format(rec.code, task[1])
             new_result.append((rec.id, name))
         return new_result
+
+    @api.model
+    def _name_search(self, name="", args=None, operator="ilike", limit=100):
+        """Allow searching by code by default."""
+        if name and operator in ["=", "ilike", "=ilike", "like", "=like"]:
+            args = (
+                expression.OR([[("code", operator, name)], args])
+                if args
+                # The parent method adds the Name to the search domain. As a
+                # consequence of this action, when the args parameter is empty in the
+                # inherited method, an incomplete OR term must be added. Without this
+                # addition, the search domain would default to using an AND operator.
+                else ["|", ("code", operator, name)]
+            )
+        return super()._name_search(
+            name=name, args=args, operator=operator, limit=limit
+        )
--- a/project_task_code/tests/test_project_task_code.py
+++ b/project_task_code/tests/test_project_task_code.py
@@ -38,3 +38,25 @@ class TestProjectTaskCode(common.TransactionCase):
         )
         result = project_task.name_get()
         self.assertEqual(result[0][1], "[%s] Task Testing Get Name" % code)
+
+    def test_name_search(self):
+        number_next = self.task_sequence.number_next_actual
+        code = self.task_sequence.get_next_char(number_next)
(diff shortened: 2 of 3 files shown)


CLaurelB:
@luisg123v could you review, please?

desdelinux (approved):

project_task_code/models/project_task.py
```diff
@@ -42,3 +43,14 @@ def name_get(self):
             name = "[{}] {}".format(rec.code, task[1])
             new_result.append((rec.id, name))
         return new_result
+
+    @api.model
+    def name_search(self, name="", args=None, operator="ilike", limit=100):
+        """Allow searching by code by default."""
+        if name:
+            args = (
+                expression.OR([[("code", operator, name)], args])
```
luisg123v, line 52: It depends on whether the operator is positive or negative.

project_task_code/models/project_task.py
```diff
@@ -42,3 +43,14 @@ def name_get(self):
...
+                if args
+                else ["|", ("code", operator, name)]
```
luisg123v, line 58: This is an incomplete term because it's a binary operator followed by only one operand.

luisg123v (changes requested):
Could you add myself as maintainer, please?

luisg123v (changes requested):
Commit description

First mention the improvement, then how it's achieved.

project_task_code/tests/test_project_task_code.py
```diff
@@ -38,3 +38,14 @@ def test_name_get(self):
...
+    def test_name_search(self):
+        number_next = self.task_sequence.number_next_actual
+        code = self.task_sequence.get_next_char(number_next)
+        project_task = self.project_task_model.create(
+            {
+                "name": "Task Testing Name Search",
+            }
+        )
+        result = project_task.name_search(name=code)
+        self.assertEqual(result[0][1], project_task.display_name)
```
luisg123v: I'd check 1st position (ID), because we're not validating name in this test but the gotten result matches our task.

luisg123v (approved):
LGTM :+1: 

@moylop260 could you review and merge, please?

Regards,

desdelinux (approved):

project_task_code/models/project_task.py
moylop260: Could you add a unit testing to cover this line of code, please?

Spoiler Alert: My local testing is raising the following error:
- `ValueError: Domain ['|', ('default_code', 'ilike', '0')] is syntactically not correct.`

(Maybe I'm doing something wrong)

CLaurelB:
Hello @pedrobaeza @yajo 

Could you review, please?

Regards.
cc. @moylop260 @luisg123v

project_task_code/models/project_task.py
yajo: Probably happening to you because you're doing a search without a term (a.k.a. "give me all results"). Then you hit https://github.com/odoo/odoo/blob/8f115ccc20f93f52a93dd6836b021463e7227c91/odoo/models.py#L1927 which probably triggers that bug.

yajo (changes requested):

project_task_code/models/project_task.py
moylop260: Yeah, using the unittesting is working well and I have tested from odoo-bin shell

thank you!

So, approve?

yajo (approved):
/ocabot merge patch

PR #1252, opened 2024-03-12
[16.0][IMP] project_stock_product_set: Adapt wizard to changes in product_set

pilarvargas-tecnativa:
Pending from:

- [x] https://github.com/OCA/product-attribute/pull/1548

Moved some of the logic shared by the wizard defined in the module sale_product_set to the module product_set so the dependency of project_stock_product_set should now be on product_set.

cc @Tecnativa

@pedrobaeza @victoralmau please review


rousseldenis (approved):

pedrobaeza (commented):
In `add_set`, call always super, and don't add the `self.check_partner()` call.

pilarvargas-tecnativa:
> In `add_set`, call always super, and don't add the `self.check_partner()` call.

:+1:  thanks

victoralmau (changes requested):
Remember to only take into account product type lines (avoid sections for example).

pilarvargas-tecnativa:
> Remember to only take into account product type lines (avoid sections for example).

True, this was not taken into account before, it is already included in this module as well, thanks for reminding me.

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge major

PR #1253, opened 2024-03-18
[16.0][FIX] project_timeline: Fix unassigned task color

houzefa-abba:
Unassigned tasks are supposed to show up in white instead of default color, but the previous `== []` comparison is no longer working in Odoo 16 / with web_timeline 16.0.1.0.3.

This is because web_timeline [forces string conversion](https://github.com/OCA/web/blob/16.0/web_timeline/static/src/js/timeline_renderer.js#L515) during evaluation of these color instructions, and an empty `user_ids` is an empty string in this case.

This does look somewhat odd but fixes the issue short of rewriting py-eval instructions in web_timeline, which would be a much larger task.

# Before

Screenshot at 2024-03-18 18-26-20

# After

Screenshot at 2024-03-18 18-26-47

I know the "Unassigned" title on the left seems wrong here (as by default we group per project), but that's another issue.

### DIFF
--- a/project_timeline/views/project_task_view.xml
+++ b/project_timeline/views/project_task_view.xml
@@ -15,7 +15,7 @@
                 date_stop="date_end"
                 default_group_by="project_id"
                 event_open_popup="true"
-                colors="white: user_ids == []; #2ecb71: kanban_state == 'done'; #ec7063: kanban_state == 'blocked'"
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
/ocabot merge minor

PR #1255, opened 2024-03-27
[15.0][FIX] project_wbs: respect the code that user manually puts in the project

AaronHForgeFlow:
If the user selects a custom code in the project the analytic account should use the same code.

In order to reproduce just fill the code field in the form view. Without this change, the original code generated with the analytic sequence (ir.sequence) cannot be changed.

cc @ForgeFlow

Lets put an example: 

Create a project, the code is set:

Then I go to the form view and change the code here:

And I expect the code to be set in the project and in the analytic account:

But without this change the code is not set:

I checked if the code is not null because there is a constraint that prevents duplicated codes and CI fill fail due to duplicated null codes.

### DIFF
--- a/project_wbs/models/project_project.py
+++ b/project_wbs/models/project_project.py
@@ -145,12 +145,16 @@ class Project(models.Model):
         return None
 
     def prepare_analytics_vals(self, vals):
-        return {
+        analytic_vals = {
             "name": vals.get("name", _("Unknown Analytic Account")),
-            "company_id": vals.get("company_id", self.env.user.company_id.id),
+            "company_id": vals.get("company_id", self.env.company.id),
             "partner_id": vals.get("partner_id"),
             "active": True,
         }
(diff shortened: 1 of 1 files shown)


GuillemCForgeFlow (approved):
LGTM! 👍🏿

project_wbs/models/project_project.py
```diff
@@ -145,12 +145,16 @@ def _resolve_analytic_account_id_from_context(self):
         return None
 
     def prepare_analytics_vals(self, vals):
-        return {
+        analytic_vals = {
             "name": vals.get("name", _("Unknown Analytic Account")),
             "company_id": vals.get("company_id", self.env.user.company_id.id),
```
MiquelRForgeFlow:
```suggestion
            "company_id": vals.get("company_id", self.env.company.id),
```

JordiBForgeFlow (approved):

Issue #1267, opened 2024-04-17
Issue migrating project_milestone from 14.0 to 15.0


remytms:
I'm migrating a database from 14.0 to 16.0, and I face issue with migration of project_milestone between 14.0 and 15.0.

## Module
project_milestone

## Describe the bug
The migration script fails to run when performing an update with OpenUpgrade in 15.0 on a database with project milestone installed.

## To Reproduce
**Affected versions**: 15.0

1. create a database with project_milestone installed in 14.0
2. run OpenUpgrade in 15.0 on this database
3. the upgrade fails.

**Expected behavior**
Migration script of `project_milestone` in 15.0 should not fail and block migration.

**Additional context**
The errors comes from the fact that :
- `project_task_milestone` is already existing in the database event when not installed, so renaming the module fails.
- the field `deadline` is already existing because module `project` has already been updated before `project_milestone`, so renaming fails.

I can not figure out what is the case where this migration script will work.

I suggest the following solution:

In OpenUpgrade 15.0, use `apriori.py` file to rename module `project_milestone` into `project_task_milestone` and add the column rename in the migration script of the `project` module so that the rename happen before the `deadline` columns creation by the update of the `project` module.

Remove the empty `project_milestone` module from 15.0 branch.

What do you think ? I will try such a solution, and come back with feedback.

remytms:
I've done some tests.

I added [this pr](https://github.com/OCA/OpenUpgrade/pull/4397) to OpenUpgrade to cover the change between 14.0 to 15.0. And I added [this pr](https://github.com/OCA/OpenUpgrade/pull/4398) to cover change between 15.0 to 16.0.

It works well except that for the time being, stages need to be activated manually in project configuration.

remytms:
@patrickrwilson You may be interested in this issue as maintainer of `project_milestone`.

PR #1268, opened 2024-04-18
[15.0][REM] project_milestone

remytms:
Milestones are present in module `project` starting from 15.0 and scripts that exists in this empty module does not work.

See #1267 

See also https://github.com/OCA/OpenUpgrade/pull/4397


rvalyi (approved):
Migration is covered in OpenUpgrade

pedrobaeza:
Isn't this only a placeholder for the migration scripts?

rvalyi:
> Isn't this only a placeholder for the migration scripts?

it doesn't seem so as per the comments in #1267

so removing the module seems legit.

pedrobaeza (approved):
OK

/ocabot merge nobump

sebalix:
[`project_template_milestone`](https://github.com/OCA/project/blob/15.0/project_template_milestone/__manifest__.py) in this repository is still referencing this removed module, but the CI is happy with that, weird :thinking:

PR #1270, opened 2024-04-25
[16.0][FIX] project_role: pass company from env

ByteMeAsap:
Pass company from env rather than users default company to prevent multicompany issues


alexey-pelykh (approved):
LGTM

PR #1273, opened 2024-04-30
[MIG] project_category

petrus-v:


mymage:
Hi @petrus-v, I'm looking for this module as I said here https://github.com/OCA/project/issues/859#issuecomment-2105951334.
In the meantime I did my migration PR #1278 just to check compatibility.
I'll leave it on draft: let me know if you will go further with yours, I think you need only to re-push.
... and I just found that it is only a change of the module to project_type so we can close our PR

petrus-v:
Thanks @mymage for your comments.

This PR as been open in quiet dirty mode.

I've move this PR draft, @Rom10811 here could you figure out if this module is really needs (as pointed out by @mymage) and follow the OCA's process to keep commits if needs (cf https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0).

Regards,

rvalyi:
this doesn't respect the OCA migration process and also project_category is now replaced by project_type in v16 https://github.com/OCA/project/tree/16.0/project_type

PR #1278, opened 2024-05-12
[16.0][MIG] project_category

mymage:
Migration test


mymage:
I just found that it is only a change of the module to project_type

mymage:
I just found that it is only a change of the module to project_type

PR #1279, opened 2024-05-13
[14.0][FIX] project_wbs: Don't change project.project description

pedrobaeza:
The change of the project.project model description causes that all the modules in the same repository adopt the new description and export it as the value to translate, making than any instance using one of the modules of OCA/project potentially has this translation, which is catastrophic.

@Tecnativa TT49118


victoralmau (approved):

rvalyi (approved):

carolinafernandez-tecnativa (approved):

AaronHForgeFlow:
Thank you!

PR #1113, opened 2023-05-12
[16.0][ADD] project_stage_extra_info

dsolanki-initos:
After the discussion with @pedrobaeza on [ project_status PR](https://github.com/OCA/project/pull/1109) I've created this PR
This module adds a Description and Is closed stage field on Project stages.

PR #1282, opened 2024-05-15
[16.0][MIG] project_status: Migration to 16.0

atularvind:


FrancoMaxime (approved):
LGTM: code review

thomaspaulb:
The previous MR migrating this was [closed](https://github.com/OCA/project/pull/1109) because the functionality is now in [project_stage_extra_info](https://github.com/OCA/project/pull/1113).

Just the migration script is missing, it needs something like [this](https://github.com/OCA/project/pull/892) but then for `project_status` - just an empty module that carries only the pre-16.0.1.0.0 migration script and that carries over the fields.

rousseldenis:
@atularvind Could you take @thomaspaulb comment into account ?

coleste (approved):
LGTM

dreispt (approved):

PR #1290, opened 2024-05-24
[17.0][ADD] `project_timesheet_holidays_contract`

yankinmax:
Module allows to take into account employee contract along with provided end date when timesheet on time off.


project_timesheet_holidays_contract/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2023 Camptocamp SA
```
vvrossem:
```suggestion
# Copyright 2024 Camptocamp SA
```

project_timesheet_holidays_contract/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Project Timesheet Holidays Contract",
+    "summary": "Schedule when on time Off with employee contracts taken into account",
```
vvrossem: `summary` is a bit unclear, IMHO (but so is [project_timesheet_holidays](https://github.com/odoo/odoo/blob/17.0/addons/project_timesheet_holidays/__manifest__.py#L8)'s summary).
What do you think of something of the like of 
```
"summary": "Automatically log employees' leave in timesheets according to their contracts.",
```

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,133 @@
...
+    def _get_work_days_data_batch(
+        self,
+        from_datetime,
+        to_datetime,
+        compute_leaves=True,
+        calendar=None,
+        domain=None,
+    ):
+        """Override calendar in favor of the one taken from employee's contract
+        Only if one has not been already given in arguments"""
```
vvrossem: Use `#` comment line to avoid overriding the docstring (if any)

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,133 @@
...
+                    domain=domain,
+                )
+            )
+        return result
+
+    def _get_leave_days_data_batch(
+        self, from_datetime, to_datetime, calendar=None, domain=None
+    ):
+        """Override calendar in favor of the one taken from employee's contract
+        Only if one has not been already given in arguments"""
```
vvrossem: Same: use `#` comment line to avoid overriding the docstring (if any)

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,133 @@
...
+                    domain=domain,
+                )
+            )
+        return result
+
+    def list_work_time_per_day(
+        self, from_datetime, to_datetime, calendar=None, domain=None
+    ):
+        """Override calendar in favor of the one taken from employee's contract
+        Only if one has not been already given in arguments"""
```
vvrossem: Same: use `#` comment line to avoid overriding the docstring (if any)

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,133 @@
...
+                    to_datetime=to_datetime,
+                    calendar=calendar,
+                    domain=domain,
+                )
+            )
+        return result
+
+    def list_leaves(self, from_datetime, to_datetime, calendar=None, domain=None):
+        """Override calendar in favor of the one taken from employee's contract
+        Only if one has not been already given in arguments"""
```
vvrossem: same: use `#` comment line to avoid overriding the docstring (if any)

project_timesheet_holidays_contract/tests/common.py
```diff
@@ -0,0 +1,118 @@
+# Copyright 2023 Camptocamp SA
```
vvrossem:
```suggestion
# Copyright 2024 Camptocamp SA
```

project_timesheet_holidays_contract/tests/test_hr_employee_calendar.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2023 Camptocamp SA
```
vvrossem:
```suggestion
# Copyright 2024 Camptocamp SA
```

project_timesheet_holidays_contract/tests/test_hr_leave_days_timesheet.py
```diff
@@ -0,0 +1,101 @@
+# Copyright 2023 Camptocamp SA
```
vvrossem:
```suggestion
# Copyright 2024 Camptocamp SA
```

vvrossem (approved):
pre-approving, some minor suggestions

imlopes (approved):

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,133 @@
...
+            .with_context(active_test=True)
+            .env["hr.contract"]
+            .search([("employee_id", "=", self.id)])
+        )
+
+        if date_from:
+            date_from = (
+                date_from if not isinstance(date_from, datetime) else date_from.date()
+            )
+            contracts = contracts.filtered(lambda c: c.date_start <= date_from)
```
SilvioC2C: `if date_from` at line 24 will always be `True`; if not, you would've exited the function at line 15.
So you don't need this check, and you can use `date_from` directly in the `search()` domain at line 21, instead of filtering the results after.

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,133 @@
...
+
+        date_to = self.env.context.get("date_to")
+        if date_to:
+            date_to = date_to if not isinstance(date_to, datetime) else date_to.date()
+            contracts = contracts.filtered(
+                lambda c: not c.date_end or c.date_end >= date_to
+            )
```
SilvioC2C: Filtering after searching is a bad practice perf-wise, and should be avoided unless necessary (for example, filtering the recordset according to a non-searchable field is ok, since you cannot do it otherwise).
Please handle both `date_from` and `date_to` variables before the `search()` and implement its domain according to those parameters, instead of searching and filtering.

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,133 @@
...
+                return contracts[0].resource_calendar_id
+            # But if not, do not take cancelled ones and see what's left
+            contracts = contracts.filtered(lambda c: c.state != "cancel")
+            if contracts:
+                if len(contracts) == 1:
+                    return contracts[0].resource_calendar_id
+            # There are multiple contracts for this date period
+            raise ValidationError(_("This period overlaps multiple contracts!"))
+
+        return res
```
SilvioC2C: If any contract is found, either a contract's calendar is returned, or an error is raised.
If no contract is found, the result of the `super()` call is returned.
But the `super()` call is executed before knowing if any contract is found, which makes that call redundant if you find any contract.
Please refactor the code so that you call `super()._get_calendar(date_from)` only if necessary, to improve performances and avoid executing code unnecessarily.

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,133 @@
...
+        to_datetime,
+        compute_leaves=True,
+        calendar=None,
+        domain=None,
+    ):
+        # OVERRIDE calendar in favor of the one taken from employee's contract
+        # Only if one has not been already given in arguments
+        result = {}
+        for employee in self:
+            calendar = calendar or employee.with_context(
```
SilvioC2C: This code contains a bug:
- let `self` be a non-empty recordset, not a single record, for example `hr.employee(1, 2, 3, 4, 5)`
- let `calendar` be `None` when the method is called
- if `employee.with_context(date_to=to_datetime)._get_calendar(from_datetime)` returns a calendar for a given employee (for example, `hr.employee(3,)`, you will override the `calendar` variable for every following employee
- this means that every other following record (ie: `hr.employee(4, 5)`) will use the calendar you retrieved earlier (for `hr.employee(3)`)

Please fix this.
Besides, since you're overriding the method only to pass a custom calendar if the original variable is not set, you can exit the override and avoid the `for` loop if the variable is set. This way, you'll call `super()` once on the whole recordset, instead of calling it on each record one-by-one.

PS: since you're updating each record's context with the same values, you should update the context on the whole recordset _before_ cycling it in the `for` loop.

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,133 @@
...
+        return result
+
+    def _get_leave_days_data_batch(
+        self, from_datetime, to_datetime, calendar=None, domain=None
+    ):
+        # OVERRIDE calendar in favor of the one taken from employee's contract
+        # Only if one has not been already given in arguments
+        result = {}
+        for employee in self:
+            calendar = calendar or employee.with_context(
```
SilvioC2C: Same remarks as above

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,133 @@
...
+        return result
+
+    def list_work_time_per_day(
+        self, from_datetime, to_datetime, calendar=None, domain=None
+    ):
+        # OVERRIDE calendar in favor of the one taken from employee's contract
+        # Only if one has not been already given in arguments
+        result = []
+        for employee in self:
+            calendar = calendar or employee.with_context(
```
SilvioC2C: Same remarks as above

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,133 @@
...
+                )
+            )
+        return result
+
+    def list_leaves(self, from_datetime, to_datetime, calendar=None, domain=None):
+        # OVERRIDE calendar in favor of the one taken from employee's contract
+        # Only if one has not been already given in arguments
+        result = []
+        for employee in self:
+            calendar = calendar or employee.with_context(
```
SilvioC2C: Same remarks as above

SilvioC2C (changes requested):

project_timesheet_holidays_contract/models/hr_employee.py
SilvioC2C: PS: check at line 43 is useless, you can remove it: first you check if the recordset contains at least 1 record, then you check if it contains exactly 1 record => you should keep only the second check.

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,129 @@
...
+
+from odoo import _, fields, models
+from odoo.exceptions import ValidationError
+
+
+class HrEmployee(models.Model):
+    _inherit = "hr.employee"
+
+    def _get_datetime_date(self, date):
+        return date if not isinstance(date, datetime) else date.date()
```
SilvioC2C: I think it's unnecessary to have a new method just to convert `datetime` objects to `date`.

But if we really want to extract it, then I'd rather go with one of the following options:
1. use `@staticmethod` decorator
2. move this method outside of the class
3. move this method inside of `_get_calendar()` method

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def _get_calendar(self, date_from=None):
+        res = super()._get_calendar(date_from)
+        if not date_from:
+            return res
+        date_from = self._get_datetime_date(date_from)
+        date_to = self.env.context.get("date_to", fields.Date.today())
+        date_to = self._get_datetime_date(date_to)
```
SilvioC2C: What if you get `"date_to": False` in the context?

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,129 @@
...
+            )
+        )
+        if contracts:
+            # Filter out cancelled contracts
+            contracts = contracts.filtered(lambda c: c.state != "cancel")
+            if len(contracts) == 1:
+                return contracts[0].resource_calendar_id
+            # Multiple non-cancelled contracts for this date period
+            if contracts:
+                raise ValidationError(_("This period overlaps multiple contracts!"))
```
SilvioC2C: In the previous version of the code:
- if only 1 contract was retrieved by `search()`, it was returned regardless of its state;
- if there were more than 1, then the contracts would've been filtered by `state != "cancel"`, and if there were only 1 non-cancelled contracts, it would've been returned, else an error would be raised

Now:
- contracts are filtered by state beforehand, and if there's only 1 non-cancelled contract, it is returned
- if there are multiple non-cancelled contracts, an error is raised
- if all retrieved contracts are cancelled, nothing happens and `super()` is called

Is this change in the code behavior expected?

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,129 @@
...
+class HrEmployee(models.Model):
+    _inherit = "hr.employee"
+
+    def _get_datetime_date(self, date):
+        return date if not isinstance(date, datetime) else date.date()
+
+    def _get_calendar(self, date_from=None):
+        res = super()._get_calendar(date_from)
+        if not date_from:
+            return res
```
SilvioC2C: You only need to call `super()` if:
- `date_from` is not defined
- no calendar is found according to the logic below

Therefore, calling `super()` at the beginning of the method, _before_ you check the conditions mentioned above, just doesn't make any sense. If you call `super()` now:
- if you don't find any calendar later on, you'll end up calling `super()` again at line 45), so Odoo will perform twice the same method
- if you do find a calendar, the result of the `super()` call will be ignored, making the execution of the `super()` itself useless
- the only case in which you actually use the `super()`'s result is when `date_from` is not set

Instead, what you should have is:
```
        if not date_from:  # no `date_from` => call `super()`
            return super()._get_calendar(date_from)
        ... # try to retrieve calendars
        if calendar:
            return calendar  # calendar found => no need to call `super()`
        return super()._get_calendar(date_from)  # no calendar found => call `super()`
```

SilvioC2C (changes requested):

yankinmax:
@SilvioC2C PR is updated, thanks for the perfect suggestions.

project_timesheet_holidays_contract/models/hr_employee.py
yankinmax: @SilvioC2C I don't have the requirements nor the exact behavior. The original request was to deliver this module as an OCA one instead of a custom.

SilvioC2C (approved):
Could use some more tests, but overall LGTM

Thanks for your work!

project_timesheet_holidays_contract/README.rst
```diff
@@ -0,0 +1,78 @@
...
+    :target: https://translation.odoo-community.org/projects/project-17-0/project-17-0-project_timesheet_holidays_contract
+    :alt: Translate me on Weblate
+.. |badge5| image:: https://img.shields.io/badge/runboat-Try%20me-875A7B.png
+    :target: https://runboat.odoo-community.org/builds?repo=OCA/project&target_branch=17.0
+    :alt: Try me on Runboat
+
+|badge1| |badge2| |badge3| |badge4| |badge5|
+
+Module allows to take into account employee contract along with provided
+end date when timesheet on time off.
```
grindtildeath, line 32: IMO you could develop a bit more how this works, because after reading it twice, I'm still not sure what the module does.

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,184 @@
...
+                ]
+            )
+        )
+
+        # If only one contract, return it regardless of its state
+        if len(contracts) == 1:
+            return contracts[0].resource_calendar_id
+
+        # Filter out cancelled contracts if there are multiple contracts
+        valid_contracts = contracts.filtered(lambda c: c.state != "cancel")
```
grindtildeath, line 44: So in case there's only one contract that is canceled, it still must be used? :raised_eyebrow:

project_timesheet_holidays_contract/models/hr_employee.py
```diff
@@ -0,0 +1,184 @@
...
+                compute_leaves=compute_leaves,
+                calendar=calendar,
+                domain=domain,
+            )
+
+        # Update context on the whole recordset
+        self = self.with_context(date_to=to_datetime)
+
+        result = {}
+        for employee in self:
```
grindtildeath, line 80: For all these functions, wouldn't grouping employees by calendar improve perf instead of iterating through all the employees?

grindtildeath (commented):
Code looks ok but after reviewing it I'm still not sure what this module brings.

project_timesheet_holidays_contract/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+from . import common
```
simahawk, line 1:
```suggestion
```

project_timesheet_holidays_contract/tests/common.py
```diff
@@ -0,0 +1,118 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from datetime import datetime
+
+from odoo import Command
+from odoo.tests import common
+
+
+class ProjectTimesheetHolidaysContractCommon(common.TransactionCase):
```
simahawk, line 12:
```suggestion
class ProjectTimesheetHolidaysContractCommon(common.BaseCase):
```
To have all tracking disabled

project_timesheet_holidays_contract/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+Module allows to take into account employee contract along with provided end date when timesheet on time off.
```
simahawk, line 1: Can you elaborate here a bit?

simahawk (changes requested):

project_timesheet_holidays_contract/tests/__init__.py
yankinmax: why?

project_timesheet_holidays_contract/tests/common.py
yankinmax: To have all tracking disabled:
````
from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT


class ProjectTimesheetHolidaysContractCommon(common.TransactionCase):
    @classmethod
    def setUpClass(cls):
        super().setUpClass()
        cls.env = cls.env(context=dict(cls.env.context, **DISABLED_MAIL_CONTEXT))
````

File: project_timesheet_holidays_contract/models/hr_employee.py (hunk 22 above)
Comment (SilvioC2C) [inline project_timesheet_holidays_contract/models/hr_employee.py:44]: I tested this w/ Odoo standard code, and I think states are not handled correctly.

Contracts can be differentiated in these macro-groups:
- `state == 'close'`: the contract is considered as expired for the employee
- `state == 'cancel'`: the contract is considered as cancelled for the employee
- `state == 'open' and kanban_state != 'blocked'`: the contract is considered as running for the employee
- `state == 'open' and kanban_state == 'blocked'`: the contract is considered as running for the employee, but will be closed as soon as the end date is reached
- `state == 'draft' and kanban_state != 'done'`: the contract is not running for the employee
- `state == 'draft' and kanban_state == 'done'`: the contract is not running for the employee, but will be moved to "open" as soon as the contract's start date is met

This is something that should be taken into account when retrieving the employee's contract, and therefore its calendar.

File: project_timesheet_holidays_contract/readme/DESCRIPTION.md (hunk 26 above)
Comment (rousseldenis) [reply project_timesheet_holidays_contract/readme/DESCRIPTION.md:1]: @yankinmax

Comment (yankinmax) [reply project_timesheet_holidays_contract/readme/DESCRIPTION.md:1]: Hello, @rousseldenis I don't have time now, sorry.

File: project_timesheet_holidays_contract/models/hr_employee.py [hunk 27]
```diff
@@ -0,0 +1,184 @@
...
+
+from odoo import _, fields, models
+from odoo.exceptions import ValidationError
+
+
+class HrEmployee(models.Model):
+    _inherit = "hr.employee"
+
+    def _get_calendar(self, date_from=None):
+        if not date_from:  # no `date_from` => call `super()`
```
leemannd, line 13: Why do we want to call super only when we don't have a `date_from` ?

PR #1291, opened 2024-05-30
[17.0][MIG] project_administrator_restricted_visibility: Migration to 17.0

Abranes:
Standard migration to 17.0


pedrobaeza:
/ocabot migration project_administrator_restricted_visibility

hitrosol (approved):
Functionally tested.
Maybe can we prevent non original PM add additional member ? because it will open to other member that sometimes the original PM doesn't want to share to the one s/he wants to work with ?

jaenbox (approved):
LGTM

rousseldenis (approved):

PR #1296, opened 2024-06-11
[16.0][FIX] project_stock: Avoid AccessError in the form view of tasks for users without permissions in stock.move

victoralmau:
Avoid AccessError in the form view of tasks for users without permissions in `stock.move`

@Tecnativa TT49583

### DIFF
--- a/project_stock/views/project_task_view.xml
+++ b/project_stock/views/project_task_view.xml
@@ -40,13 +40,14 @@
                 />
                 <field name="stock_analytic_date" />
             </field>
-            <field name="stage_id" position="before" groups="stock.group_stock_user">
+            <field name="stage_id" position="before">
                 <button
                     name="action_confirm"
                     string="Confirm materials"
                     type="object"
                     class="oe_highlight"
                     attrs="{'invisible':[('allow_moves_action_confirm','=',False)]}"
+                    groups="stock.group_stock_user"
                 />
                 <button
                     name="action_assign"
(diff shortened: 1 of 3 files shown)


chienandalu (approved):

project_stock/views/project_task_view.xml
```diff
@@ -97,12 +103,14 @@
                 />
             </field>
             <xpath expr="///page[@name='extra_info']" position="before">
+                <!-- Field without groups used as domain in stock_analytic_* fields !-->
+                <field name="use_stock_moves" invisible="1" />
```
pedrobaeza, line 107: No need to move this field.

project_stock/views/project_task_view.xml
```diff
@@ -113,7 +121,6 @@
                         name="move_ids"
                         context="{'tree_view_ref': 'project_stock.view_stock_move_raw_tree', 'form_view_ref':'stock.view_move_form', 'default_company_id': company_id, 'default_state': 'draft', 'default_raw_material_task_id': id}"
                         attrs="{'readonly': [('done_stock_moves', '=', True),('stock_moves_is_locked', '=', True)]}"
-                        groups="stock.group_stock_user"
```
pedrobaeza, line 116: Why removing this group? It shouldn't be unprotected. In any case, just add another invisible occurrence, but it seems another field that requires this one should be protected instead.

pedrobaeza (changes requested):

project_stock/views/project_task_view.xml
victoralmau: Yes it is necessary (the comment tries to clarify this).

The `stock_analytic_*` fields have the analytic groups and use the `use_stock_moves` field in the domain, so it should not have groups.

project_stock/views/project_task_view.xml
victoralmau: The groups is removed because the correct thing to do is to set it at the `<page name=“stock_info”` level.

project_stock/views/project_task_view.xml
pedrobaeza: OK, I understand, as you are placing in the page a group now.

project_stock/views/project_task_view.xml
pedrobaeza: OK as well.

pedrobaeza (approved):
/ocabot merge minor

PR #1308, opened 2024-07-05
[15.0][FIX] project_hr: remove unnecessary migration causing errors

HekkiMelody:
The field `employee_ids` of `project.task` is computed. So, during the regular upgrade process, the m2m relation is populated by the compute method.

When executing the post-migration
1) the m2m relation has already been populated
2) the migration script causes errors of the type:

```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "hr_employee_project_task_rel_pkey"
DETAIL:  Key (project_task_id, hr_employee_id)=(1, 1) already exists.
```


renda-dev (approved):

pedrobaeza (approved):
It's true that the migration is not correct, but let's see if we need to add a pre-migration script to speed up the migration if this compute is very heavy.

PR #1310, opened 2024-07-09
[14.0][MIG] project_task_note: Migration to 14.0

RogerSans:
Standard migration
T-6330

@HaraldPanten
@ValentinVinagre


manuelregidor (changes requested):
Please, remove the odoo version from the commit message.
I think you should also write "backport" instead of "migration"

HaraldPanten:
According this:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

It wouldn't be a problem if you leave [MIG] in the commit message. But I have seen that other contributors have put [BCK] or similar and it has been accepted by other PSC.

@pedrobaeza @etobella  We will remove the Odoo version, for sure, but which should be the "best" commit message extension when migrating a module, when we do it to older versions? IMHO we should keep [MIG].

THX in advance.

etobella:
I don't have an strong opinion on this topic, but I would use [MIG]

pedrobaeza:
Yes, `[BCK]` is not a valid tag, as you can see in the provided link. Some people in the community is not as strict as others and accept "anything" (even with debatable commit messages), but that shouldn't be if we want to have reliable homogeneity across whole community.

A migration commit is a migration commit, no matter if back or forward.

HaraldPanten:
Thanks a lot, guys!

@RogerSans you'll have to delete the Odoo version, as @manuelregidor said, but you'll have to keep [MIG]; as it's the correct tag/extension for migrations.

pedrobaeza:
Little joke: your login seems a True Type font :stuck_out_tongue_closed_eyes:

HaraldPanten:
> Little joke: your login seems a True Type font 😝

That's good, isn't it? 😃

ValentinVinagre (changes requested):
@RogerSans change the name of commit:
[14.0][MIG]project_task_note: Migration to 14.0
by:
[MIG]project_task_note: Migration to 14.0

RogerSans:
Done. Sorry about that

manuelregidor:
@RogerSans The commit name (not the PR title) is still not right. Could you have a look, please? Thank you.

ValentinVinagre (approved):
LGTM 👍🏻

etobella (approved):

PR #1312, opened 2024-07-17
[16.0][MIG] project_task_milestone: Migration to 16.0 

LorenzoC0:


LorenzoC0:
I just noticed that this module is already implemented in Odoo 16 standard.

PR #1313, opened 2024-07-19
[MIG] project_template to 17.0

stenlyk:
All files are original except manifest and views/project.xml (fix invisible attr)


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0.

PR #84, opened 2015-05-06
Extend state and fold statusbar to Issues

dreispt:
- Extends Project Stage State to Project Issues
-  Adds a "Folded on Statusbar" flag, to pick the Stages visible on Tasks and Issue statubars (header pipeline).

:warning:  Depends on #75 (MERGED)

PR #1323, opened 2024-08-08
[16.0][MIG]project_task_recurring_activity: migration to `16.0`

PicchiSeba:


HekkiMelody (approved):
Code review, LGTM

project_task_recurring_activity/models/recurring_activity.py
```diff
@@ -0,0 +1,110 @@
...
+            ):
+                raise UserError(
+                    _(
+                        f"Assigned user {record.user_id.name} has no access"
+                        " to the document and is not able to handle this activity."
+                    )
+                )
+
+    @api.onchange("activity_type_id")
+    def _onchange_activity_type_id(self):
```
rousseldenis: Could you change `user_id`, `description`and `summary`fields to compute/stored/readonly=False fields in order to remove the onchange ? Thanks

In an improvement commit

PicchiSeba: @rousseldenis 
Why should we remove `required=True` from user_id? What's the reason for this change in the new version?

rousseldenis: I didn't speak about required :sweat_smile:

PicchiSeba: Oh damn, I'm sorry i read the wrong thing

project_task_recurring_activity/models/recurring_activity.py
```diff
@@ -66,17 +78,14 @@ def _check_user_id(self):
...
-        self.description = (
-            self.activity_type_id.default_description
-            if not self.description or self.description == "<p><br></p>"
-            else self.description
-        )
-        self.summary = self.summary if self.summary else self.activity_type_id.summary
+    @api.depends("activity_type_id")
+    def _compute_on_activity_type_id(self):
+        if not self.user_id:
+            self.user_id = self.activity_type_id.default_user_id
```
rousseldenis: Could you also add tests to check those values ? thanks

PicchiSeba:
Compared to the previous version, with the deletion of the onchange lines 83-84 of `recurring_activity.py` are never hit. Should we remove the check entirely or restore (at least a part of) the previous behavior?

rousseldenis:
> Compared to the previous version, with the deletion of the onchange lines 83-84 of `recurring_activity.py` are never hit. Should we remove the check entirely or restore part of the previous behavior?

No, that's ok, only the #84. I suppose because user_id is already set. If you want to improve coverage, do a test without assigning a user (or a default ?)

PicchiSeba:
> If you want to improve coverage, do a test without assigning a user (or a default ?)

With the current implementation it is not possible not to assign a user. That's why at the current state those checks are not useful. The issue lies in `required=True`.

Maybe we could remove that requirement from the field but maintain the same behavior.
For example we could add checks in the `write` method that, if not user is set, we use the one defined in `activity_type_id`, and if it doesn't have an user too then we raise an error.

PicchiSeba:
I removed the check for `user_id` because it was unnecessary

HekkiMelody (approved):

project_task_recurring_activity/models/recurring_activity.py
```diff
@@ -66,17 +78,12 @@ def _check_user_id(self):
...
-        )
-        self.description = (
-            self.activity_type_id.default_note
-            if not self.description or self.description == "<p><br></p>"
-            else self.description
-        )
-        self.summary = self.summary if self.summary else self.activity_type_id.summary
+    @api.depends("activity_type_id")
+    def _compute_on_activity_type_id(self):
+        if (not self.description) or self.description == "<p><br></p>":
```
rousseldenis: Please change this to make it multi recordset compliant

rousseldenis (changes requested):

rousseldenis (approved):

coleste (approved):
LGTM

dreispt (approved):

PR #1327, opened 2024-08-21
[16.0][ADD] task_project_status

adasatorres:
- Hi, this addon allow see the project status on task form view.


task_project_status/models/project.py
```diff
@@ -0,0 +1,15 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl-3.0).
+from odoo import fields, models
+
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+    project_status = fields.Many2one(
+        "project.project.stage",
+        string="Project Status",
+        copy=False,
```
flotho, line 11: Hi, thanks for the work.
As your using a related field without store option I don't think you neet copy index ondelete etc...
Yet if your purpose is to be able to group task by project status, you should store the related field and keep those options

flotho (changes requested):

flotho (approved):
LGTM

szalatyzuzanna (approved):

PR #1328, opened 2024-08-21
[16.0][FIX] project_pivot: Fix sequence pivot view

carolinafernandez-tecnativa:
Before this commit, pivot view shows at first place.
With this fix, now it's appear in the order set in action view.

@Tecnativa
@pedrobaeza @carlos-lopez-tecnativa 
TT50574

### DIFF
--- a/project_pivot/readme/ROADMAP.rst
+++ b/project_pivot/readme/ROADMAP.rst
@@ -1,1 +1,1 @@
-This module is not compatible with 'project_list'
+This module is not compatible with 'project_list' and 'project_timeline' because both modules replace the same actions.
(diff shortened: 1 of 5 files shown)


project_pivot/readme/ROADMAP.rst
```diff
@@ -1 +1 @@
-This module is not compatible with 'project_list'
+This module is not compatible with 'project_list' and 'project_timeline'
```
carlos-lopez-tecnativa: Could you document the reason.
```suggestion
This module is not compatible with 'project_list' and 'project_timeline' because both modules replace the same actions
```

carolinafernandez-tecnativa: Done

pedrobaeza (approved):
/ocabot merge patch

PR #1329, opened 2024-08-29
[16.0][ADD] project_scrum

adasatorres:
Hi, this addon allow you to manage your project with scrum methodology.

You can access the sprints and blacklog from the project options on kanban view or from the project settings form view. 

image

On the backlog view you found the all task for this project, except the task with kanban state is done or task with a sprint assigned.

image

On the sprint view you see the all sprint for this project.

image

- The sprint's have a timeline view.
- You can see your task from sprint.
- When a task have a sprint assigned, only can assigned the members on task.

On Sprint Timeline view you can found the task group by sprint.

image

You can change the sprint state with buttons on  sprint form view or using the cron.

Added a new filter on search task for filter using a sprint state.


project_scrum/models/project_sprint.py
```diff
@@ -0,0 +1,100 @@
...
+    name = fields.Char(required=True, track_visibility="onchange")
+    user_ids = fields.Many2many(
+        comodel_name="res.users",
+        string="Members",
+        required=True,
+        domain="[('share', '=', False), ('active', '=', True)]",
+        track_visibility="onchange",
+        relation="project_sprint_user_rel",
+    )
+    objetive = fields.Text(track_visibility="onchange")
```
tarteo: Spelling error

project_scrum/models/project_sprint.py
```diff
@@ -0,0 +1,100 @@
...
+            "context": {
+                "default_project_id": self.project_id.id,
+                "default_sprint_id": self.id,
+            },
+        }
+
+    @api.model
+    def cron_update_sprint_state(self):
+        for sprint in self.search([("state", "=", "draft")]):
+            if fields.Datetime.now() >= sprint.date_start:
```
tarteo: Small note you can do `fields.Datetime.now()` store it in a variable and reuse it every time.

project_scrum/views/project_sprint_views.xml
```diff
@@ -0,0 +1,158 @@
...
+                        <field name="name" />
+                    </h1>
+                    <group>
+                        <group>
+                            <field name="user_ids" widget="many2many_avatar_user" />
+                            <field name="project_id" />
+                        </group>
+                        <group>
+                            <field name="date_start" />
+                            <field name="date_end" />
```
tarteo: Functional suggestion: Would be nice if we can have a select field here with some options like "1 week, 2 weeks, 1 month" which will automatically set the end date

project_scrum/models/project_sprint.py
```diff
@@ -0,0 +1,100 @@
...
+        track_visibility="onchange",
+        required=True,
+    )
+    task_ids = fields.One2many(
+        comodel_name="project.task",
+        inverse_name="sprint_id",
+        string="Tasks",
+        domain="[('project_id', '=', project_id)]",
+    )
+    date_start = fields.Datetime(
```
tarteo: I would have made these fields Date instead of Datetime, but I guess this is a preference.

project_scrum/views/project_views.xml
```diff
@@ -0,0 +1,147 @@
...
+        </field>
+    </record>
+    <record id="project_task_inherit_search_view" model="ir.ui.view">
+        <field name="name">project.task.inherit.search.view</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.view_task_search_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//group" position="before">
+                <separator />
+                <filter
```
tarteo: Functional suggestion: Would be nice if there's a built-in group by sprint, so we can quickly put tasks on a sprint from the backlog.

project_scrum/views/project_views.xml
```diff
@@ -0,0 +1,147 @@
...
+                    name="sprint_in_progress_task"
+                    string="Sprint In Progress"
+                    domain="[('sprint_state', '=', 'in_progress')]"
+                />
+            </xpath>
+        </field>
+    </record>
+    <record id="view_task_timeline" model="ir.ui.view">
+        <field name="model">project.task</field>
```
tarteo: This will override / be overridden the timeline view introduced by `project_timeline`

tarteo (approved):

szalatyzuzanna (approved):

tarteo:
Awesome lets get this merged! @OCA/project-service-maintainers

project_scrum/data/ir_cron_data.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
leemannd: Open question. Should this CRON be in `noupdate` ? This will allow the user to update the frequency without any issue when updating the module

adasatorres: You are right, thanks for the input.

mileo:
Nice work, but it's worth considering that a sprint can sometimes span multiple projects, particularly in contexts where teams manage several projects concurrently or employ scaled agile practices like SAFe. In these cases, linking sprint backlog items to their respective projects can optimize organization and progress tracking across initiatives.

References:

https://scaledagileframework.com/#largesolution
https://scaledagileframework.com/safe-scrum/
https://scaledagileframework.com/iteration-planning/

tarteo:
@mileo So I think what this technically entails is making project_id on project.sprint a many2many. What do you think @adasatorres? Would this be something you like?

Or we can use project_parent for this?

When the field is empty the sprint can be considered company-wide.

project_scrum/views/project_sprint_views.xml
```diff
@@ -0,0 +1,162 @@
...
+                    </h1>
+                    <group>
+                        <group>
+                            <field name="user_ids" widget="many2many_avatar_user" />
+                            <field name="project_id" />
+                        </group>
+                        <group>
+                            <field name="date_start" />
+                            <field name="date_option" />
+                            <field
```
tarteo, line 59: Can you add `force_save="1"` to here, please

tarteo (changes requested):

tarteo:
@adasatorres I made a PR to your branch can you please take a look :) https://github.com/BinhexTeam/project/pull/1

tarteo (approved):

olaf-wagner:
So is this https://github.com/OCA/project-agile/blob/14.0/project_scrum/ officially discontinued now?
It's not a good idea to have two modules with the same name in different OCA repositories.

pedrobaeza:
Oh, I wasn't aware of that one... Maybe we should move this one to there?

olaf-wagner:
Well, if project-agile is not thought to be superfluous, it would rather belong there. But there is already a module with the same name and different code. So they should either be merged (and ideally a migration provided for existing installations) or one has to be renamed (which should be done anyway in order to avoid confusion and update problems with the older module). There is an open PR for the 16.0 port of the older module: https://github.com/OCA/project-agile/pull/39

pedrobaeza:
Please, Binhex team, can you check?

afbenevides:
Hello @pedrobaeza, we are using the https://github.com/OCA/project-agile/ also. We are also in the work of adding epics ( https://github.com/OCA/project-agile/issues/41 ) to that line of agile module. These already take care of sprints, user story ( and storypoints concept ) and now will also get epics (all in 14). Then we will migrate all of it in 16, but there is already some work going on for the migration: https://github.com/OCA/project-agile/pull/39 and https://github.com/OCA/project-agile/pull/38 . Have a look @adasatorres. Please keep us informed so we get a uniform agile line of module in 16 OCA. We would like to contribute to the version that will last throughout versions of odoo like https://github.com/OCA/project-agile/ did since 10.

tarteo:
I don't see the advantages of having the scrum modules in a different repository, IMO we should consolidate them and move functionality that is missing here

adasatorres:
Good evening, first of all, I would like to apologize for the commotion caused by this incident. Secondly, the main idea behind this addon was to create a simpler version, as upon reviewing the *project_scrum* addon in the *project_agile* repository, I found elements that I don't believe are truly necessary or that, personally, don't make sense to me. 

If it is necessary to include this addon in *project-agile*, is it possible to refactor the *project_scrum* from that repository?

olaf-wagner:
The discussion here seems to have stalled. Sorry to bother everybody again, but can we somehow address the situation with the same module name? It is rather unfortunate and has caused some confusion and problems in our CI. Currently we must take care to prune the new module from the OCA repos, as some database instances use the old one and the data models are of course not compatible. It's one of a couple of differences we have to manage regarding OCA repositories, and I really would like to reduce these kind of differences. I'm also not sure if this is the right channel to discuss this; perhaps the discussion was continued at some other place?

PR #1330, opened 2024-08-29
[16.0][FIX] project_timesheet_time_control: Avoid access error with ir.model

victoralmau:
Avoid access error with ir.model

Only "advanced" users have read access to `ir.model`, for this reason, we need to apply `.sudo()`

@Tecnativa


pedrobaeza (approved):
Weird that this didn't happen before. Fast-tracking it:

/ocabot merge patch

PR #1331, opened 2024-08-29
[17.0][FIX] project_timesheet_time_control: Avoid access error with ir.model

victoralmau:
FWP from 16.0: https://github.com/OCA/project/pull/1330

Avoid access error with ir.model

Only "advanced" users have read access to `ir.model`, for this reason, we need to apply `.sudo()`

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #1345, opened 2024-10-01
[14.0][IMP] project_timeline: Add planned_date_* fields to the task tree view

victoralmau:
Add planned_date_* fields to the task tree view

@Tecnativa TT50618

### DIFF
--- a/project_timeline/views/project_task_view.xml
+++ b/project_timeline/views/project_task_view.xml
@@ -136,4 +136,15 @@
             </xpath>
         </field>
     </record>
+    <record id="view_task_tree2" model="ir.ui.view">
+        <field name="name">project.task.tree</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.view_task_tree2" />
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Surprised that the daterange widget works in the tree view and without having both fields shown.

chienandalu (commented):
No always working as expected:

Errors trying to change the dates:


Here the date range didn't work for the date_end:

timeline-3

And here the date end didn't set the date_start:

Peek 01-10-2024 09-40

This one was ok:


Peek 01-10-2024 09-39

victoralmau:
Changed to show fields without `widget=“daterange”` in tree view because in some cases https://github.com/OCA/project/pull/1345#pullrequestreview-2339368831 it does not work correctly (does not save the value of `planned_date_end`)

tarteo (approved):

chienandalu (approved):
Great :+1: , better to have a reliable behavior

PR #1349, opened 2024-10-11
[16.0][ADD] project_tag_security: New module

victoralmau:
New module

@Tecnativa TT51195


project_tag_security/views/project_project_views.xml
```diff
@@ -0,0 +1,15 @@
...
+<odoo>
+    <record id="edit_project" model="ir.ui.view">
+        <field name="name">project.project.form</field>
+        <field name="model">project.project</field>
+        <field name="inherit_id" ref="project.edit_project" />
+        <field name="arch" type="xml">
+            <field name="tag_ids" position="attributes">
+                <attribute
+                    name="domain"
+                >['|', ('allowed_project_ids', 'in', [id]),('allowed_project_ids', '=', False)]</attribute>
```
pedrobaeza, line 11: Does this work?

```suggestion
                >[('allowed_project_ids', 'in', [False, id])]</attribute>
```

project_tag_security/views/project_tags_views.xml
```diff
@@ -0,0 +1,23 @@
...
+            </field>
+        </field>
+    </record>
+    <record id="project_tags_form_view" model="ir.ui.view">
+        <field name="name">project.tags.form</field>
+        <field name="model">project.tags</field>
+        <field name="inherit_id" ref="project.project_tags_form_view" />
+        <field name="arch" type="xml">
+            <field name="color" position="after">
+                <field name="allowed_project_ids" widget="many2many_tags" />
```
pedrobaeza: Put a placeholder saying "Empty for having all the projects allowed" (the same for the other if it's editable)

project_tag_security/views/project_project_views.xml
victoralmau: I assumed you would suggest that but it doesn't work, with that domain you only get tags that have that project as allowed but not tags that have no project.

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #1371, opened 2024-11-20
[17.0][ADD] sale_project_task_recurrency: new module to Configuring Task Recurrence from the Product Form

carlos-lopez-tecnativa:
TT51788
@Tecnativa @pedrobaeza @victoralmau could you  please review this.


pedrobaeza (approved):
I see that the date deadline preserves the hour at which the task is created. Maybe it's better to put a neutral hour like 12:00.

victoralmau (approved):
Code and functional review OK.

PR #1372, opened 2024-11-20
[16.0] [FIX] project_timeline: Add list view on projects

ByteMeAsap:
The module replaced the list view for projects with timeline view. This Pr aims to add the list view back

### DIFF
--- a/project_timeline/views/project_project_view.xml
+++ b/project_timeline/views/project_project_view.xml
@@ -15,12 +15,12 @@
         </field>
     </record>
     <record id="project.open_view_project_all" model="ir.actions.act_window">
-        <field name="view_mode">kanban,timeline,form</field>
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Seeing the original definition, it's true that the tree view is removed:

https://github.com/odoo/odoo/blob/d1be9e4a4ec2cf6eaafe982b5b0036afa5421207/addons/project/views/project_views.xml#L984

https://github.com/odoo/odoo/blob/d1be9e4a4ec2cf6eaafe982b5b0036afa5421207/addons/project/views/project_views.xml#L1004

This probably comes from an outdated definition from previous versions.

Let's fast-track it the obvious patch:

/ocabot merge minor

Can you check if the same happens in 17?

zamberjo (approved):

PR #1374, opened 2024-11-21
[17.0][FIX] project_timeline: Remove overshadowing of date_end functionality

victoralmau:
FWP from 16.0: https://github.com/OCA/project/pull/1369

Default behaviour of Odoo is to set the date_end if the task moves to a folded stage (e.g. done or cancelled). This method I removed overshadows this functionality, which is not needed anymore because of the new planned dates fields

@Tecnativa


houssine78 (approved):
LGTM

pedrobaeza (approved):
/ocabot merge patch

PR #1383, opened 2024-11-27
[18.0] [MIG] project_task_default_stage: Migration to 18.0

Borruso:
Migration from 17.0 to 18.0

Issue: https://github.com/OCA/project/issues/1347


project_task_default_stage/views/project_view.xml
```diff
@@ -28,15 +28,22 @@
...
-                        </tree>
+                            <field
+                                name="rating_template_id"
+                                optional="show"
+                                groups="project.group_project_rating"
+                            />
+                            <field
+                                name="auto_validation_state"
+                                optional="hide"
+                                groups="project.group_project_rating"
```
zamberjo, line 44: The groups was not in previous versions, why add it?

Borruso: I saw list in Odoo and Odoo add this group
I can also remove it there is no problem

zamberjo: No, if it is added by Odoo I see it consistent to apply the same groups.

zamberjo (approved):
Code review

project_task_default_stage/views/project_view.xml
```diff
@@ -10,8 +10,8 @@
             </field>
         </field>
     </record>
-    <record id="task_type_tree" model="ir.ui.view">
-        <field name="name">project.task.type.tree.inherit</field>
+    <record id="task_type_list" model="ir.ui.view">
```
coleste: That doesn't seem right. `id=` should be the same as that of the inherited.

coleste (approved):
LGTM

project_task_default_stage/models/project.py
```diff
@@ -0,0 +1,14 @@
...
+# Copyright 2015 Incaser Informatica S.L. - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ProjectProject(models.Model):
+    _inherit = "project.project"
+
+    def _get_default_type_common(self):
+        ids = self.env["project.task.type"].search([("case_default", "=", True)])
```
rousseldenis: IMHO, ids should be renamed (records or task_types or ...) as it is not ids

project_task_default_stage/models/project.py
```diff
@@ -0,0 +1,14 @@
...
+        return ids
+
+    type_ids = fields.Many2many(default=lambda self: self._get_default_type_common())
```
rousseldenis, line 14: Please put field on top

project_task_default_stage/tests/test_project_task_default_stage.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2015 Incaser Informatica S.L. - Sergio Teruel
+# Copyright 2015 Incaser Informatica S.L. - Carlos Dauden
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo.tests.common import TransactionCase
+
+from odoo.addons.base.tests.common import DISABLED_MAIL_CONTEXT
+
+
+class TestProjectCaseDefault(TransactionCase):
```
rousseldenis: FYI, you can inherit from `BaseCommon` class instead to avoid context line.

rousseldenis (changes requested):
LGTM. Some refinements

carlos-lopez-tecnativa (approved):
Tested, LGTM
TT54238
@Tecnativa @pedrobaeza I think this PR is ready for merge. The comment from @rousseldenis related to placing the field at the top of the file, before the default function, should not be blocking. Let's continue with the merge please if it's OK with you. @rousseldenis, could you update your review, please?
@Borruso, if you can apply the suggestions from @rousseldenis, that would be great. Additionally, I think it's possible to squash some of the administrative commits (not blocking).

pedrobaeza (approved):
Yes, I think we can continue for unblocking this.

/ocabot merge nobump

PR #1385, opened 2024-11-29
[17.0][FIX] project_task_add_very_high: move priority field outside title in sharing project task view form

sanderlienaerts:
Before this commit, the priority field was misplaced in the project's sharing project task view form (portal users editable task view). 
Screenshot 2024-11-29 at 11 17 00

After this commit, the exact solution as used in the webclient's project task view form is used for the editable portal view. This solves the misplacement of the field.
Screenshot 2024-11-29 at 11 11 36

### DIFF
--- a/project_task_add_very_high/views/project_task_view.xml
+++ b/project_task_add_very_high/views/project_task_view.xml
@@ -11,4 +11,13 @@
             </xpath>
         </field>
     </record>
+    <record id="project_sharing_project_task_view_form" model="ir.ui.view">
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.project_sharing_project_task_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//div[hasclass('oe_title')]" position="before">
+                <field name="priority" position="move" />
+            </xpath>
+        </field>
+    </record>
 </odoo>


joepsanders (approved):

coleste (approved):
LGTM

pedrobaeza (changes requested):
In Odoo core, the priority field is placed in the same line as the title:

imagen

With your change, you are wasting a whole line just for the stars. The change should be instead to extend the space for the 3 stars.

Please also don't change the manifest version, as that's done by the bot on merging (and undo also the index.html change).

sanderlienaerts:
Hi @pedrobaeza, this PR recreates the same behavior in the portal view as the form in the webclient for logged in project users. Which, if I am not mistaken, is currently also not following Odoo's core in-line behavior. 

Do you suggest both forms should follow the Odoo core's behavior?

pedrobaeza:
Well, at least the backend form view. Adding 2 stars shouldn't take one vertical line more, making the ergonomics worst.

sanderlienaerts:
All right. I agree with you that it should be in-line.

However testing this bug in v18, it seems the misalignment is no longer present there and debugging why this is happening in v17, is going to be a lot of work (probably an odoo client thing).

Therefore my suggestion would be to implement this fix for v17, so that the portal and back-end views replicate the same implementation and behaviours. Then for the migration to v18, both views adjustments should be removed, so that the priority field is in-line for the back-end and portal view.

v18:
Screenshot 2024-12-19 at 12 30 17

Removing the view adjustments (x2) for v18:
```xml            
<xpath expr="//div[hasclass('oe_title')]" position="before">
   <field name="priority" position="move" />
</xpath>
```

To make sure we're on the same page, this is the current v17 behaviour
### Back-end form
Screenshot 2024-12-19 at 12 38 46

### Portal share view
Screenshot 2024-12-19 at 12 39 22

pedrobaeza:
I don't see this patch brings any enhancement. At the end, we have the same 2 lines before and after. I think a simple CSS override may fix it.

pedrobaeza:
@pilarvargas-tecnativa can you try a bit to fix the situation?

pilarvargas-tecnativa:
> @pilarvargas-tecnativa can you try a bit to fix the situation?

Ok I'll take a look

sanderlienaerts:
Any news @pilarvargas-tecnativa ?

pilarvargas-tecnativa:
No need to modify templates.
Just add this css with a note to remove in v18

```
.o_field_widget.o_priority, .o_field_widget .o_priority {
    white-space: nowrap;
}
```

image

PR #1387, opened 2024-12-05
[16.0][FIX] project_tag_security: Apply specific context to obtain tags correctly in tasks

victoralmau:
Apply specific context to obtain tags correctly in tasks

We define this context to prevent the search_read() method in the project module from filtering project tags in a way we are not interested. Example: applying a limit to indirectly obtain tags to be displayed (problematic if we have +100)

@Tecnativa TT52091


victoralmau:
https://github.com/odoo/odoo/pull/189698

PR #1393, opened 2024-12-16
[FIX] [16.0] project_administrator_restricted_visibility: Restricted Project Manager always can manage its own projects

Shide:
If you change default visibility to "Invited internal users", Project Administrator with restricted access cannot create projects.

This fix adds a new rule to allow them to create projects always (only create).

Once the project is created, the creator of the projects adds automatically to followers and other rules are applied.

MT-7868 @moduon @rafaelbn @Gelojr @fcvalgar @yajo @ernestotejeda please review if you want 😄

### DIFF
--- a/project_administrator_restricted_visibility/tests/test_project_administrator_restricted_visibility.py
+++ b/project_administrator_restricted_visibility/tests/test_project_administrator_restricted_visibility.py
@@ -7,7 +7,6 @@ class TestProjectAdministratorRestrictedVisibility(TransactionCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
-        cls.project_obj = cls.env["project.project"]
         cls.user_admin = cls.env.ref("base.user_admin")
         cls.user_user_padmin = new_test_user(
             cls.env,
@@ -33,12 +32,28 @@ class TestProjectAdministratorRestrictedVisibility(TransactionCase):
             }
         )
 
+    def _create_defaults(self):
+        """Define default field values"""
+        self.env["ir.default"].sudo().create(
+            {
+                "field_id": self.env.ref(
+                    "project.field_project_project__privacy_visibility"
+                ).id,
+                "json_value": '"followers"',
+                "user_id": self.env.user.id,
+            }
+        )
+
     @users("restricted-project-admin", "project-admin")
     def test_create_new_project(self):
         """'Restricted project administrator' can create
         projects like a 'Project administrator'.
         """
-        self.project_obj.create({"name": "Another project"})
+        self._create_defaults()
+        project = self.env["project.project"].create({"name": "Another project"})
+        self.assertEqual(project.privacy_visibility, "followers")
+        project.write({"name": "Another test project"})
+        project.unlink()
 
     @users("restricted-project-admin", "project-user")
     def test_cant_see_restricted_projects(self):
@@ -47,3 +62,9 @@ class TestProjectAdministratorRestrictedVisibility(TransactionCase):
         """
         all_project = self.env["project.project"].search([])
         self.assertNotIn(self.restricted_project, all_project)
+
+    @users("project-admin")
+    def test_can_see_all_projects(self):
+        """'Full project administrator' can see all projects."""
+        all_project = self.env["project.project"].search([])
+        self.assertIn(self.restricted_project, all_project)
--- a/project_administrator_restricted_visibility/README.rst
+++ b/project_administrator_restricted_visibility/README.rst
@@ -81,6 +81,7 @@ Contributors
 * `Moduon <https://www.moduon.team>`_:
 
   * Eduardo López
+  * Eduardo de Miguel
 
 Maintainers
 ~~~~~~~~~~~
--- a/project_administrator_restricted_visibility/readme/CONTRIBUTORS.rst
+++ b/project_administrator_restricted_visibility/readme/CONTRIBUTORS.rst
@@ -6,3 +6,4 @@
 * `Moduon <https://www.moduon.team>`_:
 
   * Eduardo López
+  * Eduardo de Miguel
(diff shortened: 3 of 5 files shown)


pedrobaeza (dismissed):
I don't think this is a good idea, as they should only create projects with the corresponding allowed permissions. Adding a wildcard is not the option.

Shide:
But it's not a wildcard, default visibility should be freely updated at anyone what the customer wants without an impact on any role. When the project is created, the first that Odoo does is to put the creation of the project on the followers (and then the other rules are applied).
This doesn't change anything on the normal flow (except this corner case) and permissions

pedrobaeza:
Yes, but it should be updated by a manager, not a restricted visibility one.

Shide:
But the goal of this module is to make that project admins (that can create projects in any scenario), could do the same things as always, but see only projects that are followers with visibility of internal users (the other visibility options gives internal users all options).

Why Restricted Administrator cannot create projects? This role will be almost the same as Internal User.

There is no specific rule for visibility = followers nor check the user is the project manager.
image

### Other rule options
- `[('user_id', '=', user.id)]`: The user that are creating the project will be the project responsible.
- `[('privacy_visibility', '=', 'followers'), ('message_partner_ids', 'in', [user.partner_id.id])]`: Visibility followers with the user as follower (the responsible in the creation must be the creation user)

### Unsolvable issue (even with the 1=1 rule)
If defaults are: 
- Visibility = internal users 
- **Project Responsible = another user**

Raises the same error and there is no way to handle this because the follower is the responsible at the first time. 

### Recap
- The `[(1, '=', 1)]` on Create rule does the same as the other option rules.
- The `[(1, '=', 1)]` rule it's simple to maintain and it's very clear (means full permissions to create projects and that's all)
- The edge case where the responsible of the project are set by default to an specific user is unsolvable in any case. (This scenario means that only full administrators can create projects, no matter what). With this set-up of defaults, you could give the ability to create projects only to full administrators.

yajo:
> * `[('user_id', '=', user.id)]`: The user that are creating the project will be the project responsible.

This rule is better because it's more UX-consistent than the current one. It can be applied to all CRUD operations, so it will make sure that a project you created you can read it later. Otherwise, the restricted admin could create a project but then never read it, resulting in a false sense of it not having been created.

project_administrator_restricted_visibility/hooks.py
```diff
@@ -10,12 +10,15 @@ def uninstall_hook(cr, registry):
...
-        "project_administrator_restricted_visibility.group_full_project_manager"
+    module_name = "project_administrator_restricted_visibility"
+    group_full_project_manager = env.ref(f"{module_name}.group_full_project_manager")
+    rule_restricted_project_manager = env.ref(
+        f"{module_name}.project_project_restricted_manage_own_project_rule"
     )
     env.ref("project.project_project_manager_rule").write(
         {"groups": [(3, group_full_project_manager.id)]}
     )
+    rule_restricted_project_manager.unlink()
```
yajo: issue: you don't really need to remove this record. The record is namespaced to its module, so Odoo will remove it automatically when uninstalling.

The purpose of these hooks is to undo changes on records that belong to other modules.

So, IUUC, all changes in this file could be removed.

project_administrator_restricted_visibility/tests/test_project_administrator_restricted_visibility.py
```diff
@@ -32,13 +31,25 @@ def setUpClass(cls):
...
         )
+        # Define some default field values
+        cls.env["ir.default"].create(
+            {
+                "field_id": cls.env.ref(
+                    "project.field_project_project__privacy_visibility"
+                ).id,
+                "json_value": '"followers"',
+            }
+        )
```
yajo: suggestion: these defaults would affect all tests, but I think that they are only relevant for the use case you're covering. IMHO it would be better to include it inside the test itself, and not inside the `setUpClass`.

yajo (changes requested):
This looks much better now. I still found one issue though.

yajo (approved):

rafaelbn (approved):
👍🏼 

@pedrobaeza I'm going to merge after all comments solved, OK?

pedrobaeza:
Please wait till my confirmation, as I want to make sure this doesn't provoke any problem in our use cases.

pedrobaeza:
@victoralmau can you please check the impact in our customers?

EmilioPascual (approved):

victoralmau:
Reviewing the whole thread I think the same as pedro, I don't like this solution because without this module (standard) this is a problem that already happens and would be solved in another way (create it with other visibility and changing it later for example) and therefore, you are changing the behavior taking advantage of this permission.

Having clarified the above, I **BELIEVE** that it will not have any side effect and I will not oppose it either.

yajo:
I hope this video lets you understand better why we need this fix and why we think it fits with this modules' users expectations:


https://github.com/user-attachments/assets/5ffb4a9d-111a-4d6c-acaf-76ea8f81b164



> I **BELIEVE** that it will not have any side effect and I will not oppose it either.

So, do you agree to merge?

Thanks.

rafaelbn:
Hello @victoralmau ! Could you please kindly review the video of @yajo after you comment? 😄 

Thank you! ❤️ and Happy new year! 🥳

victoralmau:
Hello again,

Thanks for the explanation video, I understand (and understood) what was happening, and that is why I do not oppose this change and I think it will not have any side effect. Currently it would be possible to create a project with the visibility “Invited portal users and all internal users”, become a follower of that project and then change the visibility to “Invited internal users” (indeed it could not be done currently with the Restricted Administrator permission because it is intentional).

Best regards.

Shide:
So, taking in consideration the @victoralmau response, @pedrobaeza can you approve this PR?

PR #1411, opened 2025-01-07
[18.0][MIG] project_task_material: Migration to 18.0

PierreGillissen:


coleste:
git history was not ported. you need to go through `git format-patch // git am`.

PR #1412, opened 2025-01-13
[FIX] project_timeline: fix migration when source version is 14.0.2.0…

blaskurain:
….0 or newer.


pedrobaeza:
You should check instead if the column exists/not exists instead of a hard version check.

dreispt (approved):

PR #1414, opened 2025-01-15
[18.0][MIG] project_timesheet_time_control

jdebetaz:


project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -0,0 +1,127 @@
...
+
+from dateutil.relativedelta import relativedelta
+
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = "account.analytic.line"
+    _order = "date_time desc"
```
rousseldenis, line 16: I'm always reluctant to change standard `_order` as it can cause problems in side modules.

jguenat (approved):
Reviewed migration commit only, LGTM

project_timesheet_time_control/models/account_analytic_line.py
jguenat: @jdebetaz Agree with that, maybe you could find in the code where date_time desc is needed and use a sorted instead?
do it in another commit in case backport is interesting ?

pedrobaeza: This is needed because we are introducing this new field with datetime precision, so if not done, timesheets from the same day but different time will be unsorted. It should be declared at model level. It shouldn't be changed.

coleste (approved):
LGTM

pedrobaeza (approved):
/ocabot merge nobump

PR #1416, opened 2025-01-20
[FIX] sale_project_copy_tasks: make partner_id field usable in context

yajo:
Without this patch, upgrading the addon raises this error:

    Field 'partner_id' used in context ({'default_partner_id': partner_id, 'default_sale_order_id': active_id}) must be present in view but is missing.

<details>

```
2025-01-20 11:08:33,861 1 INFO ? odoo.modules.loading: Loading module sale_project_copy_tasks (277/285) 
2025-01-20 11:08:34,354 1 INFO ? odoo.modules.registry: module sale_project_copy_tasks: creating or updating database tables 
2025-01-20 11:08:34,456 1 INFO ? odoo.schema: Keep unexpected index product_template_message_main_attachment_id_index on table product_template 
2025-01-20 11:08:34,456 1 INFO ? odoo.schema: Keep unexpected index sale_order_message_main_attachment_id_index on table sale_order 
2025-01-20 11:08:34,456 1 INFO ? odoo.schema: Keep unexpected index sale_order_date_order_index on table sale_order 
2025-01-20 11:08:34,456 1 INFO ? odoo.schema: Keep unexpected index product_product_message_main_attachment_id_index on table product_product 
2025-01-20 11:08:34,482 1 INFO ? odoo.modules.loading: loading sale_project_copy_tasks/views/product_template_views.xml 
2025-01-20 11:08:34,501 1 INFO ? odoo.modules.loading: loading sale_project_copy_tasks/views/sale_order_views.xml 
2025-01-20 11:08:34,523 1 DEBUG ? odoo.tools.convert: while parsing /opt/odoo/auto/addons/sale_project_copy_tasks/views/sale_order_views.xml:5
Ocurrió un error al validar vista cercana

<form string="Sales Order" class="o_sale_order">
            <header>
                <field name="authorized_transaction_ids" invisible="1"/>

Field 'partner_id' used in context ({'default_partner_id': partner_id, 'default_sale_order_id': active_id}) must be present in view but is missing.

View error context:
{'file': '/opt/odoo/auto/addons/sale_project_copy_tasks/views/sale_order_views.xml',
 'line': 1,
 'name': 'sale.order.form.inherit.sale.project.copy.tasks',
 'view': ir.ui.view(5814,),
 'view.model': 'sale.order',
 'view.parent': ir.ui.view(3682,),
 'xmlid': 'view_order_form_inherit_sale_project_copy_taks'}
 
Traceback (most recent call last):
[...]

PR #1418, opened 2025-01-20
[FIX] sale_project_copy_tasks: find the appropriate `project_id` fiel…

yajo:
…d always

This reverts https://github.com/OCA/project/pull/1416 and adds instead a different, more accurate fix for the problem (which kept yielding different errors).

The main source of problem is that the same view is adding a `project_id` field and editing another `project_id` fields' attributes. For this reason, installing the module didn't give any problems, but updating it was problematic.

The new field appeared above in the view hierarchy than the old one. The xpath for the old one was less specific than necessary, and then found the new one instead. The new one is within a tree subview, so the model is different (`sale.order.line` instead of `sale.order`). Thus the fields exposed to that field are different, and failures happened a bit randomly.

@moduon MT-3426

### DIFF
--- a/sale_project_copy_tasks/views/sale_order_views.xml
+++ b/sale_project_copy_tasks/views/sale_order_views.xml
@@ -10,11 +10,10 @@
             ref="sale_project.view_order_form_inherit_sale_project"
         />
         <field name="arch" type="xml">
-            <field name="project_id" position="before">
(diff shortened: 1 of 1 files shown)


Shide (approved):

rafaelbn (approved):
/ocabot merge patch

PR #1420, opened 2025-01-22
[ADD] project_group

etobella:
Allows us to define groups directly on projects.

Usage
=====

Now, when we use Invited internal users options on visibility, a groups
field will appear. Then, all users with access to the project will be
able to see project and related tasks.
image


coleste (approved):
LGTM, simple enough

luisDIXMIT (approved):
Tested on runboat, LGTM!!

PR #1422, opened 2025-01-24
[14.0][IMP] project_forecast_line: sudo calls updating forecast lines

hbrunn:
this is necessary to not interfere with ie portal users given permissions to write some fields on tasks


NL66278 (approved):
:+1: LGTM

PR #1425, opened 2025-01-30
[17.0] [MIG] project_task_link: Migration to 17.0

JordiToledo:


JordiToledo:
We close the PR since Odoo has incorporated it.
see: https://github.com/OCA/OCB/blob/17.0/addons/project/views/project_project_views.xml#L46

PR #1426, opened 2025-01-31
[17.0][MIG] project_parent_task_filter: Migration to 17.0

JordiToledo:
We remove the feature of adding the subtask number in the kanban view as Odoo already includes a similar feature now, allowing the user to open the form view of each subtask from the kanban view.
See: [https://github.com/OCA/OCB/blob/17.0/addons/project/views/project_task_views.xml#L634](https://github.com/OCA/OCB/blob/17.0/addons/project/views/project_task_views.xml#L634)


coleste (approved):
LGTM

anmarmo1 (approved):
Functional review is correct

dreispt (approved):

PR #1440, opened 2025-02-24
[16.0][IMP] project_task_pull_request: Moving PR-URI field

anusriNPS:
Moving PR-URI field in the main section of project task form view similar to v14.
  
  Currently in v16, PR URI field is available under "Extra Info" tab which is visible only by activating debug mode.

  So, moving the field to the main section of task form view for ease of visibility and usage.

### DIFF
--- a/project_task_pull_request/views/project_task_pull_request_view.xml
+++ b/project_task_pull_request/views/project_task_pull_request_view.xml
@@ -7,10 +7,7 @@
         <field name="model">project.task</field>
         <field name="inherit_id" ref="project.view_task_form2" />
         <field name="arch" type="xml">
-            <xpath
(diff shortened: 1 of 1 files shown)


HekkiMelody (approved):

PicchiSeba (approved):
Code review: LGTM

pedrobaeza (changes requested):
The problem with this is that the header takes too much space, reducing the effective notebook height, and making for example Timesheets entry less comfortable.

If changed, this should go into another place, but I think it's OK to have it there, as it's a technical thing.

I also don't conceive using Odoo without `base_technical_features` and the technical features enabled.

anusriNPS:
Closing this PR as not moving the PR URI to header

PR #1444, opened 2025-02-27
[16.0][ADD] project_task_code_portal: use task code

halbtonjazz:
Task codes offer improved clarity for backend users. However, portal users
currently see task IDs instead of human-readable task codes, which can be
confusing and lead to potential issues. This update replaces task IDs with
task codes in the portal interface (both list and detail views) and updates
the search functionality accordingly.

Task: 3241


GabbasovDinar (commented):
LGTM

GabbasovDinar (approved):

project_task_code_portal/controllers/portal.py
```diff
@@ -0,0 +1,97 @@
...
+    ):
+        try:
+            project_sudo = self._document_check_access(
+                "project.project", project_id, access_token
+            )
+        except (AccessError, MissingError):
+            return request.redirect("/my")
+        Task = request.env["project.task"]
+        if access_token:
+            Task = Task.sudo()
```
coleste: I know this code is adapted from Odoo, but is this test enough?
This line from Odoo assumes that all tasks are accessible if you have read rights on the project.

If project is public, `access_token` will not be validated and could be garbage. In which case you can access any of its tasks' attachments.

The `portal_my_task` mecanism above is better and could be adapted here.

halbtonjazz: I agree, you’re right. Thanks for the review, I added an extra check in portal_my_project_task as suggested and did a bit of code refactoring

coleste (approved):
LGTM

BlazejewskaA (approved):
functional LGTM

marcelsavegnago (approved):
LGTM

dreispt (approved):

PR #1445, opened 2025-03-07
[ADD] project_group_hr_timesheet

etobella:
We need to create this rules to make it work with hr_timesheet.

At the end, this is a small glue module.


luisDIXMIT (approved):
LGTM!

coleste (approved):
LGTM

PR #1446, opened 2025-03-13
[17.0][FIX] project_task_default_stage: Do not set user_id on default stages

carlos-lopez-tecnativa:
Odoo has a global record rule that filters records based on user_id. Since this field has a default value of self.env.uid, these records become owned by the Odoo bot. If you try to access a record using `self.env.ref("project_task_default_stage.project_tt_specification").name`, an error occurs.

https://github.com/odoo/odoo/blob/8a2eae27f46247a218facd3b8e27a67161934c8b/addons/project/security/project_security.xml#L122-L126

image

Behavior before the fix: New projects were created without default stages
image

Behavior after the fix: New projects are now created with default stages
image

@pedrobaeza could you please review this

### DIFF
--- a/project_task_default_stage/__manifest__.py
+++ b/project_task_default_stage/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Project Task Default Stage",
     "summary": "Recovery default task stages for projects from v8",
-    "version": "17.0.1.0.0",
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Thanks for the fix. Fast-tracking for solving the problem

/ocabot merge nobump

Issue #1433, opened 2025-02-10
Project_timeline - DuplicateColumn planned_date_end


HugoCLU:
On 16.0, I have this issue when I make a build (below) each time I tried to update my branch. 

## Module
project_timeline

## Describe the bug
bad query: 
    CREATE view report_project_task_user as
         SELECT 
                (select 1) AS nbr,
                t.id as id,
                t.id as task_id,
                t.active,
                t.create_date as create_date,
                t.date_assign as date_assign,
                t.date_end as date_end,
                t.date_last_stage_update as date_last_stage_update,
                t.date_deadline as date_deadline,
                t.project_id,
                t.priority,
                t.name as name,
                t.company_id,
                t.partner_id,
                t.parent_id as parent_id,
                t.ancestor_id as ancestor_id,
                t.stage_id as stage_id,
                t.is_closed as is_closed,
                t.kanban_state as state,
                t.milestone_id,
                pm.is_reached as milestone_reached,
                pm.deadline as milestone_deadline,
                NULLIF(t.rating_last_value, 0) as rating_last_value,
                AVG(rt.rating) as rating_avg,
                t.working_days_close as working_days_close,
                t.working_days_open  as working_days_open,
                t.working_hours_open as working_hours_open,
                t.working_hours_close as working_hours_close,
                (extract('epoch' from (t.date_deadline-(now() at time zone 'UTC'))))/(3600*24)  as delay_endings_days
        ,
                (t.effective_hours * 100) / NULLIF(t.planned_hours, 0) as progress,
                t.effective_hours as hours_effective,
                t.planned_hours - t.effective_hours - t.subtask_effective_hours as remaining_hours,
                NULLIF(t.planned_hours, 0) as hours_planned,
                t.overtime as overtime
        ,
            t.planned_date_begin as planned_date_begin,
            t.planned_date_end as planned_date_end
        ,
            t.planned_date_start,
            t.planned_date_end
           FROM 
                project_task t
                    LEFT JOIN rating_rating rt ON rt.res_id = t.id
                        AND rt.res_model = 'project.task'
                        AND rt.consumed = True
                        AND rt.rating >= 1
                    LEFT JOIN project_milestone pm ON pm.id = t.milestone_id
        
          WHERE 
                t.project_id IS NOT NULL
        
       GROUP BY 
                t.id,
                t.active,
                t.create_date,
                t.date_assign,
                t.date_end,
                t.date_last_stage_update,
                t.date_deadline,
                t.project_id,
[...]

PR #1447, opened 2025-03-13
[15.0][FIX] project_timeline: Allow to install this module with project_enterprise

alan196:
This PR fixes an issue added in commit https://github.com/OCA/project/commit/3689f1dfccf755c5d862c0e779ab0364215dea3a

When the module project_timeline is installed in a database with the module project_enterprise, it raises the following error:

```
psycopg2.errors.DuplicateColumn: column "planned_date_end" specified more than once
```

This is because the field planned_date_end is defined in both modules.

This PR checks if the field is already defined in the query before adding it.

Fixes https://github.com/OCA/project/issues/1433

### DIFF
--- a/project_timeline/report/project_report.py
+++ b/project_timeline/report/project_report.py
@@ -8,14 +8,14 @@ class ReportProjectTaskUser(models.Model):
     _inherit = "report.project.task.user"
 
     planned_date_start = fields.Datetime(readonly=True)
-    planned_date_end = fields.Datetime(readonly=True)
+    planned_date_end_custom = fields.Datetime(readonly=True)
 
     def _select(self):
         return (
             super()._select()
             + """,
             t.planned_date_start,
-            t.planned_date_end"""
+            t.planned_date_end AS planned_date_end_custom"""
         )
 
     def _group_by(self):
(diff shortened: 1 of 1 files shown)


alan196:
@HugoCLU @carlos-lopez-tecnativa Could you help me to review 🙏

project_timeline/report/project_report.py
```diff
@@ -11,18 +11,33 @@ class ReportProjectTaskUser(models.Model):
     planned_date_end = fields.Datetime(readonly=True)
 
     def _select(self):
+        res = super()._select()
+        if "planned_date_end" in res:
```
pedrobaeza: This is too broad. You may have other field `planned_date_end_custom` and do matching.

alan196: I already make the change @pedrobaeza could you help me to review if this is the change you requested 🙏

pedrobaeza:
No, that was not what I requested. What I requested is that you make the `if` sentence more explicit, including quotes, comma or something, to make it match only when the exact column name is found, not in other case.

AaronHForgeFlow:
Superseded in https://github.com/OCA/project/pull/1564

PR #1449, opened 2025-03-20
[18.0][MIG] project_tag_hierarchy: Migration to 18.0

victoralmau:
Migration to 18.0

Please @pilarvargas-tecnativa and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT55549


carlos-lopez-tecnativa (approved):

pilarvargas-tecnativa (approved):

arnaudlayec:
Hello @victoralmau @pilarvargas-tecnativa @carlos-lopez-tecnativa 

Is there any reason why the 3 modules `project_tag_hierarchy`, `project_tag_multicompany` and `project_tag_security` are configured with `"auto_install": True` in there manifest?

I hardly think this is wanted, at least not for everybody.
A simple example: let's say I don't need those modules but I downloaded the OCA/project repo to my Odoo instance. I installed native Odoo `project`, and those 3 modules will install!
Another example: I'm just running tests of any module within OCA/project with `odoo -c test.conf -i project_sequence --test-enable` => the tests are failing because those 3 modules install

Could it be considered to set a more standard `"auto_install": False` for those 3 modules?
From v16.0 to 18.0

Kind regards

victoralmau:
> Hello @victoralmau @pilarvargas-tecnativa @carlos-lopez-tecnativa
> 
> Is there any reason why the 3 modules `project_tag_hierarchy`, `project_tag_multicompany` and `project_tag_security` are configured with `"auto_install": True` in there manifest?
> 
> I hardly think this is wanted, at least not for everybody. A simple example: let's say I don't need those modules but I downloaded the OCA/project repo to my Odoo instance. I installed native Odoo `project`, and those 3 modules will install! Another example: I'm just running tests of any module within OCA/project with `odoo -c test.conf -i project_sequence --test-enable` => the tests are failing because those 3 modules install
> 
> Could it be considered to set a more standard `"auto_install": False` for those 3 modules? From v16.0 to 18.0
> 
> Kind regards

In my opinion, the problem here is not limiting the modules one needs from a repository and leaving all of them available in Odoo.

carlos-lopez-tecnativa:
I didn’t take into account that these modules are set as `auto_install=True`, and I agree with @arnaudlayec, new modules should not be auto-installable by default (which I don’t think is a bad thing either).
However, I’d prefer these not to be auto-installable. What do you think, @pedrobaeza?

pedrobaeza:
Yes, no need of being auto installed

PR #1451, opened 2025-03-20
[18.0][MIG] project_tag_security: Migration to 18.0

victoralmau:
Migration to 18.0

Please @pilarvargas-tecnativa and @carlos-lopez-tecnativa can you review it?

@Tecnativa TT55551


project_tag_security/views/project_task_views.xml
```diff
@@ -0,0 +1,15 @@
...
+    <record id="view_task_form2" model="ir.ui.view">
+        <field name="name">project.task.form</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.view_task_form2" />
+        <field name="arch" type="xml">
+            <field name="tag_ids" position="attributes">
+                <attribute
+                    name="domain"
+                >['|', ('allowed_project_ids', 'in', [project_id]),('allowed_project_ids', '=', False)]</attribute>
+            </field>
```
carlos-lopez-tecnativa: Is it possible to add the domain at the field level?
Or could you add this domain to the list view? The list view is `multi_edit`, and it's not working here.

image

image

carlos-lopez-tecnativa (changes requested):

project_tag_security/views/project_task_views.xml
victoralmau: Remember that this applies only to projects, not to tasks.

carlos-lopez-tecnativa: > Remember that this applies only to projects, not to tasks.

But if you open the task `form` view, the tags are filtered, whereas in the `list` view, this does not happen. To me, this seems inconsistent. Additionally, the inherited view is from `project.task`, which affects tasks, not just projects. If this should only apply to projects, why inherit from `project.task`?

victoralmau: Sorry for the confusion, you are right. I see that this is something that happened also in v16 (this way of editing it had not been taken into account).

I have tried to define the domain in the fields and it works, thanks.

project_tag_security/models/project_project.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+from odoo import fields, models
+
+
+class Projectproject(models.Model):
+    _inherit = "project.project"
+
+    tag_ids = fields.Many2many(
+        domain="""['|', ('allowed_project_ids', 'in', [id]),
+        ('allowed_project_ids', '=', False)]"""
```
carlos-lopez-tecnativa, line 11: Could you please try this to simplify the domain and avoid an `OR`?
```suggestion
        domain="""[('allowed_project_ids', 'in', [id] + [False])]"""
```

carlos-lopez-tecnativa (approved):
LGTM, just a minor change if you think it's applicable. However, it's okay for me. Thanks!

project_tag_security/models/project_project.py
victoralmau: I just tried it but it doesn't work (a pity), if you find something similar in core please indicate it for reference. It seems that only something similar works in the rules.

carlos-lopez-tecnativa: I imagine this is evaluated as JS, so I'm not sure if using `[0]` or `[false]` will work.

Anyway, let's go!

project_tag_security/README.rst
```diff
@@ -0,0 +1,98 @@
...
+Project tags are limited to the desired projects.
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
```
pilarvargas-tecnativa, line 40: 1. Go to Project > Configuration > Tags

pilarvargas-tecnativa (approved):

marcelsavegnago (approved):
LGTM

PR #1454, opened 2025-03-26
[18.0][MIG] project_stock > project_task_stock: Migration to 18.0

victoralmau:
Migration to 18.0

@Tecnativa TT55547


project_stock/models/project_project.py
```diff
@@ -34,13 +34,15 @@ class ProjectProject(models.Model):
 
     @api.onchange("picking_type_id")
     def _onchange_picking_type_id(self):
-        self.location_id = self.picking_type_id.default_location_src_id.id
-        self.location_dest_id = self.picking_type_id.default_location_dest_id.id
+        self.location_id = self.picking_type_id.default_location_src_id
```
pedrobaeza: No need to change this, or is there any side effect I don't know?

project_stock/models/project_project.py
```diff
@@ -34,13 +34,15 @@ class ProjectProject(models.Model):
...
+        self.location_dest_id = self.picking_type_id.default_location_dest_id
 
     def write(self, vals):
         """Update location information on pending moves when changed."""
         res = super().write(vals)
         field_names = ("location_id", "location_dest_id")
         if any(vals.get(field) for field in field_names):
-            self.task_ids._update_moves_info()
+            self.task_ids.filtered(
```
pedrobaeza: Why limiting this? Isn't there the possibility of having a desynchronization in any moment due to this?

project_stock/models/project_task.py
```diff
@@ -74,7 +75,6 @@ class ProjectTask(models.Model):
         help="Move created will be assigned to this analytic account",
     )
     stock_analytic_distribution = fields.Json(
-        "Analytic Distribution",
```
pedrobaeza: If you change this label, maybe others should be adjusted as well to reflect the "Stock" part.

project_stock/models/project_task.py
```diff
@@ -121,17 +121,25 @@ def _compute_stock_state(self):
...
+        price_dp = self.env["decimal.precision"].precision_get(
+            "Product Unit of Measure"
+        )
         for item in self:
             already_reserved = item.mapped("move_ids.move_line_ids")
-            any_quantity_done = any([m.quantity_done > 0 for m in item.move_ids])
-            item.unreserve_visible = not any_quantity_done and already_reserved
+            any_quantity = any(
+                [
+                    not float_is_zero(m.quantity, precision_digits=price_dp)
```
pedrobaeza: Are we allowing negative quantities?

project_stock/models/project_task.py
```diff
@@ -121,17 +121,25 @@ def _compute_stock_state(self):
...
+                    for m in item.move_ids.filtered(lambda x: x.state == "done")
+                ]
+            )
+            item.unreserve_visible = not any_quantity and already_reserved
 
     @api.onchange("picking_type_id")
     def _onchange_picking_type_id(self):
-        self.location_id = self.picking_type_id.default_location_src_id.id
-        self.location_dest_id = self.picking_type_id.default_location_dest_id.id
+        self.location_id = self.picking_type_id.default_location_src_id
```
pedrobaeza: The same about this change that was asked previously.

project_stock/models/project_project.py
victoralmau: There is no side effect, but I think the right thing to do is not to define the id (as it happens in other odoo core examples).

pedrobaeza: I think Odoo doesn't have any guideline about that, so the normal trend is to keep minimum the diff.

project_stock/models/project_task.py
victoralmau: The string has been removed to avoid the warning.

`WARNING devel odoo.addons.base.models.ir_model: Two fields (stock_analytic_distribution, analytic_distribution) of project.task() have the same label: Analytic Distribution. [Modules: project_stock and analytic]`

Defining "Stock Analytic Distribution" would also cause the pre-commit error, therefore, "Analytic Distribution" has been defined in the view.

project_stock/models/project_task.py
victoralmau: I hadn't thought about negative amounts, actually. 
The quantity done (`quantity`) cannot be defined, so IMO it is not a "real" possibility.

pedrobaeza: Well, I think the previous `> 0` is enough then.

project_stock/models/project_task.py
pedrobaeza: Yes, I was talking about expanding this for having proper names in the search window.

project_stock/models/project_project.py
victoralmau: Sorry, no need, I was testing a use case that generated confusion and auto-changed the stock move state causing a side effect not taken into account.

project_stock/models/project_task.py
victoralmau: Ok, to clarify, you want to indicate in the `stock_analytic_distribution` field the string "Stock Analytic Distribution" even if that is what will be defined by default (and pre-commit will display an error). 

Do you also want to change any other field (`stock_analytic_line_ids` for example with “Stock Analytic Lines”)?

pedrobaeza: Yeah, I mean other fields like "Analytic date" to prefix "Stock" to be clear its usage. The other day we were confused for what it's used that field, and more being in so prominent place in the main block, and not in a group inside a tab. I would say it's convenient to move all these fields to the "Additional information" tab.

victoralmau:
Please @pilarvargas-tecnativa can you review it?

project_stock/models/project_task.py
```diff
@@ -62,7 +62,7 @@ class ProjectTask(models.Model):
         index=True,
         check_company=True,
     )
-    stock_analytic_date = fields.Date(string="Analytic date")
+    stock_analytic_date = fields.Date(string="Stock Analytic date")
```
pilarvargas-tecnativa: Doesn't precommit warn you that it is redundant?

victoralmau: Oops, pre-commit does not give error because the string is not “Stock Analytic **D**ate”, actually that would be the right one, so I will remove the string.

pedrobaeza: But please change translations then.

pedrobaeza:
We have to re-examinate the core feature and if still needed, rename the module: https://github.com/OCA/OCB/tree/18.0/addons/project_stock

victoralmau:
> We have to re-examinate the core feature and if still needed, rename the module: https://github.com/OCA/OCB/tree/18.0/addons/project_stock

The `project_stock` module has a different behavior and does not serve as a dependency or cover the need for this module, so it is still needed. Maybe it is appropriate to rename it to `project_task_stock`, what do you think?

pedrobaeza:
OK then

pedrobaeza (approved):
Code review

pilarvargas-tecnativa (approved):

PR #1458, opened 2025-04-03
[18.0][FIX] project_timesheet_time_control: hide show_time_control

david-banon-tecnativa:
Forward-port of #1457

@Tecnativa @pedrobaeza @sergio-teruel 
TT55789

PR #1459, opened 2025-04-03
project_timesheet_time_control : invisible to column_invisible

leblanc-simon:
Since Odoo 17.0, in tree view, we must use column_invisible instead of invisible : https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0#:~:text=On%20a%20tree%20view%2C%20if%20you%20have%20a%20field

This merge fix this change for project_timesheet_time_control


leblanc-simon:
This change can be backported to 17.0 branch.

Should I create a new merge request on the 17.0 branch or are you handling this differently?

pedrobaeza:
Already handled in #1458 and previously merged to 17.0. A pity that both have worked at the same time the same fix, but the important thing is to fix the issue :)

PR #1465, opened 2025-04-04
[17.0][IMP] project_key - improvement on the display_name for the projects and tasks

jdidderen:
Before the change for the tasks : 
image

After the change for the tasks :
image

### DIFF
--- a/project_key/models/project_project.py
+++ b/project_key/models/project_project.py
@@ -30,6 +30,14 @@ class Project(models.Model):
             else:
                 rec.key = ""
 
+    @api.depends("key", "name")
+    def _compute_display_name(self):
+        super()._compute_display_name()
+        for project in self:
+            if project.key:
+                project.display_name = f"[{project.key}] {project.display_name}"
+        return
+
     @api.model_create_multi
     def create(self, vals_list):
         for vals in vals_list:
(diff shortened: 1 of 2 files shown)


coleste (approved):
LGTM

project_key/models/project_project.py
```diff
@@ -30,6 +30,14 @@ def _onchange_project_name(self):
             else:
                 rec.key = ""
 
+    @api.depends("key", "name")
+    def _compute_display_name(self):
+        for project in self:
+            if project.key:
+                project.display_name = f"[{project.key}] {project.name}"
+            else:
+                project.display_name = f"{project.name}"
```
yostashiro: Can you please check if the following suggestion is okay? This is to avoid conflicts with other modules that manipulate with the display name.
```suggestion
        super()._compute_display_name()
        for project in self:
            if project.key:
                project.display_name = f"[{project.key}] {project.display_name}"
        return
```

The same applies to the method for `project.task`.

jdidderen: Good catch. I have adapted the code based on your suggestion. Thanks 👍

yostashiro (approved):
Code review.

AungKoKoLin1997 (approved):

jdidderen:
@OCA/project-service-maintainers Would it be possible to have this merged ?

AaronHForgeFlow (approved):
Code review LGTM

gurneyalex (approved):

PR #1469, opened 2025-04-09
[17.0][IMP] project_task_related: show only same client tasks by default

david-banon-tecnativa:
Adds default search filter due to customer request
@Tecnativa @pedrobaeza @juancarlosonate-tecnativa

TT55793


pedrobaeza (commented):
We use customer, not client as terminology.

david-banon-tecnativa:
Done @pedrobaeza

pedrobaeza (approved):
Fast-tracking, as the module just landed in the repo, and it's an incremental improvement.

/ocabot merge patch

PR #1472, opened 2025-04-10
 [17.0][MIG] project_task_name_with_id: Migration to 17.0

lbarry-apsl:
Module migrated to version 17.0

cc https://github.com/APSL 7983
@miquelalzanillas @javierobcn @mpascuall @BernatObrador @ppyczko @peluko00 please review

https://github.com/OCA/project/issues/1191


mpascuall (approved):
LGTM!

peluko00 (approved):
LGTM!

BernatObrador (changes requested):
This isn't working correctly.

Example from version 16:
image

Current behavior:
image

The ID is no longer inside brackets [ ], now the name is inside instead. And the name it's between ' '. Something seems to be wrong with this.

Also, when you create a new task, it ends up like this:
image

lbarry-apsl:
> This isn't working correctly.
> 
> Example from version 16: image
> 
> Current behavior: image
> 
> The ID is no longer inside brackets [ ], now the name is inside instead. And the name it's between ' '. Something seems to be wrong with this.
> 
> Also, when you create a new task, it ends up like this: image

Done @BernatObrador !!

BernatObrador (approved):
LGTM!

project_task_name_with_id/tests/test_project_task_id.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2023 Quartile Limited
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import re
+
+from odoo.tests.common import TransactionCase
+
+
+class TestProjectTaskID(TransactionCase):
+    def setUp(self):
```
BernatObrador: Just a minor comment, test should be using @classsmethod
```suggestion
    @classmethod
    def setUpClass(cls):
```

project_task_name_with_id/models/project_task.py
```diff
@@ -0,0 +1,28 @@
...
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    @api.depends("name")
+    def _compute_display_name(self):
+        for task in self:
+            parts = [f"[{task.id}]"]
+            if hasattr(task, "key") and task.key:
+                parts.append(task.key)
```
yostashiro, line 16: Wasn't sure about the validity of this part. Shouldn't we add `super()._compute_display_name()` at the beginning to adjust the updated `display_name`, instead?

project_task_name_with_id/__manifest__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2023 Quartile Limited
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+{
+    "name": "Project Task Name with ID",
+    "category": "Project",
+    "version": "17.0.1.0.1",
```
yostashiro:
```suggestion
    "version": "17.0.1.0.0",
```

project_task_name_with_id/tests/test_project_task_id.py
lbarry-apsl: Done!

project_task_name_with_id/__manifest__.py
lbarry-apsl: Done!

project_task_name_with_id/models/project_task.py
lbarry-apsl: @yostashiro  this is to make the module compatible with project_key without depending on it. Or do you think it's not necessary?

yostashiro: @lbarry-apsl IMO, we shouldn't add special logic just to be "compatible" with another module. Would there be any issue if we simply did the following?

```py
    @api.depends("name")
    def _compute_display_name(self):
        for task in self:
            task.display_name = f"[{task.id}] {task.display_name}"
```

lbarry-apsl: There wasn't an error, just that the task.key wasn't displayed in the ticket name added by the project_key module.

@yostashiro if it's not a problem, I can leave it as you've told me.

lbarry-apsl: > @lbarry-apsl IMO, we shouldn't add special logic just to be "compatible" with another module. Would there be any issue if we simply did the following?
> 
> ```python
>     @api.depends("name")
>     def _compute_display_name(self):
>         super()._compute_display_name()
>         for task in self:
>             task.display_name = f"[{task.id}] {task.display_name}"
> ```

@yostashiro  I have done the tests with what you propose here, but the result with the project_key is this and it is not correct ``[5] 5['OD-1', 'Energy Certificate']``

lbarry-apsl:
Hi @pedrobaeza , can you check this migration?

project_task_name_with_id/models/project_task.py
yostashiro: > I have done the tests with what you propose here, but the result with the project_key is this and it is not correct `[5] 5['OD-1', 'Energy Certificate']`.

@lbarry-apsl I think that's because how `project_key` doesn't use `super()` in `_compute_display_name()`. Likewise, the current implementation of this module may break the presentation of the display name if there is another module which manipulates with the field.

lbarry-apsl: Okay, I understand. I mentioned in this PR #1503 that it adds the `super()` function without modifying the logic.

AungKoKoLin1997: @lbarry-apsl @yostashiro I created PRs to avoid complete overidding.
https://github.com/OCA/project/pull/1502
https://github.com/OCA/project/pull/1503

PR #1482, opened 2025-04-24
[16.0][ADD] project_task_add_very_high: implement for portal

hbrunn:
before, the portal showed only one star, now we see three:

image


coleste (approved):
LGTM

dreispt (approved):
I would say this is a [IMP] or even a fix [FIX], but code review :+1:

PR #1486, opened 2025-04-25
[17.0][IMP] project_timesheet_time_control: default start date configurable and improved time calculation

CRogos:
- live value calculation in form. If any of these values is changed, the other are updated accordantly. 
image

- start date configurable to be "now" or directly "after the previous entry" if available, otherwise use employee working hour as start date
image

@pilarvargas-tecnativa, @david-banon-tecnativa could you please review? Feedback apricated.

### DIFF
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -1,10 +1,13 @@
 # Copyright 2016 Tecnativa - Antonio Espinosa
 # Copyright 2016 Tecnativa - Sergio Teruel
 # Copyright 2016-2018 Tecnativa - Pedro M. Baeza
+# Copyright 2025 glueckkanja AG - Christopher Rogos
+
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from datetime import datetime
+from datetime import datetime, time, timedelta
 
+import pytz
 from dateutil.relativedelta import relativedelta
 
 from odoo import _, api, fields, models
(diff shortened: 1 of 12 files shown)


david-banon-tecnativa (changes requested):
You should add tests for this new functionality and update the [project_timesheet_time_control.pot](https://github.com/OCA/project/blob/17.0/project_timesheet_time_control/i18n/project_timesheet_time_control.pot) file.

david-banon-tecnativa:
Functional Review OK, works as expected

CRogos:
@david-banon-tecnativa thank you for the quick response and review. I've now added unit tests and updated the .dot file.

david-banon-tecnativa (approved):

CRogos:
@lbarry-apsl could you have a look on this change as well? This might also be a benefit when using hr_timesheet_calendar.

lbarry-apsl:
The "Time Connecting" configuration is not working correctly for me or I haven't tested it properly.

Can you explain what the configuration has to do?

CRogos:
When you create a new account.analytic.line and there is not an explicit date_time and date_time_end. (e.g. only a duration submitted) than this change should check, when did the previous entry of the same employee of the same day ended and uses this timestamp as the start of the new account.analytic.line. 

But indeed you found an issue, because this should start at 08:00:00 and not at 00:26:00.
image

I'll have a look.

CRogos:
I am not sure what was actually the reason for the behavior. Locally I could not reproduce it and I assume this was caused by the demo data. I did some minor improvements, and now it is also working in runboat.

image

This change also updates the End Time and Hours Spent when changing the other values.
[62c4a847-4185-40f6-a5e8-a97c0373a56c.webm](https://github.com/user-attachments/assets/02fe5691-93a3-4423-8bbe-e020686c28e1)

lbarry-apsl:
I've tried it again. The previous option now works fine.

Is it normal for the Day and Half-day settings to use the previous entry hours?

CRogos:
Good question. I now leave this option also available for Day/half-days:
image

CRogos:
@lbarry-apsl could you do another review?

lbarry-apsl:
I've tried it again, but now I would say that it's not working as expected.

Set Start Date to Now: It's taking the previous time, and Align to previous entry: It's taking the current time.

It's as if they're working backwards.

CRogos:
For me it is working in runbot. To make it more clear I've deleted all existing time entries.

https://github.com/user-attachments/assets/7386fcc7-47c1-49fe-89a0-c34e99fd96a0

lbarry-apsl (approved):
Correct functionality tested on the Runboat

lbarry-apsl:
> For me it is working in runbot. To make it more clear I've deleted all existing time entries.
> Screen.Recording.2025-05-22.144556.mp4

Now I can see the results. Good job!

CRogos:
@pedrobaeza your feedback is always welcome, otherwise merge.

pedrobaeza (changes requested):
For me this is overcomplicating the module and affecting the performance of each `account.analytic.line` creation, and it should go through the default method that is doing lots of things. This should go in a separate module on top of the current one.

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -15,8 +18,63 @@ class AccountAnalyticLine(models.Model):
...
+        timesheet_alignment = params.get_param(
+            "project_timesheet_time_control.timesheet_alignment"
+        )
+        # default to now
+        now = fields.Datetime.now()
+        start_time = datetime.combine(
+            now.date(), time(hour=now.hour, minute=now.minute, second=0)
+        )
+        if timesheet_alignment == "now":
+            return start_time
```
CRogos, line 35: @pedrobaeza this is the default case when using the setting as before. I agree it is more than only fields.Datetime.now(), but is this really that performance critical?

I could also move the def _get_default_start_time(self) to the hr_timesheet_calendar module? 
(I am not very happy to create for every code line a new module)

pedrobaeza: You have to read each time the ICP, compute the dates, etc, and the analytic lines are also created for moves, invoices... so I think it's heavy. I have customers with a very high number of these lines.

CRogos: Would moving _get_default_start_time(self) to the hr_timesheet_calendar module fine for you?

pedrobaeza: Well, I don't know about that module, and its link with this one.

CRogos: I'll prepare a new PR and reconsider if/what I would leave in this module. Thx for your feedback.

PR #1487, opened 2025-04-29
[18.0][MIG] project_merge: Migration to 18.0

Anushapadinharepurakkal:


project_merge/wizards/project_task_merge.py
```diff
@@ -51,44 +72,29 @@ def merge_tasks(self):
...
             self.task_ids.filtered(lambda t: t.description).mapped(
-                lambda task: "Description from task <b>%s</b>:<br/>%s"
-                % (task.name, task.description)
+                lambda task: f"Description from task <b>"
+                f"{task.name}</b>:<br/>{task.description}"
             )
         )
 
-    def _merge_followers(self, merged_tasks):
+    def _get_merge_followers(self, merged_tasks):
```
coleste: Since this method does not return a value, it seems weird to call it `get`.

Would it make more sense to call it `_subscribe_merged_followers` or something similar?

dreispt (commented):
I agree with coleste's comment.

PR #1489, opened 2025-04-30
[18.0][MIG] project_task_parent_completion_blocking: Migration to version 18.0

david-banon-tecnativa:
TT56201

@tecnativa @pedrobaeza  @victoralmau


pedrobaeza (approved):

project_task_parent_completion_blocking/tests/test_project_task_parent_completion_blocking.py
```diff
@@ -0,0 +1,66 @@
+from odoo.exceptions import ValidationError
+from odoo.tests.common import TransactionCase
+
+
+class TestParentChildBlock(TransactionCase):
```
victoralmau:
```suggestion
from odoo.addons.base.tests.common import BaseCommon


class TestParentChildBlock(BaseCommon):
```

With this change it is no longer necessary to use `with_context(mail_create_nolog=True)` (or similar).

victoralmau (changes requested):

victoralmau (approved):

PR #1490, opened 2025-05-02
[18.0][MIG] project_stock_product_set > project_task_stock_product_set: Migration to 18.0

victoralmau:
Migration to 18.0

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT55548


project_task_stock_product_set/readme/CONFIGURE.md
```diff
@@ -0,0 +1,5 @@
+It is important to clarify some concepts at the level of project stages:
```
pedrobaeza, line 1: This change from RST to MD must be in the `pre-commit autofixes` commit, not in the migration one.

pedrobaeza (changes requested):

project_task_stock/tests/common.py
```diff
@@ -7,7 +7,7 @@
 from odoo.addons.base.tests.common import BaseCommon
 
 
-class TestProjectStockBase(BaseCommon):
+class TestProjectTaskStockBase(BaseCommon):
```
pedrobaeza: Why renaming this here? It's not needed at all, and it messes a bit the commit diff.

victoralmau: Because the unwritten rules (or at least mine) mean that it must be consistent with the module name, and otherwise it would be confusing with an existing module (`project_stock`) in core; anyway, it is already changed.

pedrobaeza: Yeah, but now that it was not done in the previous migration PR, let's keep it for not having cross-changes in this migration PR.

pedrobaeza (approved):

pilarvargas-tecnativa (approved):

PR #1495, opened 2025-05-09
[16.0][FIX] project_stock: Generate analytical lines in both use cases (tasks and pickings)

### LINKED ISSUE #1492: project_task_stock 18.0 (used to be project_stock in 16.0) does not create analytic items.
## Module
The name of the module that has a bug.

project_task_stock 18.0 (used to be project_stock in 16.0)

## Describe the bug

Analytic items for consumed products on task are not created when picking is validated from Inventory app.
When "Transfer materials" button from Task is used, analytic items are created correctly.

## To Reproduce
**Affected versions**:
16.0
18.0

1. Install project_task_stock (project_stock) on a fresh instance
2. Configure according to module instructions
3. Create project with task in stage that allows Stock pickings (Stock info tab)
4. Add products to task and confirm them (have to have cost > 0)
5. Validate stock picking through Inventory app

6. No analytic items created!!!

**Expected behavior**

6. Analytic items should have been created

victoralmau:
Generate analytical lines in both use cases (tasks and pickings)

In addition to generating the analytical lines at the end of the task, they must also be generated if the pickings is done.

Fixes #1492

Please @pedrobaeza and @etoimene can you review it?

@Tecnativa

### DIFF
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -1,4 +1,4 @@
-# Copyright 2022 Tecnativa - Víctor Martínez
+# Copyright 2022-2025 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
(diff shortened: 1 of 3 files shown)


project_stock/models/project_task.py
```diff
@@ -220,13 +220,7 @@ def action_done(self):
...
         ):
             move.quantity_done = move.reserved_availability
-        moves_todo = self.mapped("move_ids")._action_done()
-        # Use sudo to avoid error for users with no access to analytic
-        analytic_line_model = self.env["account.analytic.line"].sudo()
-        for move in moves_todo:
-            vals = move._prepare_analytic_line_from_task()
-            if vals:
-                analytic_line_model.create(vals)
+        self.mapped("move_ids")._action_done()
```
pedrobaeza: No need of the `mapped` with relational fields.

```suggestion
        self.move_ids._action_done()
```

pedrobaeza (approved):

pedrobaeza:
Also tested by @etoimene, so merging

/ocabot merge patch

PR #1496, opened 2025-05-09
[18.0] [MIG] project_purchase_link: Migration to 18.0

bizzappdev:


project_purchase_link/models/project_project.py
```diff
@@ -0,0 +1,180 @@
...
+            purchase_invoice_line_total = 0
+            for group in groups:
+                purchase_invoice_line_total += group["price_subtotal"]
+            project.purchase_invoice_count = len(groups)
+            project.purchase_invoice_line_total = purchase_invoice_line_total
+
+    def button_open_purchase_order(self):
+        self.ensure_one()
+        return {
+            "name": _("Purchase Order"),
```
AaronHForgeFlow:
```suggestion
            "name": self.env._("Purchase Order"),
```

project_purchase_link/models/project_project.py
```diff
@@ -0,0 +1,180 @@
...
+            "domain": self._domain_purchase_order(),
+            "type": "ir.actions.act_window",
+            "view_mode": "list,form",
+            "res_model": "purchase.order",
+        }
+
+    def button_open_purchase_order_line(self):
+        self.ensure_one()
+        return {
+            "name": _("Purchase Order Lines"),
```
AaronHForgeFlow:
```suggestion
            "name": self.env._("Purchase Order Lines"),
```

project_purchase_link/models/project_project.py
```diff
@@ -0,0 +1,180 @@
...
+        domain = expression.AND(
+            [safe_eval(action.domain or "[]"), self._domain_purchase_invoice()]
+        )
+        action_dict.update({"domain": domain})
+        return action_dict
+
+    def button_open_purchase_invoice_line(self):
+        self.ensure_one()
+        return {
+            "name": _("Purchase Invoice Lines"),
```
AaronHForgeFlow:
```suggestion
            "name": self.env._("Purchase Invoice Lines"),
```

AaronHForgeFlow (commented):
Functional review ok! Just some minor comments

AaronHForgeFlow (approved):
ok to me!

project_purchase_link/models/project_project.py
```diff
@@ -0,0 +1,180 @@
...
+        purchase_invoice_lines_ids = [
+            int(record.get("id")) for record in self._cr.dictfetchall()
+        ]
+        domain = [("id", "in", purchase_invoice_lines_ids)]
+        return domain
+
+    def _compute_purchase_info(self):
+        for project in self:
+            groups = self.env["purchase.order.line"].read_group(
+                project._domain_purchase_order_line(),
```
acsonefho, line 119: If we have many project, that could generate a huge amount of query.
IMO this should be improved to build a global domain before the loop and also execute the read_group.
Something like that:
```
from odoo.osv import expression
...
main_domain = expression.OR([record._domain_purchase_order_line() for record in self])
global_read_group = self.env["purchase.order.line"].read_group(main_domain, ["order_id"], ["price_subtotal"])
for project in self:
    ...
```
This could be used also into others compute of this file

project_purchase_link/models/project_project.py
```diff
@@ -0,0 +1,180 @@
...
+            "name": self.env._("Purchase Order Lines"),
+            "domain": self._domain_purchase_order_line(),
+            "type": "ir.actions.act_window",
+            "view_mode": "list,form",
+            "res_model": "purchase.order.line",
+        }
+
+    def button_open_purchase_invoice(self):
+        self.ensure_one()
+        action = self.env.ref("account.action_move_in_invoice_type")
```
acsonefho: `action = self.env["ir.actions.act_window"]._for_xml_id("account.action_move_in_invoice_type")`

project_purchase_link/tests/test_project_purchase_link.py
```diff
@@ -0,0 +1,109 @@
...
+        self.purchase.write({"invoice_status": "to invoice"})
+        self.assertFalse(self.project.purchase_count)
+        self.assertFalse(self.project.purchase_line_total)
+        self.assertFalse(self.project.purchase_invoice_count)
+        self.assertFalse(self.project.purchase_invoice_line_total)
+        self.purchase.write(
+            {
+                "order_line": [
+                    (
+                        0,
```
acsonefho: You should replace by `Command.create(...)`

acsonefho (changes requested):

project_purchase_link/tests/test_project_purchase_link.py
```diff
@@ -0,0 +1,109 @@
+# Copyright 2019 Oihane Crucelaegui - AvanzOSC
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo.osv import expression
+from odoo.tests import common
+from odoo.tools.safe_eval import safe_eval
+
+
+class TestProjectPurchaseUtilities(common.TransactionCase):
```
acsonefho: You should use `BaseCommon`
`from odoo.addons.base.tests.common import BaseCommon`

acsonefho (changes requested):

project_purchase_link/models/project_project.py
bizzappdev: @acsonefho Using a global domain is not feasible in this case because the project does not have a direct relationship with the purchase.order.line model. If a global domain is applied, it will include the price_subtotal values from all projects, not just the relevant one.

coleste (approved):
LGTM

project_purchase_link/models/project_project.py
dreispt: @acsonefho Why is your suggestion better?

project_purchase_link/models/project_project.py
dreispt: Resolving conversation. PRs fixing the performance concern will be welcome.

dreispt (approved):

project_purchase_link/models/project_project.py
bizzappdev: In the previous version, it was getting the browse object and then read the record, while _for_xml_id is directly giving the dict with needed values. FYI : https://github.com/odoo/odoo/blob/122dece7eeedbf670254aebd2a2d69642b381547/odoo/addons/base/models/ir_actions.py#L227C23-L227C39

coleste:
@acsonefho, can you confirm your issues are solved?

acsonefho (approved):

acsonefho:
> @acsonefho, can you confirm your issues are solved?

Yes :muscle: Thank you for your work! Let's merge :partying_face:

PR #1355, opened 2024-10-22
[MIG] project_task_code: Migration to 17.0

nedaszilinskas:

PR #1499, opened 2025-05-12
[17.0][MIG] project_task_code

etobella:
Superseed of #1355 

Restore the kanban as it was....

image


pedrobaeza:
There's a merge commit inside the commit history. You should do a rebase and remove it.

AaronHForgeFlow (commented):
It works ok!

project_task_code/models/project_task.py
```diff
@@ -0,0 +1,48 @@
...
+                vals["code"] = (
+                    # `sudo()` for portal users
+                    self.env["ir.sequence"].sudo().next_by_code("project.task") or "/"
+                )
+        return super().create(vals_list)
+
+    @api.depends("name", "code")
+    def _compute_display_name(self):
+        for task in self:
+            task.display_name = f"[{task.code}] {task.name}" if task.code else task.name
```
yostashiro: Can we make it like this?
```suggestion
        super()._compute_display_name()
        for task in self:
            if task.code:
                task.display_name = f"[{task.code}] {task.display_name}"
        return
```

Related: #1502 #1503

etobella: @yostashiro you are completly right. I made a small change to adapt to your suggestion :wink:

project_task_code/tests/test_project_task_code.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2016 Tecnativa <vicent.cubells@tecnativa.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import odoo.tests.common as common
+
+
+class TestProjectTaskCode(common.TransactionCase):
```
carlos-lopez-tecnativa: Please inherit from [BaseCommon](https://github.com/odoo/odoo/blob/1b23d3b9607ad8adb7f4e2685a0bd41df9d2cc26/odoo/addons/base/tests/common.py#L21)

project_task_code/tests/test_project_task_code.py
```diff
@@ -0,0 +1,73 @@
...
+    def setUp(self):
+        super().setUp()
+        self.project_task_model = self.env["project.task"]
+        self.ir_sequence_model = self.env["ir.sequence"]
+        self.task_sequence = self.env.ref("project_task_code.sequence_task")
+        self.project_task = self.env.ref("project.project_1_task_1")
```
carlos-lopez-tecnativa: Why isn’t this done in `setUpClass`?

carlos-lopez-tecnativa (approved):
LGTM, just a few minor changes if you consider them applicable.

TT54672

project_task_code/tests/test_project_task_code.py
etobella: Because it is a migration and usually, I don't want to modify tests too much on migrations. However, in this case, it makes sense to do the suggested changes :wink:

pedrobaeza:
Please check CI

etobella:
Fixed! It was a collision with another module. The way to solve it was to check the name by regex :wink:

yostashiro (approved):
LGTM.

PR #1500, opened 2025-05-12
[18.0][FIX] project_task_stock: Generate analytical lines in both use cases (tasks and pickings)

### LINKED ISSUE #1492: project_task_stock 18.0 (used to be project_stock in 16.0) does not create analytic items.

victoralmau:
FWP from 16.0: https://github.com/OCA/project/pull/1495

Generate analytical lines in both use cases (tasks and pickings)

In addition to generating the analytical lines at the end of the task, they must also be generated if the pickings is done.

Fixes #1492

Please @pedrobaeza and @etoimene can you review it?

@Tecnativa

### DIFF
--- a/project_task_stock/models/project_task.py
+++ b/project_task_stock/models/project_task.py
@@ -231,14 +231,6 @@ class ProjectTask(models.Model):
         ):
             move.quantity = move.product_uom_qty
         moves_to_do.picking_id.with_context(skip_sanity_check=True).button_validate()
-        moves_done = self.move_ids.filtered(lambda x: x.state == "done")
-        moves_todo = moves_done - moves_to_skip
-        # Use sudo to avoid error for users with no access to analytic
-        analytic_line_model = self.env["account.analytic.line"].sudo()
-        for move in moves_todo:
-            vals = move._prepare_analytic_line_from_task()
-            if vals:
-                analytic_line_model.create(vals)
 
     def action_see_move_scrap(self):
         self.ensure_one()
--- a/project_task_stock/models/stock_move.py
+++ b/project_task_stock/models/stock_move.py
@@ -1,4 +1,4 @@
-# Copyright 2022 Tecnativa - Víctor Martínez
+# Copyright 2022-2025 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 from odoo import api, fields, models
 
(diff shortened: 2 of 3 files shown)


pedrobaeza (approved):

etoimene:
I'll give it a try today.

etoimene:
I am getting invalid Stock moves message.
image

I don't see source and destination locations on a product, just operation.
Could that be a cause?

etoimene:
Operation is defined as previously ..
image

Created a fresh DB to test.

etoimene:
Had to add operation locations on the Extra info tab. 

Previously it would have picked locations from the project but now I can only define operation.
Looks like source and destination locations a re missing.
image

etoimene:
Analytics are created correctly when stock settings are manually edited one the task.
Project stock settings are not propagated to the task correctly.

victoralmau:
> Had to add operation locations on the Extra info tab.
> 
> Previously it would have picked locations from the project but now I can only define operation. Looks like source and destination locations a re missing. image

As in v16, the user must have the “_Manage Multiple Stock Locations_” permission to be able to see the locations (Source Location + Destination Location) in the project. https://github.com/OCA/project/blob/996f2f2a30bf6c8fc756798560966d512bcb5046/project_stock/views/project_project_view.xml#L13

etoimene:
image
I have the privilege.

Check the picture in https://github.com/OCA/project/pull/1500#issuecomment-2871758176 !

Stock part is missing the locations and they are not propagated to stock move on the task unless I explicitly define them on Task Extra Info tab.

etoimene:
I have refreshed or restarted test server in the meantime and all is as one expect.

I would approve PR.

PR #1510, opened 2025-05-26
[16.0][FIX] project_key: consider inactive projects for key uniqueness

HekkiMelody:
This small commit fixes a bug where inactive projects would not be considered (on installation or later) when creating a unique key

But the SQL constraint on key uniqueness would still trigger an error later on.


PicchiSeba (approved):
Code review: LGTM

rvalyi (approved):

dreispt (approved):

PR #1512, opened 2025-05-27
[16.0][BACKPORT] project_task_tag: Migration to 16.0

maisim:
backport from 17.0 https://github.com/OCA/project/tree/17.0/project_task_tag


dreispt (approved):

coleste (approved):
LGTM, no changes

leemannd:
Hello, this is a duplicate of https://github.com/OCA/project/pull/1159
A review on the original PR would have it merged.

dreispt (changes requested):
It seems that this is  NOT a migration, but a new 16.0 module, also proposed at #1159?
And it includes a commit for 17.0 migration ... but it target 16.0...
Very confusing.
It needs clarification.

maisim:
> It seems that this is NOT a migration, but a new 16.0 module, also proposed at #1159? And it includes a commit for 17.0 migration ... but it target 16.0... Very confusing. It needs clarification.

Hi @dreispt 
It is a backport from 17.0 https://github.com/OCA/project/tree/17.0/project_task_tag
I hadn't seen the other PR, though, it would be a bit odd to have 2 modules with the same name in 2 different branches that aren't iso. We sould look deeper at #1159 before to choose to merge it

PR #1438, opened 2025-02-18
[17.0][MIG] project_analytic_code

petrus-v:
migration from https://github.com/OCA/project/pull/815

PR #1517, opened 2025-06-02
[FIX] project_analytic_code: fix tests

dreispt:
Supersedes #1438


dreispt:
Testes are failing. Superseded by #1517

petrus-v:
Hi @dreispt !

Thanks for taking care of conflict.

Just to let you know that in v14 project_sequence is declared as rebel module (cf `rebel_module_groups` in `.copier-answers.yml`, the former PR has been done 10 days before project_sequence was merged, maybe the solution is to declare it in the same way ?

There is probably a conflict as far I remember !

petrus-v:
I've rebase the former PR https://github.com/OCA/project/pull/1438 and manage the conflict by declaring some rebel modules, I think we can close this one !

project_analytic_code/tests/test_project.py
```diff
@@ -0,0 +1,37 @@
...
+                "code": "Test Weird code",
+                "plan_id": cls.plan.id,
+            }
+        )
+
+    def test_search_project_by_account_analytic_code(self):
+        result = self.env["project.project"].name_search(
+            name=self.project.analytic_account_id.code
+        )
+        expected = f"[{self.project.analytic_account_id.code}] {self.project.name}"
```
alexey-pelykh, line 28: `name_search` returns a list of `(id, display_name)` tuples in Odoo 17, not recordsets. `result[0].display_name` will raise an error -- the existing code on 17.0 (`result[0][1]`) is the correct way to access the name from the result.

alexey-pelykh (changes requested):
Thanks for the effort on this, but this PR should be closed. The original migration PR #1438 was already rebased and merged to 17.0 on Dec 18, 2025 -- as @petrus-v noted in the comments. Additionally, the test "fix" here introduces a bug: `name_search` in Odoo 17 returns `(id, display_name)` tuples, so the current `result[0][1]` on 17.0 is correct; `result[0].display_name` would fail since you cannot access attributes on a tuple. CI failures confirm this.

PR #1519, opened 2025-06-03
[16.0][ADD] project_scoped_tags

maisim:
New module to modify tags behavior for projects and tasks, inspired by GitLab’s tag scoping.

This allow to limit tag availability by projects

Sould be used with "Optional quick create" module

This project differs from project_task_tag, which requires applying the tags you want to use on tasks to the project itself.
Here, you define the tasks available for the project and its tasks.

This is a start; we can later easily add the ability to filter by project type, make tags specific to tasks, etc.


mourad-ehm (approved):
Code review

maisim:
[project_tag_security](https://github.com/OCA/project/tree/16.0/project_tag_security) do the same

PR #1521, opened 2025-06-09
[16.0][FEAT]portal: new module to hide timesheet from portal

quirino95:
This module hides details about timesheets and worked hours on tasks in the portal pages.


project_portal_hide_timesheet/hooks.py
```diff
@@ -0,0 +1,26 @@
...
+
+from odoo import SUPERUSER_ID, api
+
+_logger = logging.getLogger(__name__)
+
+views_to_switch = [
+    "hr_timesheet.portal_tasks_list_inherit",
+    "hr_timesheet.portal_timesheet_table",
+    "hr_timesheet.portal_my_task",
+    "sale_timesheet.portal_my_task_inherit",
```
HekkiMelody, line 11: **issue**: module `sale_timesheet` needs to be added to the module dependencies

project_portal_hide_timesheet/hooks.py
```diff
@@ -0,0 +1,26 @@
...
+]
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    _logger.info("Turning off views")
+    for view in views_to_switch:
+        env["ir.ui.view"].search([("key", "=", view)]).write({"active": False})
```
HekkiMelody: **suggestion** (non-blocking): 

```suggestion
    env["ir.ui.view"].search([("key", "in", views_to_switch)]).write({"active": False})
```

project_portal_hide_timesheet/hooks.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def uninstall_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    _logger.info("Turning on views")
+    for view in views_to_switch:
+        env["ir.ui.view"].search([("key", "=", view)]).write({"active": True})
```
HekkiMelody: **issue**: the uninstall_hook doesn't work, most likely because `search` cannot find inactive records by default. Tested in runboat.

Alternatively to the search `env.ref` can be used, but either method is fine.

Please also add a test for the hooks.

HekkiMelody (changes requested):
For adding a new module, the commit message should be

`[ADD] project_portal_hide_timesheet: new module`

(since the name of the module is already descriptive enough)

project_portal_hide_timesheet/hooks.py
quirino95: Applied suggestion

project_portal_hide_timesheet/hooks.py
quirino95: Added dependency

project_portal_hide_timesheet/hooks.py
quirino95: Improved still using `search`. Add tests for the hooks.

project_portal_hide_timesheet/views/project_portal_hide_timesheet.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template
+        id="portal_home_timesheet"
```
HekkiMelody: **chore**:

Some formal requirements according to guidelines https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

1. File name should be <model_name>_views.xml (I think `_templates.xml` is also acceptable here since there are templates)
2. These two views should be split on two files since one is related to the `timesheet.sheet` model and the other is related to `project.task`
3. When inheriting, the `id` of the new view should be the same as the original view (minus the module name), so `id="portal_my_home_timesheet"` and `id="portal_tasks_list"`

HekkiMelody (commented):
Minor formal changes but otherwise LGTM

HekkiMelody (approved):

HekkiMelody:
@OCA/project-service-maintainers We've been using this successfully in production for a couple of months now. Could you please take a look? Thanks!

dreispt:
Init hooks are a surprising way to modify views. It sounds dangerous.
I expected QWeb and view extensions to be used for that.

alexey-pelykh (commented):
Thanks for this contribution. The module approach is clean -- using `post_init_hook`/`uninstall_hook` to toggle existing portal views is a well-established OCA pattern, and the XML overrides are minimal and focused.

I reviewed the final commit (`e5135a4`) and the full review history. HekkiMelody's earlier feedback (missing `sale_timesheet` dep, loop-to-`in` operator, inactive record search, tests, XML split) has all been addressed. CI is green.

One item I'd flag:

**Test coverage is vacuous (false positive)**

In `test_hooks.py`, `setUpClass` fetches views with:

```python
cls.views = cls.env["ir.ui.view"].search([("key", "in", views_to_switch)])
```

By the time tests run, `post_init_hook` has already fired during module installation, so these views are `active=False`. Since `search()` applies `active_test=True` by default, `cls.views` is an **empty recordset**.

Both assertions then pass vacuously because `all()` on an empty iterable returns `True`:

```python
self.assertTrue(all(not v.active for v in self.views))  # True (empty)
self.assertTrue(all(v.active for v in self.views))       # True (empty)
```

To make the tests meaningful, the search in `setUpClass` should include inactive records:

```python
cls.views = cls.env["ir.ui.view"].with_context(active_test=False).search(
    [("key", "in", views_to_switch)]
)
```

And assertions should also verify the recordset is non-empty, e.g.:

```python
def test_post_init_hook(self):
    post_init_hook(self.env.cr, self.env.registry)
    self.assertTrue(self.views)
    self.assertTrue(all(not v.active for v in self.views))
```

Otherwise the hooks could be completely broken and tests would still pass.

Everything else looks good -- manifest, dependencies, license, XML structure, OCA scaffolding are all correct.

SirPyTech:
> Init hooks are a surprising way to modify views. It sounds dangerous.
> I expected QWeb and view extensions to be used for that.

Thanks @dreispt for having a look!
Indeed hooks are not the usual way, I thought the same when I first saw this, but note that the hooks are only disabling the views, just as the user would do if they didn't want them.

I tried using `xpath`s but some of these views are adding several nodes here and there, and although hiding them all is possible, it results in a lot of `xpath`s in the new inherited view that could be hard to maintain.

We could save one hook adding the `active="False"` attribute in the views having `id="original_module.original_view"`, but then reverting them on module uninstall is still needed and the list of views to activate would be far in the XML, so it's not easy to check that all the deactivated views have been activated back.

Please let me know what you think!

---
> **Test coverage is vacuous (false positive)**

@alexey-pelykh it was indeed, I fixed it, please check again.

SirPyTech (approved):

BhaveshHeliconia (approved):
LGTM!

PR #1522, opened 2025-06-09
[16.0][FEAT]portal: show task only if follower

quirino95:
This module changes behavior of task visibility: portal users now will only see task where they are explicitly set as follower, so visibility on project only is no longer enough.
However, when creating a new task, all users following the project will automatically follow the new task too.


project_portal_task_visibility/hooks.py
```diff
@@ -0,0 +1,24 @@
...
+rule_to_switch = (
+    "Project/Task: portal users: (portal and following project) "
+    "or (portal and following task)"
+)
+
+
+def post_init_hook(cr, registry):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    _logger.info("Turning off rule")
+    env["ir.rule"].search([("name", "=", rule_to_switch)]).write({"active": False})
```
HekkiMelody: **issue**: Using the name is very fragile and prone to error, especially if someone uses a default language that's different from English. Let's use the XML ID to find the rule, and `env.ref()` to get it

project_portal_task_visibility/security/project_portal_task_visibility.xml
```diff
@@ -0,0 +1,24 @@
...
+        <field name="perm_unlink" eval="False" />
+        <field name="groups" eval="[(4, ref('base.group_portal'))]" />
+    </record>
+<!--    <record id="project.project_task_rule_portal" model="ir.rule">-->
+<!--        <field name="domain_force">[-->
+<!--        ('project_id.privacy_visibility', '=', 'portal'),-->
+<!--        ('active', '=', True),-->
+<!--        ('message_partner_ids', 'child_of', [user.partner_id.commercial_partner_id.id]),-->
+<!--        ]</field>-->
+<!--    </record>-->
```
HekkiMelody: **chore**: leftover code

HekkiMelody (changes requested):
For adding a new module, the commit message should be

`[ADD] project_portal_task_visibility: show task only to followers`

HekkiMelody:
Tests are failing because of incompatibility with `project_task_code_portal`. There are a few ways we could possibly address it, none of them are very elegant.

One of them could be to change the setup of the tests of `project_task_code_portal` to activate the rule again just in case. (with a comment describing why)

project_portal_task_visibility/hooks.py
quirino95: Used `env.ref()` instead of search by name

quirino95:
> Tests are failing because of incompatibility with `project_task_code_portal`. There are a few ways we could possibly address it, none of them are very elegant.
> 
> One of them could be to change the setup of the tests of `project_task_code_portal` to activate the rule again just in case. (with a comment describing why)

Forced the rule to active in `setUpClass` method, now tests are ok

project_portal_task_visibility/security/project_portal_task_visibility.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
HekkiMelody: **chore**: file name should be `project_task_security.xml`

HekkiMelody (approved):
Minor formal requirement, but LGTM

HekkiMelody (approved):

HekkiMelody:
@OCA/project-service-maintainers We've been using this successfully in production for a couple of months now. Could you please take a look? Thanks!

leemannd (approved):

PR #1526, opened 2025-06-16
[17.0][IMP] project_hr: restore odoo default user_ids domain

Abranes:
By default, in Odoo tasks, when you try to assign a user, the field has the domain `[('share', '=', False), ('active', '=', True)]`. However, with this module, a new domain is applied based on the employee category. If no category is selected, all users are shown, which breaks the standard behavior.

With this I restore by default the Odoo domain.


pedrobaeza (approved):
Fast-tracking it.

/ocabot merge patch

PR #1530, opened 2025-06-23
[16.0][IMP] project_key: improved key visualization

quirino95:
Added two improvements about `key` field visualization:

1) Due to an obsolete class (`oe_read_only`), key was never displayed near the task title. I deleted the class and now the key is shown correctly:
immagine

2) I added a warning if in the key there is a dash, because it could work bad with the module itself (e.g. renaming key):
immagine

### DIFF
--- a/project_key/models/project_project.py
+++ b/project_key/models/project_project.py
@@ -15,6 +15,8 @@ class Project(models.Model):
 
     key = fields.Char(size=10, required=False, index=True, copy=False)
 
+    show_key_warning = fields.Boolean(store=False, compute="_compute_show_key_warning")
+
     _sql_constraints = [
         ("project_key_unique", "UNIQUE(key)", "Project key must be unique")
     ]
@@ -206,3 +208,7 @@ class Project(models.Model):
 
             for task in project.task_ids:
                 task.key = project.get_next_task_key()
+
+    @api.depends("key")
+    def _compute_show_key_warning(self):
+        self.show_key_warning = self.key and "-" in self.key
--- a/project_key/views/project_key_views.xml
+++ b/project_key/views/project_key_views.xml
@@ -9,8 +9,18 @@
         <field name="model">project.project</field>
         <field name="inherit_id" ref="project.edit_project" />
         <field name="arch" type="xml">
+            <field name="company_id" position="before">
+                <div
+                    class="alert alert-warning"
+                    role="alert"
+                    attrs="{'invisible': [('show_key_warning', '=', False)]}"
+                >
+                    Using '-' in project keys is not recommended.
+                </div>
+            </field>
             <field name="partner_id" position="before">
                 <field name="key" required="1" />
+                <field name="show_key_warning" invisible="1" />
             </field>
         </field>
     </record>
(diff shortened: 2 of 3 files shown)


project_key/models/project_project.py
```diff
@@ -206,3 +208,10 @@ def _set_default_project_key(self):
 
             for task in project.task_ids:
                 task.key = project.get_next_task_key()
+
+    @api.onchange("key")
+    def _onchange_key(self):
+        if "-" in self.key:
+            self.show_key_warning = True
+        else:
+            self.show_key_warning = False
```
HekkiMelody: **suggestion**(non-blocking): 
```suggestion
        self.show_key_warning = "-" in self.key
```

project_key/models/project_project.py
```diff
@@ -15,6 +15,8 @@ class Project(models.Model):
 
     key = fields.Char(size=10, required=False, index=True, copy=False)
 
+    show_key_warning = fields.Boolean(store=False)
```
HekkiMelody: **suggestion**: It's probably best to use a computed non-stored field here, instead of `onchange`

project_key/views/project_key_views.xml
```diff
@@ -9,8 +9,18 @@
         <field name="model">project.project</field>
         <field name="inherit_id" ref="project.edit_project" />
         <field name="arch" type="xml">
+            <field name="company_id" position="before">
+                <div
+                    class="alert alert-warning oe_edit_only"
```
HekkiMelody: **chore**: in Odoo 16, `oe_edit_only` and `oe_read_only` shouldn't be used anymore, since we're basically always in edit mode.

project_key/views/project_key_views.xml
```diff
@@ -42,7 +52,8 @@
         <field name="inherit_id" ref="project.view_task_form2" />
         <field name="arch" type="xml">
             <field name="name" position="before">
-                <field name="key" readonly="1" nolabel="1" class="oe_read_only" />
+                <field name="key" readonly="1" nolabel="1" class="oe_inline" />
+                <span class="oe_inline"> - </span>
```
HekkiMelody: **nitpick**: `-` is already used in the key itself at least once. Maybe we could use another character as separator between key and task title to make it clearer?
I'm thinking `:`, or alternatively enclose the key in `[]` like with product code? What do you think?

HekkiMelody (changes requested):
Functionally works well.
Code review with some minor suggestions.

project_key/models/project_project.py
quirino95: Better, I've also managed if key is `False`.

project_key/views/project_key_views.xml
quirino95: Using `:`, it's simple and linear with the task name.

HekkiMelody (approved):

coleste (commented):
LGTM

Warning could be variable by making `show_key_warning = fields.Char` and setting it "Using '-' in project keys is not recommended" or whatever other message is required. The current implementation is therefore easy to improve.

coleste (approved):

HekkiMelody:
@OCA/project-service-maintainers Can you take a look at this? Thanks

rvalyi (approved):

PR #1409, opened 2025-01-03
[18.0][MIG] project_task_code

wpichler:
Standard Migration - based on https://github.com/nedaszilinskas/project/tree/17.0-mig-project_task_code

PR #1532, opened 2025-06-25
[18.0][MIG] project_task_code: Migration to 18.0

carlos-lopez-tecnativa:
Supersedes https://github.com/OCA/project/pull/1409
TT54672
@Tecnativa @pedrobaeza @victoralmau could you please review this?


pedrobaeza (approved):
Initially, I thought the module was incorrect, as I got this view in the kanban:

imagen


but it's another module causing the problem, as on the DB with only base and installing just this module, everything is OK:

imagen

so there's a problem in other of the migrated modules (probably `project_key`, which I don't like too much).

Anyway, as this was already approved in the superseded PR, merging:

/ocabot merge nobump

PR #1534, opened 2025-07-01
[ADD] project_fte: New module project_fte

miquelpascual:
With this module you can manage FTE (Full-Time Equivalent) contracts and evolution in projects.

cc https://github.com/APSL 13148
@miquelalzanillas @lbarry-apsl @javierobcn @peluko00 @BernatObrador @ppyczko please review


peluko00 (approved):
LGTM!

lbarry-apsl:
Captura desde 2025-07-02 08-45-37

The end date is required, but it's set to read-only on the form. It can't be set.

Did I miss any steps?

miquelpascual:
> Captura desde 2025-07-02 08-45-37
> 
> The end date is required, but it's set to read-only on the form. It can't be set.
> 
> Did I miss any steps?

Hi, thanks for your comment! Yes, as explained in the README, you need to set the lines, and the end date will be computed automatically.

lbarry-apsl (approved):
Code and runboat reviewed, LGTM

miquelalzanillas (approved):
Functional review. LGTM.

ppyczko (approved):
LGTM

leemannd:
Hello @mpascuall could you have a look at this -> https://github.com/OCA/project/pull/1534#issuecomment-3670400760

Once fixed I will merge the PR

alexey-pelykh (changes requested):
Thanks for this new module -- the FTE tracking concept is solid and the test coverage is good. I have a number of findings that should be addressed before merge.

---

## Bugs / Correctness

### 1. Direct assignment to computed stored field `executed_hours`

In `account_analytic_line.py`, the `unlink` method directly assigns:
```python
fte_line.executed_hours -= line.unit_amount
```
`executed_hours` is a computed stored field (computed by `_compute_executed_hours` in `project_fte_month_line.py`). Direct assignment to a stored compute field bypasses the compute method and can lead to inconsistencies -- for example, if the compute method is later triggered by a dependency change, it will overwrite this manual adjustment. The `_update_fte_month_lines` method used in `create`/`write` correctly invalidates the field (by searching and triggering recomputation). The `unlink` path should follow the same pattern: invalidate the field so the compute method recalculates, rather than doing arithmetic directly.

### 2. `action_copy_last_fte_line` crashes on empty FTE lines

In `project_project.py`:
```python
sorted_fte_lines = self.fte_month_line_ids.sorted(...)
last_fte_line = sorted_fte_lines[0]
```
If `fte_month_line_ids` is empty, this raises an `IndexError`. The button in the view should either be hidden when there are no FTE lines (via `invisible` attribute), or the method should guard against an empty list.

### 3. Cron warning logic skips overload check when low-execution fires

In `_cron_check_fte_execution`, the `continue` after the low-execution warning (< 50% at mid-month) means that if execution later jumps above 100% in the same cron run, the overload warning will never fire for that line because the low-execution branch already matched and continued. Consider removing the `continue` statements and using `elif` instead, or reordering the checks so the highest-severity condition (> 100%) is checked first.

---

## Translation / i18n

### 4. Module-level `_()` calls do not translate at runtime

`MSG_NO_EXEC` and `MSG_WITH_EXEC` are wrapped in `_()` at module load time:
```python
MSG_NO_EXEC = _(
    "<p>This project has ..."
)
```
In Odoo, `_()` at module level captures the translation at import time using the server's default language, not the user's language. These constants should be plain strings, and `_()` should be called at the point of use:
```python
MSG_NO_EXEC = (
    "<p>This project has <b>%(fixed)s ..."
)
# Then at the point of use:
line.fte_line_message = _(MSG_NO_EXEC) % {...}
```
Note: the usage site already calls `_(MSG_NO_EXEC)`, which means `_()` is called twice -- once at import (wrong) and once at use (right but on an already-translated string). Just remove the module-level `_()` wrapping.

### 5. Missing `.pot` template file

Only `i18n/es.po` is provided. OCA convention requires a `i18n/project_fte.pot` template file for other translators to use as a base.

---

## Security

### 6. HTML field with `sanitize=False` -- potential XSS vector

`monthly_html_table` in `project_project.py` is defined with `sanitize=False` and is built using f-strings that include `dist.role_id.name` without escaping:
```python
f"<td>{dist.role_id.name}</td>"
```
If a role name contains HTML/JavaScript, it will be rendered unsanitized. Consider using `markupsafe.Markup` and `markupsafe.escape` for user-provided values, or set `sanitize=True` (Odoo's sanitizer preserves safe HTML table markup).

---

## Code Quality

### 7. Missing copyright/license headers

The following files are missing the `# Copyright 2025 APSL Nagarro` and `# License AGPL-3.0 or later (...)` header that is present in the other module files:
- `models/account_analytic_line.py`
- `models/project_role.py`
- `models/project_milestone.py`

OCA guidelines require the license header in every Python file.

### 8. Redundant unlink in `action_delete_fte`

```python
for line in self.fte_month_line_ids:
    line.profile_distribution_ids.unlink()
self.fte_month_line_ids.unlink()
```
`profile_distribution_ids` has `ondelete="cascade"` on its `month_line_id` Many2one, so unlinking the month lines will automatically cascade-delete the distributions. The explicit distribution unlink is redundant and can be removed.

### 9. Convoluted percentage calculation

In `project_fte_profile_distribution.py`:
```python
dist.profile_hours_percentage = (
    (dist.profile_hours * 100) / total_hours
) / 100
```
This multiplies by 100 then divides by 100, which simplifies to just `dist.profile_hours / total_hours`. Use the simpler form for clarity.

### 10. `.sudo()` usage should be documented

`_update_fte_month_lines` and `unlink` in `account_analytic_line.py` use `.sudo()` to search FTE month lines. Since timesheet users may not have access to FTE records, sudo is likely intentional, but it should have a comment explaining why it is needed.

---

## Typos

### 11. "Montly" should be "Monthly"

Found in `project_project_views.xml`, `project_fte_mass_generator_views.xml`, and `es.po`:
```xml
<separator string="Montly Summary" />
```
Should be `"Monthly Summary"`.

### 12. "inhetir" should be "inherit" in XML record ID

In `project_role.xml`:
```xml
<record id="project_role_view_form_inhetir_project_fte" ...>
```
Should be `project_role_view_form_inherit_project_fte`.

---

## Minor Suggestions (non-blocking)

- `_compute_executed_hours` uses `search().mapped('unit_amount')` followed by `sum()`. For large datasets, `read_group` would be more efficient (single SQL aggregation vs. loading all records).
- `allocated_hours` is defined as `compute + store=True + readonly=False` which is an unusual pattern (user-editable computed field). If this is intentional, a comment explaining the design choice would help future maintainers.

lbarry-apsl:
Hello @leemannd ,

Can you merge the module? I've added the fixes.

alexey-pelykh (changes requested):
Thanks for the update. Re-reviewed against current HEAD — of the 12 issues from my Mar 1 review, 1 is fixed and 11 remain.

### Fixed

- `action_copy_last_fte_line` crash on empty FTE lines: fixed via `invisible` guard on the button.

### Still Open — Bugs

- **Direct assignment to computed stored field** `executed_hours` in `account_analytic_line.py` `unlink()` — `fte_line.executed_hours -= line.unit_amount` bypasses the compute method. The `create`/`write` paths correctly call `_compute_executed_hours()`, the `unlink` path should do the same.

- **Cron warning logic** still uses `continue` after the low-execution branch, which can skip the overload check in the same run. Reorder to check highest-severity (>100%) first, or use `elif`.

### Still Open — i18n

- `MSG_NO_EXEC`/`MSG_WITH_EXEC` still wrapped in `_()` at module level — remove the module-level `_()`, the call site already handles translation.
- Still no `.pot` template file.

### Still Open — Security

- `monthly_html_table` still has `sanitize=False` with unescaped f-string insertion of `dist.role_id.name`. Use `markupsafe.escape` or `sanitize=True`.

### Still Open — Code Quality

- `account_analytic_line.py`, `project_role.py`, `project_milestone.py` missing license headers.
- Redundant `profile_distribution_ids.unlink()` in `action_delete_fte` (cascade handles it).
- `(x * 100 / y) / 100` simplifies to `x / y`.
- `.sudo()` calls lack explanatory comments.

### Still Open — Typos

- "Montly" → "Monthly" in views.
- `inhetir` → `inherit` in XML record ID.

Issue #1538, opened 2025-07-15
Remove project_scrum from the repository.


adasatorres:
Due to the issue caused by adding the addon named project_scrum, I believe it should be permanently removed from this repository.

The issue in question lies in the fact that an addon with the same name already existed in the project-agile repository, so creating a new one was unnecessary. Since this addon is already available in that repository for versions 16.0 and 18.0, I consider it most appropriate to review the differences between it and the project_scrum addon from project-agile, and, if any useful elements are identified, incorporate those differences into the latter.

pedrobaeza:
Thanks for the followup on this. I hope you can converge soon.

leemannd:
I have questions about this in https://github.com/OCA/project/pull/1539#issuecomment-3137533197
If possible I want to know if the following is possible:
- do not remove the module from a repository
- set the module as uninstallble + warning in README

My main concern is about people loosing an installed module if they work with git submodules and if they update it.

Regarding the precedence of `project_scrum` in https://github.com/OCA/project-agile, you're correct. But as project-agile is hosting only `project_scrum` I would say that I'm more in favor of merging the module into OCA/project

mathben:
Because project_scrum was remove from OCA/project, and was originally into OCA/project-agile, I create a pull request, move and adapt it.
The last version of project_scrum was only into version 14.0 from OCA/project-agile

@leemannd

leemannd:
@mathben The removal have been merged. I merged them as people maintaining them where ok with it.
I still disagree on the main ideas:
- Removing a module from a submodule. It can break instances
- Keeping the submodule `project-agile`. It contains only this specific module. It is hard to find and I think in long term we may endup with the same issue. Having an equivalent of project_scrum into this submodule.

Aside that, you're doing a good job. I don't want to stop you and your contributions.

leemannd:
@adasatorres @mathben , the PRs to remove the module from oca/project are merged and there is a PR to port the functionalities to project-agile.
I think we can close this issue.
Feel free to re-open it in case of need.

PR #1539, opened 2025-07-15
[16.0][REM] project_scrum: This addon exist on project-agile repository.

adasatorres:
There is already an addon that provides this same functionality in the project-agile repository. This addon should not exist here, as it creates conflicts between the two. Due to its precedence, it is preferable to keep it in the other repository and implement any necessary improvements there.


adasatorres:
#1538

pedrobaeza (approved):

leemannd (dismissed):
Hello, shouldn't we keep the module in older versions and prevent it to be migrated? There are maybe people that are using it. Removing it from the repository will make it disappear when they update the submodule.

I'm more in favor of updating the README to ensure people don't migrate it.

pedrobaeza:
The problem is that there can't be in OCA 2 modules with the same name, or some automation will fail, like the apps store.

leemannd:
I agree. It can lead to failure.

I have taken time to:
  - check the original PR that adds project_scrum to this repo -> https://github.com/OCA/project/pull/1329
  - check the repository project-agile

What I am currently thinking is that the repository project-agile is hosting only one finished module that is `project_scrum`.
From what I understand, it was created to land project agile/scrum and JIRA like modules. They have never been merged.

We currently keep this other repository for only one module. Seeing that I must confess I am more in favor of bringing into this repository the features from the other one.

Aside that I don't know the automations, especially the one from the app store. Will it fails also if we set one of the two modules as uninstallable?

pedrobaeza:
I'm not able to answer that right now. @petrus-v is checking the modules of the apps store.

petrus-v:
I don't know that much the source code, I've just start to maintain some part of it, I think it's possible that the product description will blink between both modules:
* https://github.com/OCA/interface-github/tree/14.0/github_connector_odoo
* https://github.com/OCA/apps-store/tree/14.0/apps_product_creator

Personally I don't use the app-store and I'm wondering if some of us use it to install their odoo? The main impact it probably marketing. My main concerns is regarding module published on pypi.org. It makes no sense to have two modules with the same name maintains in different repositories. Even without pypi it will be very confusing to know which one is the good one you want to maintain as developer.

Personally moving a module by removing the repository or disable the module with a clear message that it has been move somewhere else is acceptable. If we want to maintain 2 different modules with different functional approach you can see 2 options:
* contact the former author and see if you agree to on a same way to work
* Rename your module to avoid naming conflict and could be nice to explain in module description what's the different approach with the former module

petrus-v (approved):

pedrobaeza:
@petrus-v a lot of people like me use the app-store to discover OCA modules by searching through keywords.

petrus-v:
I'll try to keep track, once this is merged and https://github.com/OCA/project-agile/ a PR is merged to make sure appstore is properly updated. Please ping me if you feel there are something wrong while moving those modules.

(Time to refresh all OCA modules is around 24h)

edescalona (approved):
LGTM

PR #1541, opened 2025-07-17
[16.0][FIX] project_timesheet_time_control: keep the date when it comes in the creation or writing values.

angelgarciadelachica:
There is an inconsistency in the date field and date_time because if a record is created with the date field the date field will have the assigned value, but the date_time will be the current one. For example when requesting Time off with the module Time Off.

To reproduce the error:

1. Create a time off request and approve it.
2. If you go to your timesheet you will see how the date_time that appears is not the date of the time off but the date on which the record has been created. On the other hand, the date field is the correct one. In this case, the date field has to override the date_time field.

image

This PR corrects this bug and a test is added to check it.

[T-8520]

### DIFF
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -3,7 +3,7 @@
 # Copyright 2016-2018 Tecnativa - Pedro M. Baeza
 # License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
 
-from datetime import datetime
+from datetime import datetime, time
 
 from dateutil.relativedelta import relativedelta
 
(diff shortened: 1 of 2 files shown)


sara-castello (approved):
Functional Review LGTM

project_timesheet_time_control/views/account_analytic_line_view.xml
```diff
@@ -5,10 +5,10 @@
         <field name="inherit_id" ref="hr_timesheet.hr_timesheet_line_tree" />
         <field name="arch" type="xml">
             <field name="date" position="attributes">
-                <attribute name="invisible">1</attribute>
+                <attribute name="optional">show</attribute>
```
pedrobaeza: Don't change this.

project_timesheet_time_control/views/account_analytic_line_view.xml
```diff
@@ -5,10 +5,10 @@
...
             </field>
             <field name="date" position="after">
-                <field name="date_time" required="1" />
+                <field name="date_time" required="1" optional="show" />
```
pedrobaeza: A required field can't be hidden.

pedrobaeza (changes requested):

project_timesheet_time_control/views/account_analytic_line_view.xml
angelgarciadelachica: Done!

pedrobaeza (approved):
/ocabot merge patch

PR #1549, opened 2025-07-24
[18.0][MIG] project_task_personal_stage_auto_fold

Andrii9090-tecnativa:
Straight worward migration. In Odoo 18 stage Canceled changed to Cancelled.

MT-10885 @moduon @rafaelbn @yajo @EmilioPascual @u0f


project_task_personal_stage_auto_fold/readme/CONFIGURE.md
```diff
@@ -0,0 +1,4 @@
+The task stages "Done" and "Cancelled" must match the same name as the
+equivalent personal stages.
+
+The checkbox “Closing stage” must be checked in the stage configuration.
```
EmilioPascual:
```suggestion
The checkbox “Folded in Kanban” must be checked in the stage configuration.
```

EmilioPascual (changes requested):

EmilioPascual:
Good job @Andrii9090, thank you. Just a small change.

yajo (changes requested):
Please see this video. I think we don't need this module anymore in v18:

https://github.com/user-attachments/assets/1b9bdfa8-415c-4f66-9448-bee4a0b9ba10

rafaelbn:
Review with @Andrii9090 . We should migrate it as the new features are for other purposes. Anyway we should learn new features.

rafaelbn (commented):
Hello! Thanks for this migration 😄 

After reviewing the migration I would recommend not comparing the name of personal stage with the name of project.task.stage because project task stages shouldn't be the same name. Like video edition and the final stage of the task in the project is Published. The personal stage is Done.

Thanks ❤️

Gelojr (approved):
Congratulations @Andrii9090  for the work done. The following tests were performed:

1. Test 1: Verified that changing a task to a non folded stage does not alter the personal stage and it worked as expected
2. Test 2: Verified that moving a task to a folded stage with the same exact name updated the personal stage correctly
3. Test 3: Verified that moving a task to Cancelled with the same exact name updated the personal stage correctly
4. Test 4: Verified that when no matching name exists the personal stage falls back to the first folded personal stage and it worked correctly
5. Test 5: Verified that in a multi-assignee task only the personal stage of the acting user was updated and not of the others
6. Test 6: Verified that if no personal folded stages exist the personal stage remains unchanged without errors
7. Test 7: Verified that only moves to folded stages trigger personal stage update and it worked correctly

EmilioPascual (approved):

rafaelbn (approved):
👍🏼 Tested in runboat

/ocabot merge patch

Andrii9090-tecnativa:
Hello @rafaelbn! Can you try merging again?
Thank you!

PR #1556, opened 2025-08-24
[18.0][ADD] project_task_default_user

NICO-SOLUTIONS:
This module automatically assigns default users to tasks:
- Stage-based (priority rule): When creating a task or changing its stage to one with default users, stage users are applied with a configurable assignment mode (replace or merge). Stage defaults always take priority over project defaults.
- Project-based fallback: If the stage does not define any default users, project default users are applied with the same assignment mode logic.
- Assignment behavior:
  - replace: replaces existing task users with default users
  - merge: adds default users to existing task users without removing them
- Multiple default users can be configured on both projects and stages.

P.S.: I was a bit surprised I couldn't find anything about this already, so I made this module. If something similar exists, please let me know.


NICO-SOLUTIONS:
@leemannd 
hey. could you please remove the stale tag? that would be great!

project_task_default_user/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+
+{
+    "name": "Project Task Default User",
+    "summary": "Auto assign default users to tasks or when changing task stages",
+    "version": "18.0.1.0.0",
+    "category": "project",
+    "author": "NICO SOLUTIONS - ENGINEERING & IT, Odoo Community Association (OCA)",
+    "maintainers": ["NICO-SOLUTIONS"],
+    "website": "https://github.com/OCA/project",
+    "depends": ["project"],
```
alexey-pelykh, line 12: Nit: Odoo's standard category name is `"Project"` (capital P). Using lowercase may cause it to appear as a separate category in the Apps menu.

project_task_default_user/models/project_task.py
```diff
@@ -0,0 +1,42 @@
+# Copyright 2025 NICO SOLUTIONS - ENGINEERING & IT
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import models
+
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    def create(self, vals_list):
```
alexey-pelykh, line 11: Minor: it would be a bit more efficient to inject `user_ids` into the vals dicts before calling `super().create()` rather than writing after creation. That way you avoid an extra `write()` call per task. Not blocking, just a performance thought.

alexey-pelykh (approved):
Thanks for this contribution, nice and clean module. The logic is straightforward and tests cover the main scenarios well. Just left a couple of minor suggestions below.

JordiMForgeFlow (approved):

project_task_default_user/models/project_task.py
```diff
@@ -0,0 +1,35 @@
...
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        tasks = super().create(vals_list)
+
+        for task in tasks:
+            default_users = (
+                task.stage_id.mapped("default_user_ids.id") if task.stage_id else []
```
AaronHForgeFlow: task.stage_id.default_user_ids.ids

project_task_default_user/models/project_task.py
```diff
@@ -0,0 +1,35 @@
...
+            )
+            if not default_users and task.project_id:
+                default_users = task.project_id.mapped("default_user_ids.id")
+
+            if default_users:
+                task.user_ids = [(6, 0, default_users)]
```
AaronHForgeFlow: task.user_ids = [Command.set(default_users)] ?

project_task_default_user/models/project_task.py
```diff
@@ -0,0 +1,35 @@
...
+        tasks = super().create(vals_list)
+
+        for task in tasks:
+            default_users = (
+                task.stage_id.mapped("default_user_ids.id") if task.stage_id else []
+            )
+            if not default_users and task.project_id:
+                default_users = task.project_id.mapped("default_user_ids.id")
+
+            if default_users:
```
AaronHForgeFlow, line 15: here this is overwriting what the users put, isn't it?

if default_users and not task.user_ids:

project_task_default_user/models/project_task.py
```diff
@@ -0,0 +1,35 @@
...
+from odoo import api, models
+
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        tasks = super().create(vals_list)
+
```
AaronHForgeFlow: 🙏  nitpicking empty lines inside methods can be removed

AaronHForgeFlow (commented):
LGTM, just some code suggestions

project_task_default_user/models/project_task.py
NICO-SOLUTIONS: @AaronHForgeFlow 
Hey, thanks for your review!
Current behavior intentionally overwrites user_ids with stage/project defaults. This ensures stage-driven assignment is authoritative.

An alternative would be to only apply defaults when user_ids is empty (if default_users and not task.user_ids), making defaults initial-only and preventing updates on stage changes.
The original intention was to reflect stage responsibility and ensure users are assigned accordingly. We could also consider a merge approach instead of full overwrite to preserve manually assigned users.

What do you think.. Do we want “authoritative overwrite” or “initial suggestion / merge behavior”?

AaronHForgeFlow: Ok, if the intention was the "authoritative overwrite" that is fine to me, just please make it clear in the readme.

NICO-SOLUTIONS:
@AaronHForgeFlow 
I added project_task_assignment_mode and stage_task_assignment_mode to control whether default users are replaced or merged with existing task users. 
Also refactored the method structure, adapted the test suite, and updated the README accordingly. 

Could you please add another review when you have time?

AaronHForgeFlow (approved):
Thanks!  Code review LGTM!

NICO-SOLUTIONS:
@leemannd 
thx

PR #1559, opened 2025-09-02
[18.0][MIG] project_hr: Migration to 18.0

victoralmau:
Migration to 18.0

Please @pilarvargas-tecnativa and @juancarlosonate-tecnativa can you review it?

@Tecnativa TT57843


project_hr/models/project_task.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+    # This field could have been named allowed_user_ids but a field with
+    # that name already exists in the Odoo core 'project' module
+    allowed_assigned_user_ids = fields.Many2many(
+        comodel_name="res.users",
+        string="Allowed users",
+        compute="_compute_allowed_assigned_user_ids",
+        help="Technical field for computing allowed users according employee "
+        "category.",
+    )
```
carlos-lopez-tecnativa: Could you convert these many2many fields to Binary if they are only used in a domain, similar to:

https://github.com/odoo/odoo/blob/c0a7b51c9e14d29cefa96c29dd716b7aec698818/addons/account/models/account_tax.py#L3026
https://github.com/odoo/odoo/blob/c0a7b51c9e14d29cefa96c29dd716b7aec698818/addons/account/views/account_tax_views.xml#L49

carlos-lopez-tecnativa (changes requested):

project_hr/models/project_task.py
```diff
@@ -0,0 +1,93 @@
...
+    domain_hr_category_ids = fields.Binary(compute="_compute_domain_hr_category_ids")
+    # This field could have been named allowed_user_ids but a field with
+    # that name already exists in the Odoo core 'project' module
+    allowed_assigned_user_ids = fields.Many2many(
+        comodel_name="res.users",
+        string="Allowed users",
+        compute="_compute_allowed_assigned_user_ids",
+        help="Technical field for computing allowed users according employee "
+        "category.",
+    )
```
carlos-lopez-tecnativa: Just out of curiosity, is there any reason not to apply the same change made to the field `domain_hr_category_ids` to this field?

victoralmau: You're right, it can be changed too, done!

carlos-lopez-tecnativa (changes requested):
I got this error in the kanban view when I tried to assign a user.
image

juancarlosonate-tecnativa (approved):
LGTM!

pedrobaeza (approved):
/ocabot merge nobump

PR #1407, opened 2024-12-20
[18.0][MIG] project_task_description_template: Migration to 18.0

BhaveshHeliconia:

PR #1560, opened 2025-09-03
[18.0][ADD] sale_project_task_description_template

NICO-SOLUTIONS:
Depends on:

- [x] <https://github.com/OCA/project/pull/1407>


sale_project_task_description_template/models/project_task.py
```diff
@@ -0,0 +1,34 @@
...
+                template = self.env["project.task.description.template"].browse(
+                    template_id
+                )
+                if template.exists():
+                    description_parts.append(template.description)
+
+            if description_parts:
+                vals["description"] = "\n".join(description_parts)
+
+        return super().create(vals_list)
```
alexey-pelykh: This entire `create` override seems redundant with the logic in `sale_order_line._timesheet_create_task_prepare_values`, which already assembles the description (including sale line info) before the task is created. The context keys `sale_line_product` / `sale_line_qty` are never set by any code in this module.

If the intent is to also support direct task creation with a template (outside the sale flow), that is the responsibility of the base `project_task_description_template` module. Could you clarify the use case or remove this override to avoid the dual code path?

sale_project_task_description_template/models/sale_order_line.py
```diff
@@ -0,0 +1,26 @@
...
+        values = super()._timesheet_create_task_prepare_values(project)
+
+        product = self.product_id.product_tmpl_id
+        template = product.task_description_template_id
+
+        if template:
+            values["description_template_id"] = template.id
+
+        if product.include_sale_line_info_in_task:
+            product_info = f"{self.product_id.name} - Qty: {self.product_uom_qty}"
```
alexey-pelykh: The `"Qty:"` label here is not translatable. Consider using `_("Qty:")` (with the import `from odoo import _`) so translators can localize it.

sale_project_task_description_template/tests/test_sale_order_line.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+        self.assertIn("description", values)
+        self.assertEqual(values["description"], "")
+        self.assertIn("description_template_id", values)
+
+    def test_task_values_with_empty_template_and_sale_info(self):
+        values = self._prepare_task_values(
+            include_sale_line_info=True, template=self.empty_template, qty=4
+        )
+        expected_prefix = f"{self.product.name} - Qty: 4.0"
```
alexey-pelykh, line 89: Note the `4.0` here vs `5` (int) in `test_project_task.py` -- this reflects the float/int inconsistency between the two description assembly paths (one uses `product_uom_qty` which is a float, the other uses the raw context value). This is another symptom of the duplicated logic between the two models.

alexey-pelykh (commented):
Thanks for the contribution, nice bridge module. Two things I'd like to flag: (1) the `create` override in `project_task.py` duplicates the description assembly logic already handled in `sale_order_line._timesheet_create_task_prepare_values` -- the context keys `sale_line_product` / `sale_line_qty` are never set by this module, making the create override effectively dead code for the sale-to-task flow; could you clarify the intent or remove the duplication? (2) the `"Qty:"` label in the f-strings is not wrapped in `_()`, so it won't be translatable.

sale_project_task_description_template/models/project_task.py
NICO-SOLUTIONS: You're right — I'll remove the override to avoid duplication and keep the logic in the proper preparation method.

sale_project_task_description_template/models/sale_order_line.py
```diff
@@ -0,0 +1,29 @@
+from odoo import _, models
```
CristianoMafraJunior, line 4: Add the license

sale_project_task_description_template/models/product_template.py
```diff
@@ -0,0 +1,17 @@
+from odoo import fields, models
```
CristianoMafraJunior, line 4: Add the license

sale_project_task_description_template/tests/test_product_template.py
```diff
@@ -0,0 +1,34 @@
+from odoo.tests.common import TransactionCase
```
CristianoMafraJunior, line 4: Add the license

sale_project_task_description_template/tests/test_sale_order_line.py
```diff
@@ -0,0 +1,89 @@
+from odoo.tests.common import TransactionCase
+
+
```
CristianoMafraJunior, line 6: Add the license

CristianoMafraJunior (changes requested):

alexey-pelykh (approved):
Thanks for the follow-up, @NICO-SOLUTIONS — both points from my earlier review are resolved:

- The redundant `create` override in `project_task.py` is gone; the description is now assembled solely in `sale_order_line._timesheet_create_task_prepare_values`, so there's a single code path and no more dead `sale_line_product` / `sale_line_qty` context keys.
- The label is now translatable — `_("%(product)s - Qty: %(qty)s")` with named placeholders, and the matching entry is present in the `.pot`.

License headers are in place on the model and test files, and CI is green. LGTM.

BhaveshHeliconia (approved):
LGTM!

CristianoMafraJunior (approved):
LGTM

NICO-SOLUTIONS:
@pedrobaeza 

Hey Pedro. 

Can we merge this one? 
Looking forward to migrate this to 19.

NICO-SOLUTIONS:
@pedrobaeza 
Hey Pedro, thanks for the merge attempt.
It seems to be a temporary PyPI rate limit issue. I'm not sure when it will be resolved.

Could you please retry the merge whenever it suits your schedule?

pedrobaeza:
The problem is deeper. Please check https://www.odoo-community.org/groups/contributors-69/limit-to-what-we-can-submit-to-pypi-15013?mode=thread&date_begin=&date_end=

NICO-SOLUTIONS:
> The problem is deeper. Please check https://www.odoo-community.org/groups/contributors-69/limit-to-what-we-can-submit-to-pypi-15013?mode=thread&date_begin=&date_end=

@pedrobaeza 

Thanks Pedro, understood. I thought it might just be a temporary PyPI rate limit issue, but if the problem is deeper I’ll leave the PR as is and wait for the OCA/PyPI issue to be resolved.

Let me know if there is anything needed from my side.

pedrobaeza:
We can try today:

/ocabot merge nobump

NICO-SOLUTIONS:
@pedrobaeza 

Great, it went through! 
Thanks a lot, Pedro, for taking care of it and for your help with the merge. Really appreciated!

¡Muchas gracias! 😉

PR #1561, opened 2025-09-04
[16.0][FIX] project_key: handle project search safely by id, key, or name

BhaveshHeliconia:
* issue :- 
  The module extended project.project with _rec_names_search = ["key", "name", "id"].
  Including "id" in _rec_names_search caused SQL errors when users searched with text values (e.g., "inter").

  Odoo generated invalid queries like:
  "project_project"."id" = 'inter'
  which failed because id is an integer field.

* Steps to Reproduce the Error :-
1. Go to Timesheets → Tree/List View.
2. Add a custom filter:
   Field: Project
   Operator: is equal to
   Value: any string (e.g., "inter")
   Apply the filter.

You will get the following error:
ERROR: invalid input syntax for type integer: "inter"
LINE 1: ...->>'en_US' = 'inter')) OR ("project_project"."id" = 'inter')) AN...

https://github.com/user-attachments/assets/6cf52449-7de8-44eb-9447-1d6490f4028e


BhaveshHeliconia:
Blocked by : https://github.com/OCA/project/pull/1562 Once that PR is merged, the CI will be green.

BhaveshHeliconia:
@AaronHForgeFlow  @yajo , the PR is ready for review at your convenience.

AaronHForgeFlow (approved):
:+1:  Code LGTM

project_key/models/project_project.py
```diff
@@ -7,7 +7,31 @@
...
+                search_domain.append(("id", "=", int(name)))
+
+            if len(search_domain) > 1:
+                domain = ["|"] * (len(search_domain) - 1) + search_domain
+            else:
+                domain = search_domain
+        else:
+            domain = []
+
+        return self._search(domain + args, limit=limit, access_rights_uid=name_get_uid)
```
yajo: issue: IIUC, you don't need to redo everything that `_name_search` will do automatically when `_rec_names_search` is properly configured.

Besides, overriding without calling `super()` is a general bad practice.

Check this please:

```suggestion
        args = list(args or [])
        if name.isdigit():
            args.insert(0, ("id", "=", int(name)))
        return super()._name_search(name, args, operator, limit, name_get_uid)
```

yajo (changes requested):

yajo:
Oh, and please add a test. Thanks!

BhaveshHeliconia:
@yajo Sorry for the delayed fix—I missed this PR earlier. I’ve now made the changes as per your suggestion. Thank you!

project_key/models/project_project.py
```diff
@@ -7,7 +7,26 @@
 
 class Project(models.Model):
     _inherit = "project.project"
-    _rec_names_search = ["key", "name", "id"]
+    _rec_names_search = ["key", "name"]
+
+    @api.model
+    def _name_search(
+        self, name="", args=None, operator="ilike", limit=100, name_get_uid=None
```
alexey-pelykh, line 15: This `strip()` is not necessary -- `super()._name_search()` handles whitespace internally. Removing it keeps the override focused on only the ID-search concern.

project_key/models/project_project.py
```diff
@@ -7,7 +7,26 @@
...
+        res = super()._name_search(name, args, operator, limit, name_get_uid)
+
+        if name and name.isdigit():
+            args = list(args or [])
+            args.insert(0, ("id", "=", int(name)))
+            res_id = super()._name_search("", args, "ilike", limit, name_get_uid)
+            res = list(res)
+            res_ids = set(res)
+            res_id = [x for x in res_id if x not in res_ids]
+            res = res_id + res
```
alexey-pelykh: Nit: the combined `res_id + res` list may exceed `limit`. Consider `return (res_id + res)[:limit]` if the caller relies on the limit being respected.

alexey-pelykh (commented):
Thanks for the fix. The root cause is correctly identified: including `"id"` in `_rec_names_search` causes SQL errors when users search with non-numeric strings.

The approach -- removing `"id"` from `_rec_names_search` and handling integer searches manually in `_name_search` -- is the right direction. The two-call approach is more correct than a simple domain append (which would AND the ID condition with name/key, requiring both to match simultaneously).

A few observations:

- `name.strip()` is unnecessary -- `super()._name_search()` handles this internally.
- The combined `res_id + res` list may exceed `limit`. Consider `return (res_id + res)[:limit]`.
- The try/except pattern in the test works but is not idiomatic for Odoo tests -- simply calling `name_search` without wrapping already validates no exception is raised, and if it does raise, the test framework reports it with a full traceback.

Overall the fix addresses the root cause and includes a regression test. The remaining items are style/polish.

BhaveshHeliconia:
@alexey-pelykh Thanks for the detailed review and feedback.
I've applied the suggested improvements: removed the unnecessary strip(), limited the combined results using [:limit], and simplified the test to follow the usual Odoo testing style without the try/except. Please have another look when you get a chance.

project_key/models/project_project.py
```diff
@@ -7,7 +7,24 @@
...
+        res = super()._name_search(name, args, operator, limit, name_get_uid)
+
+        if name and name.isdigit():
+            args = list(args or [])
+            args.insert(0, ("id", "=", int(name)))
+            res_id = super()._name_search("", args, "ilike", limit, name_get_uid)
+            res = list(res)
+            res_ids = set(res)
+            res_id = [x for x in res_id if x not in res_ids]
+            res = (res_id + res)[:limit]
```
yajo: issue: This will break in negative searches.

project_key/tests/test_project.py
```diff
@@ -81,3 +81,17 @@ def test_12_onchange_key(self):
...
+
+        res = self.env["project.project"].name_search("Alpha")
+        ids = [x[0] for x in res]
+        self.assertIn(pid, ids, "Should find project by name 'Alpha'")
+
+        res = self.env["project.project"].name_search(str(pid))
+        ids = [x[0] for x in res]
+        self.assertIn(pid, ids, "Should find project by ID '{}'".format(pid))
+
+        self.env["project.project"].name_search("inter")
```
yajo: question: what is this line doing?

yajo (changes requested):

project_key/tests/test_project.py
BhaveshHeliconia: This is a regression test for the original bug this PR fixes. The root cause was that including "id" in _rec_names_search caused SQL errors when users searched with non-numeric strings (e.g. "inter"). This line verifies that a plain text search no longer raises an exception. I've added a comment explaining this and a basic assertion for clarity.

project_key/models/project_project.py
```diff
@@ -2,12 +2,34 @@
...
+    @api.model
+    def _name_search(
+        self, name="", args=None, operator="ilike", limit=100, name_get_uid=None
+    ):
+        res = super()._name_search(name, args, operator, limit, name_get_uid)
+
+        if (
+            name
+            and name.isdigit()
+            and operator not in expression.NEGATIVE_TERM_OPERATORS
```
yajo, line 22: issue: if you're deliberately not supporting negative operators, then you should at least document it in known issues.

yajo (changes requested):

yajo (approved):
/ocabot merge patch

PR #1566, opened 2025-09-13
[ADD] project_portal_properties

etobella:
Makes use of https://github.com/OCA/web/pull/3284


project_portal_properties/views/project_task_portal_template.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <template id="portal_my_task" inherit_id="project.portal_my_task">
+        <xpath expr="//div[@id='card_body']" position="inside">
+
+            <t t-call="web_portal_properties.portal_properties">
+                <t t-set="properties_record" t-value="task" />
+                <t t-set="properties_field" t-value="'task_properties'" />
```
alexey-pelykh, line 8: This passes `task_properties` (unfiltered) to the portal template instead of `portal_task_properties` (filtered by `view_in_portal`). Shouldn't this be `'portal_task_properties'` to respect the filtering logic defined in `project_project.py`? As-is, all properties are rendered in the portal view regardless of the `view_in_portal` flag.

project_portal_properties/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "version": "17.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Dixmit,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/project",
+    "depends": ["project", "web_portal_properties"],
+    "data": [
+        "views/project_task.xml",
+        "views/project_task_portal_template.xml",
+    ],
+    "demo": [],
```
alexey-pelykh, line 16: No tests and no demo data. Given this module exposes data to portal users (security-sensitive), tests verifying that only `view_in_portal=True` properties are visible would be valuable.

alexey-pelykh (changes requested):
Thanks for the contribution, Enric -- exposing project properties in the portal is a nice addition. A few things need attention before this can move forward: (1) The dependency `web_portal_properties` (OCA/web#3284) is still unmerged, which is why CI fails. This PR should be marked as blocked until that lands. (2) The portal template (`project_task_portal_template.xml`) passes `task_properties` to the `portal_properties` widget instead of `portal_task_properties` -- this bypasses the `view_in_portal` filtering and could expose properties that the project manager didn't intend to show to portal users. (3) The module needs tests, especially given the security surface (portal access to properties).

PR #1568, opened 2025-09-17
[18.0][MIG] project_budget

mathben:
Migrate from odoo 12
[project_budget](https://github.com/OCA/project/tree/12.0/project_budget)


mathben:
Build failed, because depend on https://github.com/OCA/account-budgeting/pull/79

mathben:
I format code on last force-pushed commit

mathben:
@rousseldenis
This pull request is fixed and all check pass after a rebase.
Thanks for your review!

rvalyi (changes requested):
something looks strange in the setup directory. This setup directory is similar to what we had until Odoo 16.0, but since 17.0 it's not supposed to be like that. 

in version 18.0, There is still a setup directory, but it's only a global setup for the repo, see for instance: https://github.com/OCA/dms/tree/18.0/setup/_metapackage  and each module don't touch that setup directory but only bring a pyproject.toml, see for instance: https://github.com/OCA/dms/blob/18.0/dms/pyproject.toml

Are you sure you use the version 18.0 pre-commit setup for this pull request? You should fix this.

mathben:
Thanks for your review @rvalyi 
I did the change manually, but do we have automated command to do it?

mathben:
Is it fixed @rvalyi , after a rebase? Thanks

alexey-pelykh (approved):
Migration looks clean. The 18.0 structure conventions are followed correctly (pyproject.toml with whool, list instead of tree). ORM fields, security groups, and views are straightforward and match the upstream account_budget_oca 18.0 API. All CI checks pass. The only suggestion would be adding basic tests, but this is not a regression from 12.0 which also lacked tests.

rvalyi (approved):
sorry for the delay. Yes it looks clean, thank you for the migration!

PR #1569, opened 2025-09-22
[17.0][MIG] project consumable

petrus-v:
Migrate project consumable https://github.com/OCA/project/pull/1410 from v14.0

with main differences:
* refactor do not depends on hr_timesheet and avoid conflict with it, the shame is that odoo is not using polymorphism to generate account.move.line 😭 so odoo assume that an account.analytic.line with a project_id are timesheet line making very hard to add a new kind of account.analytic.line linked to the project
* in project.update profitability right panel add consumable/materials somehow => smart button + other cost is fine to me for the time being.


project_consumable/models/product_template.py
```diff
@@ -0,0 +1,25 @@
...
+    project_ok = fields.Boolean(
+        string="Available in projects",
+        help="Check this box to be able to link this product with "
+        "analytic line and project or task. Product cost will be used "
+        "and displayed in project dashboard as consumable cost. "
+        "So you'll be able to analyse consumable products cost per project.",
+    )
+
+    @api.onchange("type")
+    def _onchange_type(self):
```
rousseldenis: Make `project_ok` fields compute/stored/readonly=False to avoid onchanges.

project_consumable/models/project_project.py
```diff
@@ -0,0 +1,90 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, _lt, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class Project(models.Model):
+    _inherit = "project.project"
+
+    company_currency_id = fields.Many2one(
```
rousseldenis: Is this at the right place (this module ?) ?

petrus-v: I admit I can't remember why I'm not using the existing currency_id field in the monitory field... I've moved  to the existing currency_id field

Thanks !

project_consumable/views/analytic_account_line.xml
```diff
@@ -0,0 +1,431 @@
...
+        <field name="view_mode">kanban</field>
+        <field name="sequence">6</field>
+        <field name="view_id" ref="project_consumable_product_line_kanban" />
+        <field name="act_window_id" ref="act_project_consumable_product_line" />
+    </record>
+
+    <menuitem
+        id="menu_hr_product_tracking"
+        name="My Materials"
+        parent="hr_timesheet.menu_hr_time_tracking"
```
petrus-v:
```suggestion
        parent="hr_timesheet.timesheet_menu_root"
```

project_consumable/models/project_project.py
```diff
@@ -0,0 +1,85 @@
...
+        buttons = super()._get_stat_buttons()
+        if not self.allow_consumables or not self.env.user.has_group(
+            "project.group_project_manager"
+        ):
+            return buttons
+
+        buttons.append(
+            {
+                "icon": "copy",
+                "text": _lt("Materials"),
```
alexey-pelykh, line 73: The `€` symbol is hardcoded here. For companies using a different currency this will be misleading. You could use `self.currency_id.symbol` or format via the currency's display method.

project_consumable/views/project_views.xml
```diff
@@ -0,0 +1,60 @@
...
+<odoo>
+
+    <record id="act_consumable_line_by_project" model="ir.actions.act_window">
+        <field name="name">Consumables</field>
+        <field name="res_model">account.analytic.line</field>
+        <field name="view_mode">tree,form</field>
+        <field name="view_id" ref="project_consumable_product_line_tree_user" />
+        <field
+            name="domain"
+        >[('consumable_project_id', '!=', False), ('product_id', '!=', False)]</field>
```
alexey-pelykh, line 14: Should this be `invisible="not allow_consumables"` instead of `not allow_timesheets`? Currently the consumable stat button is tied to the timesheet toggle rather than the consumable toggle.

project_consumable/models/account_analytic_line.py
```diff
@@ -0,0 +1,135 @@
...
+                line._consumable_postprocess(values)
+        return lines
+
+    def write(self, values):
+        values = self._consumable_preprocess([values])[0]
+        result = super().write(values)
+        # applied only for timesheet
+        self.filtered(lambda t: t.consumable_project_id)._consumable_postprocess(values)
+        return result
+
```
alexey-pelykh, line 52: This comment says "applied only for timesheet" but the filter is on `consumable_project_id`. Looks like a leftover from the v14 version.

project_consumable/hooks.py
```diff
@@ -0,0 +1,2 @@
+def set_project_ok_for_consumable_products(env):
```
alexey-pelykh, line 6: Missing license header (all other `.py` files in this module have the AGPL-3 header).

project_consumable/tests/test_project_consumable.py
```diff
@@ -0,0 +1,243 @@
...
+            groups="project.group_project_manager,hr_timesheet.group_timesheet_manager",
+        )
+
+    def test_onchange_product_type_project_ok_to_be_true(self):
+        self.product.project_ok = False
+        self.product.type = "consu"
+        self.product.product_tmpl_id._onchange_type()
+        self.assertTrue(self.product.project_ok)
+
+    def test_onchange_product_type_project_ok_to_be_false(self):
```
alexey-pelykh, line 49: This calls `_onchange_type()` but the module defines `_compute_project_ok` as a stored computed field with `@api.depends("type")`. In 17.0, changing `type` should trigger the compute automatically. The test works because `_onchange_type` exists on the parent model, but it isn't testing the compute defined in this module.

alexey-pelykh (commented):
Thanks for migrating this to 17.0, Pierre -- useful module for tracking project materials cost.

I have a few observations after going through the code:

### Hardcoded currency symbol

In `project_project.py` the stat button text uses a hardcoded `€` sign:

```python
"number": _lt(
    "%(amount)s € (%(count)s)",
    amount=self.consumable_total_price,
    count=self.consumable_count,
),
```

This will display incorrectly for non-EUR companies. Consider using the project's currency symbol or the company currency instead.

### Stat button visibility condition

In `views/project_views.xml`, the consumable stat button uses `invisible="not allow_timesheets"` but it seems like this should be `invisible="not allow_consumables"`. As it stands, the button only shows when timesheets are enabled on the project, regardless of whether consumables are enabled. Is that intentional (i.e., menus are under timesheet so they go together) or an oversight?

### Test `_onchange_type` calls

The tests `test_onchange_product_type_project_ok_to_be_true` and `test_onchange_product_type_project_ok_to_be_false` call `self.product.product_tmpl_id._onchange_type()`. However, the `product_template.py` model defines `_compute_project_ok` as a stored computed field with `@api.depends("type")`, not an onchange. In 17.0, setting `self.product.type = "consu"` should trigger the compute automatically. If `_onchange_type` is inherited from a parent module, the test might still work, but the tests are not really exercising your compute method -- they're relying on a side effect of an onchange defined elsewhere.

### `write` method comment

In `account_analytic_line.py` line 53, the comment says:
```python
# applied only for timesheet
```
but this actually filters for consumable lines (`t.consumable_project_id`). Looks like a leftover from the v14 code.

### Missing license header in `hooks.py`

The `hooks.py` file is missing the standard AGPL-3 license header that all other Python files in the module have.

### `fr.po` version header

The `fr.po` file header says `Project-Id-Version: Odoo Server 14.0` -- should be updated to 17.0 to match the migration target.

### PR description vs actual dependencies

Minor note: the PR description mentions "refactor do not depends on hr_timesheet" but `hr_timesheet` is still listed in `depends` and the security rules / menus reference `hr_timesheet` groups. The description could be clarified to reflect the actual state.

project_consumable/views/project_views.xml
petrus-v: coment on the wrong line but that's right

project_consumable/models/account_analytic_line.py
petrus-v: removing useless comment !

alexey-pelykh (approved):
Thanks for the updates, @petrus-v — the observations from my earlier pass are addressed:

- The stat-button number no longer hardcodes `€`; it uses `%(currency_symbol)s` fed by `self.currency_id.symbol`.
- The consumable stat button is gated on `invisible="not allow_consumables"` instead of `not allow_timesheets`.
- `project_ok` is a stored compute (`@api.depends("type")`), and the tests now exercise it by setting `product.type` directly rather than calling `_onchange_type()`.
- The misleading "applied only for timesheet" comment in `write` is removed, and `hooks.py` carries the AGPL-3 header.

One trivial, non-blocking leftover: `i18n/fr.po` still has `Project-Id-Version: Odoo Server 14.0` while the `.pot` is now 17.0 — Weblate territory, not a merge blocker. LGTM.

PR #1257, opened 2024-03-29
[15.0][MIG] project_key

vincent-hatakeyama:
Standard migration of the project_key module.

See #1055 for comments and history

PR #1570, opened 2025-09-29
[15.0][MIG] project_key: Migration to 15.0

jans23:
This is based on #1257 including another fix.


rousseldenis:
@jans23 Could you squash your commits ?

jans23:
Done. Or do you want me to squash it with even older commits?

alexey-pelykh (commented):
## Review: Migration of `project_key` to 15.0

Thanks for picking up this migration! The module is largely functional and CI is green, which is a good sign. I have a few findings that should be addressed before merging.

### PR Title

The PR title `15.0 mig project key and history` doesn't follow OCA convention. Recommended: `[15.0][MIG] project_key: Migration to 15.0`.

---

### Findings

#### 1. `invalidate_cache` removed in Odoo 15.0 (blocking)

In `project_key/models/project_project.py`, the method `_update_task_keys` calls:
```python
self.env["project.task"].invalidate_cache(["key"], self.task_ids.ids)
```

`invalidate_cache` was deprecated in Odoo 14.0 and **removed in Odoo 16.0**, but the preferred replacement in 15.0 is already `invalidate_model` or `invalidate_recordset`. While this may still work on 15.0 (it was kept for backward compatibility), it should be updated to the 15.0 pattern:
```python
self.env["project.task"].invalidate_model(["key"])
```
or, if you want to invalidate only specific records:
```python
self.task_ids.invalidate_recordset(["key"])
```

This will also make the forward-port to 16.0+ straightforward.

#### 2. `flush()` without arguments deprecated in Odoo 15.0 (blocking)

In `project_key/models/project_project.py`, the method `_update_task_keys` calls:
```python
self.flush()
```

In Odoo 15.0, `flush()` without field arguments is deprecated. Replace with:
```python
self.flush_recordset()
```
or more specifically:
```python
self.env["project.task"].flush_model(["key", "project_id"])
```

#### 3. `create()` should support `vals_list` (moderate)

In Odoo 15.0, `create()` can receive a list of dictionaries. The overrides in both `project_project.py` and `project_task.py` only handle a single `vals` dict. While this works (Odoo wraps single dicts), the OCA standard for 15.0+ is to handle `vals_list`:

```python
@api.model_create_multi
def create(self, vals_list):
    for vals in vals_list:
        if "key" not in vals:
            vals["key"] = self.generate_project_key(vals["name"])
    ...
```

This is especially relevant for `project.task` where batch creation is common.

#### 4. `name_get` and `name_search` deprecation (minor, informational)

Odoo 15.0 introduces `_compute_display_name` as the replacement for `name_get`, and `_name_search` as the replacement for `name_search`. The task model already migrated `name_search` to `_name_search` (good!), but the project model still uses the old-style `name_search`, and both models still use `name_get`. These still work in 15.0 but should ideally be migrated for forward-compatibility.

Note: The task's `_name_search` override is well-done for the 15.0 API.

#### 5. Controller uses `http.request.env()` with parentheses (minor)

In `project_key/controllers/main.py`:
```python
env = http.request.env()
```

In Odoo 15.0, `request.env` is a property (not a callable). Calling it with `()` still works but returns a new Environment each time. The idiomatic way is:
```python
env = http.request.env
```

#### 6. Missing `active_test=False` context in `_generate_project_unique_key` (carried from 14.0)

The 15.0 version correctly added `active_test=False` when searching for existing keys in `_set_default_project_key`, but `_generate_project_unique_key` still searches without it:
```python
unique_key = not bool(
    self.with_context(active_test=False).search([("key", "=", res)])
)
```
This was added in the PR (good improvement over 14.0). Just noting it is present and correct.

#### 7. Company propagation via context (good addition)

The `create` method now propagates `company_id` via context for sequence creation. This is a good improvement that addresses a real multi-company security concern.

#### 8. Deferred task creation in `project.create` (good addition)

The pattern of popping `task_ids` from `vals`, creating the project first, then writing tasks back is a sound approach to avoid issues with task key generation before the project sequence exists.

---

### Summary

| # | Finding | Severity |
|---|---------|----------|
| 1 | `invalidate_cache` should be `invalidate_model`/`invalidate_recordset` | Blocking |
| 2 | `flush()` should be `flush_recordset()` or `flush_model()` | Blocking |
| 3 | `create()` should use `@api.model_create_multi` | Moderate |
| 4 | `name_get`/`name_search` could be migrated to 15.0 patterns | Minor |
| 5 | `http.request.env()` should be `http.request.env` | Minor |

Items 1 and 2 should be fixed before merge as they use APIs that are deprecated/removed in subsequent versions and represent incomplete migration. Items 3-5 are improvements that would bring the module closer to OCA 15.0 standards.

Overall the migration is solid -- good test coverage, proper CI passage, and valuable improvements over the 14.0 version (company propagation, deferred task creation, `_name_search` migration).

jans23:
We moved on to a newer Odoo version and don't have capacity to bring this PR further.

PR #1572, opened 2025-09-29
[ADD] project_share

etobella:
This module adds a new functionality on collaborators, so we can have readonly collaborators that are able to see kanban and list views in the same way we have in internal view.

image

image


project_share/hooks.py
```diff
@@ -0,0 +1,19 @@
...
+            ('active', '=', True),
+            '|',
+                ('project_id.message_partner_ids',
+                 'child_of',
+                 [user.partner_id.commercial_partner_id.id]),
+                ('message_partner_ids',
+                 'child_of',
+                 [user.partner_id.commercial_partner_id.id]),
+            ('project_id.edit_collaborator_ids.partner_id', 'in', [user.partner_id.id]),
+        ]
```
alexey-pelykh, line 16: This domain line restricts task visibility to partners present in `edit_collaborator_ids` (i.e. non-readonly collaborators). Readonly collaborators would therefore not match this rule and would not see any tasks. Shouldn't the condition use `collaborator_ids` (all collaborators) instead of `edit_collaborator_ids` here, so that readonly collaborators can at least read tasks?

project_share/models/project_project.py
```diff
@@ -0,0 +1,26 @@
...
+        string="Collaborators",
+        copy=False,
+        domain=[("readonly", "=", False)],
+    )
+
+    def _check_project_sharing_access(self, check_readonly=False):
+        result = super()._check_project_sharing_access()
+        if (
+            check_readonly
+            and isinstance(result, models.BaseModel)
```
alexey-pelykh, line 22: Checking `isinstance(result, models.BaseModel) and result._name == 'project.collaborator'` is fragile -- the parent method's return contract could change. Consider handling unexpected return types more defensively.

project_share/models/project_task.py
```diff
@@ -0,0 +1,51 @@
...
+
+from odoo import api, models
+
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    @api.model
+    def get_view(self, view_id=None, view_type="form", **options):
+        result = super().get_view(view_id=view_id, view_type=view_type, **options)
```
alexey-pelykh, line 14: The three blocks for kanban/form/tree are nearly identical -- consider extracting a helper like `_set_readonly_view(doc, tag)` to reduce duplication.

project_share/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "license": "AGPL-3",
+    "author": "Dixmit,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/project",
+    "depends": [
+        "project",
+    ],
+    "data": [
+        "views/project_collaborator.xml",
+    ],
+    "demo": [],
```
alexey-pelykh, line 20: No tests directory in the module. Given that this module rewrites a security record rule and dynamically modifies views based on collaborator access level, test coverage is important. At minimum: a test for readonly collaborator seeing tasks (read) but not editing, and a test for edit collaborator retaining full access.

alexey-pelykh (changes requested):
Thanks for this contribution, Enric -- the idea of readonly collaborators on shared projects is a solid addition. However, the module is missing tests entirely (codecov fails), and I have a concern about the `post_init_hook` record rule rewrite: the new domain restricts task visibility to `edit_collaborator_ids`, which would seem to exclude readonly collaborators from seeing tasks -- the opposite of what the description promises. Could you add tests covering both collaborator types and double-check the record rule logic?

PR #1562, opened 2025-09-04
[16.0][FIX] project_sequence: return proper name_get values in name_search

BhaveshHeliconia:
Fix test case: I faced a test_name_search error while working on this PR → https://github.com/OCA/project/pull/1561

The test expected name_search results to include sequence_code in the display name (e.g. "23-00012 - two").

PR #1574, opened 2025-09-30
Revert "[16.0][FIX] project_sequence: return proper name_get values in name_search"

yajo:
Reverts OCA/project#1562


yajo:
This is a hotfix, as seen in https://github.com/OCA/project/pull/1562#issuecomment-3347778819, so let me use my maintainer superpowers and fast-track it.

/ocabot merge patch

PR #1578, opened 2025-10-07
[16.0][ADD] project_task_create_portal: Module added

geomer198:


project_task_create_portal/README.rst
```diff
@@ -0,0 +1,124 @@
...
+   - Deadline (optional) 
+- Press the Save button
+
+## To edit a task from the portal
+
+Note: the user can edit only their own tasks created
+
+- Go to the portal, navigate to Projects
+- Select an available project
+- Select a task in the stage permitted for task creation and task editing
```
gfcapalbo, line 80: Edit button not visible on task form. should be under   <ul>  task, history  , does not show. (explanation of reason in other comments)

project_task_create_portal/views/portal_template.xml
```diff
@@ -0,0 +1,216 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+    <template
+        id="portal_my_task_edit"
+        inherit_id="project.portal_my_task"
+        priority="20"
+    >
+        <xpath expr="//t[@t-set='entries']/ul" position="inside">
+            <li class="list-group-item" t-if="task.check_portal_edit_access()">
```
gfcapalbo, line 10: here we have our edit button, that will refer to the edit controller, only for records in correct stage and owned by users. 

Edit works perfectly, but it is incompatible with module project_task_code_portal, if that module is installed all edit functions are gone. route is still accessible by manually inserting the url.

The reason for this incompatibility lies in this line of code:  https://github.com/OCA/project/blob/16.0/project_task_code_portal/controllers/portal.py#L87
where we  freturn sudo page values: 
https://github.com/OCA/project/blob/16.0/project_task_code_portal/controllers/portal.py#L91

then leaving self.env.user as sudo in the task page.   In other comments i suggest to replace self.env.user with self.env.context['uid'], to make code work with project_task_code_portal and any other modules that may sudo the page.

project_task_create_portal/models/project_task.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @api.model
+    def _check_portal_fields_access(self):
+        return ["name", "description", "date_deadline"]
+
+    def check_portal_edit_access(self):
+        """Check if the current user has portal access to edit this task."""
+        # Portal users can only access tasks in projects with portal task creation enabled
+        if not self.project_id.is_portal_task_creation_allowed():
+            return False
```
gfcapalbo, line 37: Why would edit be forbidden , because creation is not allowed?

Yes, our edit feature is tied to "only the tasks the user has created" but this would cause problems if we have a situation a user creates many tasks, is then removed from create permissions , but still wants to modify the issues they created.


I suggest  to remove this.   edit should be depending only on the current user and   record creator .

Or create separate configuration settings for edit function.

project_task_create_portal/views/project_project_views.xml
```diff
@@ -0,0 +1,25 @@
...
+<odoo>
+    <!-- Project Project Form View -->
+    <record id="project_project_form_view_portal_task_creation" model="ir.ui.view">
+        <field name="name">project.project.form.portal.task.creation</field>
+        <field name="model">project.project</field>
+        <field name="inherit_id" ref="project.edit_project" />
+        <field name="arch" type="xml">
+            <group name="extra_settings" position="before">
+                <group
+                    string="Allow Create Task In Portal"
```
gfcapalbo: Create and edit own tasks in portal.
settings name is misleading.

project_task_create_portal/README.rst
```diff
@@ -0,0 +1,124 @@
+============================
+Project Portal Task Creation
```
gfcapalbo: Misleading name. this module supports create and has also an "own task edit" feature

project_task_create_portal/models/project_project.py
```diff
@@ -0,0 +1,40 @@
...
+            stage = project.portal_stage_id
+            if stage and stage not in project.type_ids:
+                raise ValidationError(
+                    _("The Portal Task Creation Stage must belong to this project.")
+                )
+
+    def is_portal_task_creation_allowed(self):
+        """Check if portal task creation is allowed for this project."""
+        self.ensure_one()
+        return bool(self.portal_stage_id) and self.env.user in self.portal_user_ids
```
gfcapalbo: because of a sudo in project_task_code_portal,  this will fail everytime, therefore all edit features will not show (the button is conditionally checked on  this being true)

Works if replaced by:      
`return bool(self.portal_stage_id) and self.env.context['uid']  in self.portal_user_ids.ids
`

this problem was found on runboat with all oca/project modules installed.

project_task_create_portal/models/project_task.py
```diff
@@ -0,0 +1,48 @@
...
+
+        # Portal users can only access tasks in the allowed stage
+        if self.stage_id != self.project_id.portal_stage_id:
+            return False
+
+        # Portal users can only edit their own tasks
+        if self.create_uid != self.env.user:
```
gfcapalbo: if self.env.user is sudo, as imposed in project_task_code_portal this will fail.
replace self.env.user  with self.env.context['uid']

gfcapalbo (changes requested):
Create and edit work well. tests look solid, but there are some problems.

  - incompatible with project_task_code_portal, that loads the task page with sudo user, therefore making the edit button dissapear from form (env.user != task.create_uid)
  -  the edit button is visible when accessing portal task form from projects > task , but not directly from tasks > task.  this is caused by the widespread sudo problem in portal templates.    solved by:
```
          def is_portal_task_creation_allowed(self):
                 self.ensure_one()
                 return bool(self.portal_stage_id) and self.env.context['uid'] in self.portal_user_ids.ids
```
   created an MR to this branch with the proposed fixes for this problem.

  -  settings name only refers to "create" , make clear also "edit" is possible.
  -  in settings make clear that stage_id is bothe the stage the new task will be created in and also the only stage it will be allowed to be edited in.
  -  Optionally, refactor to have, also separate "edit" setting on project form.
  -  Module name is misleading, it allows to edit and create tasks.

project_task_create_portal/models/project_task.py
```diff
@@ -0,0 +1,48 @@
...
+        return True
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        """Override create to handle portal user restrictions."""
+        records = super().create(vals_list)
+        if not self.env.user.has_group("base.group_portal"):
+            return records
+        for record in self:
+            if not record.project_id.is_portal_task_creation_allowed():
```
KKamaa: I think here you meant to loop through the record set returned by ``records = super().create(vals_list)`` otherwise I don't think the call to ``is_portal_task_creation_allowed()`` is being executed.

project_task_create_portal/models/project_task.py
KKamaa: Just to add to add it here, In the write where ``check_portal_edit_access`` is called you might want to add ``self.ensure_one()`` because write call can be called with multi-records. Either add ``self.ensure_one()`` or ``for loop`` that will loop over self just in case its called with multi recordset.

project_task_create_portal/models/project_task.py
```diff
@@ -0,0 +1,48 @@
...
+                    _("You are not allowed to create tasks in this project.")
+                )
+        records.write({"user_ids": [(6, 0, [])]})
+        return records
+
+    def write(self, vals):
+        """Override write to handle portal user restrictions."""
+        if (
+            self.env.user.has_group("base.group_portal")
+            and not self.check_portal_edit_access()
```
KKamaa: this is what I meant when write is called and check_portal_edit_access isn't secured to check multiple records or single check.

KKamaa (changes requested):

geomer198:
@gfcapalbo, @KKamaa
Thanks for the code review! I've updated the code based on your feedback. Could you please review again to confirm everything is correct?

project_task_create_portal/models/project_task.py
```diff
@@ -0,0 +1,49 @@
...
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    @api.model
+    def _check_portal_fields_access(self):
+        return ["name", "description", "date_deadline"]
+
+    def check_portal_edit_access(self):
+        """Check if the current user has portal access to edit this task."""
+        # Portal users can only access tasks in projects with portal task creation enabled
```
alexey-pelykh, line 34: Bug: `for record in self` iterates the model (empty recordset in `create` context), not the newly created records. Should be `for record in records`. As written, the permission check after `super().create()` is never executed for portal users.

project_task_create_portal/models/project_task.py
```diff
@@ -0,0 +1,49 @@
...
+        self.ensure_one()
+        if not self.project_id.is_portal_task_creation_allowed():
+            return False
+
+        # Portal users can only access tasks in the allowed stage
+        if self.stage_id != self.project_id.portal_stage_id:
+            return False
+
```
alexey-pelykh, line 42: Security: the `write()` override does not restrict which fields a portal user can write. A crafted request bypassing the controller could modify `stage_id`, `project_id`, `priority`, etc. Consider filtering `vals` against allowed fields when the user is a portal user.

project_task_create_portal/static/src/js/portal.js
```diff
@@ -0,0 +1,98 @@
...
+                })
+                .then((wysiwyg) => {
+                    wysiwyg.toolbar.$el.find("#link, #media").remove();
+                    this.$el
+                        .find(".note-editable")
+                        .find("img.float-start")
+                        .removeClass("float-start");
+                    this._wysiwyg = wysiwyg;
+                    this._wysiwyg.$editable.addClass("bg-white p-2");
+                    console.log(this._wysiwyg);
```
alexey-pelykh: Debug statement: `console.log(this._wysiwyg)` must be removed before merge.

project_task_create_portal/static/src/js/portal.js
```diff
@@ -0,0 +1,98 @@
...
+                });
+
+            return Promise.all([def, loadProm]);
+        },
+
+        /**
+         * @private
+         * @param {Event} ev
+         */
+        _onEditProfilePicClick: function (ev) {
```
alexey-pelykh: Dead code: `_onEditProfilePicClick`, `_onFileUploadChange`, and `_onProfilePicClearClick` are not bound to any events in this widget. These appear copy-pasted from `website_forum`. Please remove.

project_task_create_portal/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,8 @@
+## To configure the project for task creation from the portal
+
+- Go to the Project App
+- On a project kanban card, press the Settings  button
+- Go to the "Settings" tab
+
+- Set the **Portal Task Creation Stage** to a stage (from the project’s stages) from which portal users will be allowed to add tasks. 
+- Se the **Hide assigned user** to hide the assigned user of the task on the portal.
```
alexey-pelykh: Typo: "Se the" should be "Set the".

alexey-pelykh (changes requested):
Thanks for this contribution. The concept of portal task create/edit is genuinely useful and the test coverage is solid. However, there are several issues that need to be addressed before this can be merged.

**Security -- missing `ir.model.access.csv` and record rules.** This module adds new fields and behaviors for portal users but does not ship any `security/ir.model.access.csv` file. The controller uses `sudo()` to bypass access checks, but the model-level `write()` override relies on application-level logic rather than proper Odoo record rules. You should define proper `ir.rule` record rules restricting portal users' access.

**Security -- no field-level write restriction in `write()`.** The `write()` override checks `check_portal_edit_access()` but does not restrict which fields a portal user may modify. A crafted POST request could include `stage_id`, `project_id`, `user_ids`, or any other field, and the `sudo().write()` in the controller would apply them.

**Security -- `create()` iterates `self` instead of `records`.** In `project_task.py`, `for record in self` iterates the model class rather than the newly created records. The permission check after creation is never actually executed.

**Manifest -- syntax issue in `summary`.** The summary value spans two lines as two implicit string literals -- please verify the concatenation works as intended.

**Other issues:**
- `console.log(this._wysiwyg)` left in `portal.js` -- debug statement, must be removed.
- Dead code in `portal.js`: `_onEditProfilePicClick`, `_onFileUploadChange`, `_onProfilePicClearClick` are never bound to events. Remove them.
- Sorting entry key `create_date` with label "Created by" and order `create_uid desc` is inconsistent.
- README files use Markdown syntax inside `.rst` files.
- Typo in CONFIGURE.rst: "Se the" should be "Set the".

Happy to re-review once updated.

geomer198:
@alexey-pelykh Could you please check changes?

PR #1580, opened 2025-10-09
[16.0][IMP] project_key: URL link can redirect to portal

HekkiMelody:
Simple fix to reuse the existing feature from the `portal` module instead of manually building the URL. As a bonus, portal users can now also use the same link.

(I'm working on another PR to also display the key on the portal EDIT: #1581 )


quirino95 (approved):

rvalyi (approved):

rvalyi:
merging as major as there is a new dependency (though all instance will have the portal module installed)
/ocabot merge major

PR #1585, opened 2025-10-16
[ADD] simplify onChange and remove user_ids, tag

gfcapalbo:
changing  the 14.0 version of the module,  to make it more similar to former versions.
User_id and tag_ids will not overwrite existing user and tags if left empty on template, therefore mirroring behaviour of  v15-v18.  
They will still work, for legacy users (user_id and tag_ids was removed from v15-v18).

Description will be appended. As it is done in v15-v18 versions.

tests updated. new tests for empty user_id, empty tag_ids cases.

### DIFF
--- a/project_task_description_template/models/project_task.py
+++ b/project_task_description_template/models/project_task.py
@@ -1,5 +1,6 @@
 # Copyright (C) 2023 Cetmix OÜ
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
+import re
 
 from odoo import api, fields, models
 
@@ -36,9 +37,13 @@ class ProjectTask(models.Model):
     @api.onchange("task_template_id")
     def _onchange_task_template_id(self):
         if self.task_template_id:
-            description = self.description if self.description else ""
+            description = self.description or ""
+            template_desc = self.task_template_id.description or ""
+            # Avoid multiple template additions when multiple onchanges are triggered
+            if not re.search(rf"{re.escape(template_desc)}", description):
+                description += template_desc
             update_dict = {
-                "description": description + self.task_template_id.description,
+                "description": description,
             }
             if self.task_template_id.tag_ids:
                 update_dict["tag_ids"] = self.task_template_id.tag_ids.ids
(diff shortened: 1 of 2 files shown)


project_task_description_template/models/project_task.py
```diff
@@ -36,10 +36,5 @@ def _compute_template_visible(self):
...
         if self.task_template_id:
-            self.update(
-                {
-                    "user_id": self.task_template_id.user_id.id,
-                    "tag_ids": self.task_template_id.tag_ids.ids,
-                    "description": self.task_template_id.description,
-                }
-            )
+            description = self.description if self.description else ""
+            self.description = description + self.task_template_id.description
```
thomaspaulb: Less invasive towards the previous user of this module would be perhaps to only replace the tags or the user_id when these are filled in on the template, and otherwise leave them alone. In that case the fields and tests can stay, but at least the module is not destructive.

gfcapalbo: @thomaspaulb our initial specs were: 
"In all the 15.0-18.0 versions of it, it just assigns the description, but in 14.0 it does other things that we don't want, like update the task user_id and overwrite the description"-

We did not want it to update user_id.  wich in latter versions does not exist. Will produce a simplified version of this where it will not overwrite it if template has user_id null.

thomaspaulb: Yes, sorry, here I am to blame - for the customer the solution I specced was good, but now that I see it proposed to OCA, I think it can never get merged unless we also support the old case.

gfcapalbo: all done. green.

new version will append description, and not overwrite user_id if left empty.
In a seperate commit , i also added tag_ids not overwritten, not specified in specs, but still i believe necessary.

gfcapalbo:
@thomaspaulb added a regex to avoid duplicates on multiple onchange calls. the duplicate problem exists in  v15-v18 as well (probably in  v19  , MR just  recently  posted) . everytime an onchange is triggered the template duplicates.

thomaspaulb:
I suggest we close this; the 14.0 version differs too much from 15-19 versions to warrant fixing as we are migrating

PR #1588, opened 2025-10-17
[18.0][IMP] project_timesheet_time_control: sync unit_amount with date_time and date_time_end on form view

CRogos:
1) In the Form view, the unit_amount field is only updated when saving the form, not when updating the date_time_end field.

image

2) The second commit synchronise the date field with the date_time. Before it was possible that date and date_time field had different date values when data was written to the date field. (Can be tested by makin the date field visible in the form and change/change+save the date field)

image

3) Could you also have a look at this module, whether this should be displayed in a different module or could be added to this module? https://github.com/OCA/timesheet/pull/833

### DIFF
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -92,9 +92,16 @@ class AccountAnalyticLine(models.Model):
             else:
                 one.show_time_control = "stop"
 
+    @api.onchange("date")
+    def _onchange_date(self):
+        hour_uom = self.env.ref("uom.product_uom_hour")
+        if self.product_uom_id == hour_uom and self.date_time:
+            self.date_time = datetime.combine(self.date, self.date_time.time())
+
     @api.onchange("date_time", "date_time_end")
     def _onchange_date_time(self):
         hour_uom = self.env.ref("uom.product_uom_hour")
+        self.date = self.date_time.date()
         if self.product_uom_id == hour_uom:
             if self.date_time and self.date_time_end:
                 self.unit_amount = self._duration(self.date_time, self.date_time_end)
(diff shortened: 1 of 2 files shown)


CRogos:
@pedrobaeza could you have a look into this PR. This one is changing the existing module, but only fixing some bugs.

MohamedOsman7 (approved):
Code & functional review LGMT

CRogos:
@lbarry-apsl could you have a look into this change, because it is also related to hr_timesheet_calendar.

CRogos:
@christian-ramos-tecnativa could you also have a look on this PR? This is also included in the v19 migration.

CRogos:
@victoralmau could you do the second review?

lbarry-apsl (approved):
Code and runboat reviewed. Working correctly!

I prefer that the time spent is updated as soon as I enter the data, without having to save. For example, in my daily entries, I like to see that the recorded hours are my total Time Worker in Toggle. Waiting for it to save before confirming slows me down a bit.

CRogos:
@lbarry-apsl I just found a minor issue. The _eval_date was called twice, and to be inline with the default value, I now add the time  of now.

There is also a second bugfix here: https://github.com/OCA/timesheet/pull/844

CRogos:
@pedrobaeza could you have a look and if OK, merge?

CRogos:
@victoralmau could you merge?

CRogos:
@pedrobaeza can you please merge?

pedrobaeza (commented):
Please expand `_eval_date` instead, and I don't think the module should be overloaded with onchanges.

CRogos:
@pedrobaeza Thanks for your feedback.  
I think the onchange is necessary as you can see from test_onchange_date_time_with_hour_uom_no_end_date. Otherwise the date can be stored incorrect. 

What do you mean with "expand _eval_date"?

pedrobaeza:
I mean to do the changes on `_eval_date` instead on create/write: https://github.com/c4a8-odoo/module-oca-project/blob/18.0/project_timesheet_time_control/models/account_analytic_line.py#L56

and can't you switch fields to computed writable instead of onchanges?

CRogos:
@pedrobaeza are these changes in your favor?

Currently the create/write favors the 'date_time' value. When we have no onchange and a form updates the 'date' field, the 'date_time' field stays unchanged and is submitted with the old value. Therefore the value will not be changed in the database. Because the date_time field is introduced here, this issue should also be handled here. Otherwise there could be several modules adding this kind of code. 

Moving to _eval_date is not possible for the remaining write, because it may require the time part of data_time of reach record.

Feel free to give further feedback.

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -92,17 +99,49 @@ def _compute_show_time_control(self):
...
         return super().create(list(map(self._eval_date, vals_list)))
 
     def write(self, vals):
-        return super().write(self._eval_date(vals))
+        if "date" in vals and "date_time" not in vals:
+            res = super(
+                AccountAnalyticLine, self.filtered(lambda r: not r.date_time)
+            ).write(self._eval_date(vals))
+
+            # only overwrite the date part of date_time
```
pedrobaeza: This part is very ugly. Why do you need it? And please don't empty lines inside a method.

CRogos: When date_time is not set, all records can be updated in batch. If date_time is set, each line needs to be updated individually. 
I've changed the code, so that there is only one call for batch update.

CRogos: @pedrobaeza I know you have a lot work to do, but can we bring this to an end? If you agree with this solution I can squash the last two commits, or you merge as is.

pedrobaeza (approved):
/ocabot merge patch

CRogos:
Do I need to rebase?

PR #1592, opened 2025-10-22
[16.0][IMP] project_stock: Add the option to cancel moves in a task

victoralmau:
Add the option to cancel moves in a task

Similar process to the existing "Cancel materials" in the task (applies to all moves).

ejemplo

Please @carlos-lopez-tecnativa and @pedrobaeza can you review it?

@Tecnativa TT58839

### DIFF
--- a/project_stock/models/account_analytic_line.py
+++ b/project_stock/models/account_analytic_line.py
@@ -9,6 +9,9 @@ class AccountAnalyticLine(models.Model):
     stock_task_id = fields.Many2one(
         comodel_name="project.task", string="Project Task", ondelete="cascade"
     )
+    stock_move_id = fields.Many2one(
+        comodel_name="stock.move", string="Stock Move", ondelete="cascade", copy=False
+    )
 
     def _timesheet_postprocess_values(self, values):
         """When hr_timesheet addon is installed, in the create() and write() methods,
--- a/project_stock/views/stock_move_view.xml
+++ b/project_stock/views/stock_move_view.xml
@@ -69,6 +69,15 @@
                 />
                 <field name="quantity_done" string="Consumed" readonly="1" />
                 <field name="group_id" invisible="1" />
+                <field name="show_cancel_button_in_task" invisible="1" />
+                <button
+                    name="action_cancel_from_task"
+                    type="object"
+                    class="btn-secondary"
+                    string="Cancel"
+                    attrs="{'invisible': [('show_cancel_button_in_task', '=', False)]}"
+                    confirm="Are you sure you want to proceed?"
+                />
             </tree>
         </field>
     </record>
--- a/project_stock/README.rst
+++ b/project_stock/README.rst
@@ -71,6 +71,12 @@ Usage
     * `Cancel Materials`: Set the moves of the products as cancelled.
     * `Scrap`: Allows the defined products to be scrapped.
 
+Known issues / Roadmap
+======================
+
+#. Lot compatibility (similar to what happens in pickings).
+#. Release stock reserves if the task is deleted.
+
 Bug Tracker
 ===========
 
--- a/project_stock/models/project_task.py
+++ b/project_stock/models/project_task.py
@@ -203,8 +203,11 @@ class ProjectTask(models.Model):
         """Cancel the stock moves and remove the analytic lines created from
         stock moves when cancelling the task.
         """
-        self.mapped("move_ids.move_line_ids").write({"qty_done": 0})
+        for move in self.move_ids:
+            move.action_cancel_from_task()
         # Use sudo to avoid error for users with no access to analytic
+        # Although the action_cancel_from_task() method should have deleted
+        # all analytical entries, we want to make sure there are no old records left.
         self.sudo().stock_analytic_line_ids.unlink()
         self.stock_moves_is_locked = True
         return True
--- a/project_stock/models/stock_move.py
+++ b/project_stock/models/stock_move.py
@@ -1,6 +1,7 @@
 # Copyright 2022-2025 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
 from odoo import api, fields, models
+from odoo.tools import float_is_zero
 
 
 class StockMove(models.Model):
@@ -14,6 +15,14 @@ class StockMove(models.Model):
     raw_material_task_id = fields.Many2one(
         comodel_name="project.task", string="Task for material", check_company=True
     )
+    analytic_line_ids = fields.One2many(
+        comodel_name="account.analytic.line",
+        inverse_name="stock_move_id",
+        string="Analytic Lines",
+    )
+    show_cancel_button_in_task = fields.Boolean(
+        compute="_compute_show_cancel_button_in_task"
+    )
 
     @api.onchange("product_id")
     def _onchange_product_id(self):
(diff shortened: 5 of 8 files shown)


project_stock/models/account_analytic_line.py
```diff
@@ -9,6 +9,9 @@ class AccountAnalyticLine(models.Model):
     stock_task_id = fields.Many2one(
         comodel_name="project.task", string="Project Task", ondelete="cascade"
     )
+    stock_move_id = fields.Many2one(
+        comodel_name="stock.move", string="Stock Move", ondelete="cascade"
+    )
```
carlos-lopez-tecnativa, line 14:
```suggestion
    stock_move_id = fields.Many2one(
        comodel_name="stock.move", string="Stock Move", ondelete="cascade", copy=False
    )
```

project_stock/views/stock_move_view.xml
```diff
@@ -69,6 +69,14 @@
...
                 <field name="quantity_done" string="Consumed" readonly="1" />
                 <field name="group_id" invisible="1" />
+                <field name="show_cancel_button_in_task" invisible="1" />
+                <button
+                    name="action_cancel_from_task"
+                    type="object"
+                    class="btn-secondary"
+                    string="Cancel"
+                    attrs="{'invisible': [('show_cancel_button_in_task', '=', False)]}"
+                />
```
carlos-lopez-tecnativa, line 80: Can a group of project users or inventory users be added?
Additionally, please add the `confirm` attribute.
```suggestion
                <button
                    name="action_cancel_from_task"
                    type="object"
                    class="btn-secondary"
                    string="Cancel"
                    attrs="{'invisible': [('show_cancel_button_in_task', '=', False)]}"
                    confirm="Are you sure you want to proceed?"
                />
```

carlos-lopez-tecnativa (changes requested):
Sorry, but I didn’t test the module. I got this error when I tried to save the record.
image

project_stock/views/stock_move_view.xml
victoralmau: Thanks, I've added the confirm.

It is not necessary to define groups for the button because this tree view is only used in `project_stock`, and stock moves are only displayed if you have the corresponding group: https://github.com/OCA/project/blob/f6200ce3a365483b12e215abc1b165325e059da9/project_stock/views/project_task_view.xml#L112

victoralmau:
> Sorry, but I didn’t test the module. I got this error when I tried to save the record. image

The error you mention occurred because the task did not have any type of operation specified (in the Extra info tab). 
Remember that if this information is set in the project **before** creating the task, it will be propagated to the task.

carlos-lopez-tecnativa (changes requested):
> The error you mention occurred because the task did not have any type of operation specified (in the Extra info tab). Remember that if this information is set in the project **before** creating the task, it will be propagated to the task.

It’s curious because I tested it again, and it can now be saved normally. However, I see that the `Operation Type` field on the `Extra Info` page is not filled. It seems that, according to the code, this field is not related to the project nor populated by an onchange, so what you mentioned doesn’t happen.

image

Anyway, I left some points to review in this flow.

My settings:
**ST2:** 
```
   Use Stock Moves > True
   Done Stock Moves > False
```

**ST3:** 
```
   Use Stock Moves > True
   Done Stock Moves > True
```

**Case 1: Try to cancel or delete a line**
When the task is in stage ST2, I added the products. In this stage, I click `Confirm Materials`; if later I don’t want to move this product, I cannot cancel or delete the line. What can I do in this case?

Could you display the Cancel button in this case?

image

image


**Case 2: Select lots or serials to consume**
When the product is tracked by lot or serial, I cannot indicate which lots or serials to use. Could you add a button to see detailed operations and allow selecting them?

image

Similar to how Odoo does it in stock.picking:
image

**Case 3: Delete a task with reserved products**
When I delete a task, the reservation is preserved. I think this is a mistake, the reservation should be unreserved in this case.

victoralmau:
> > The error you mention occurred because the task did not have any type of operation specified (in the Extra info tab). Remember that if this information is set in the project **before** creating the task, it will be propagated to the task.
> 
> It’s curious because I tested it again, and it can now be saved normally. However, I see that the `Operation Type` field on the `Extra Info` page is not filled. It seems that, according to the code, this field is not related to the project nor populated by an onchange, so what you mentioned doesn’t happen.

Sorry, I didn't remember exactly how it worked without checking the code. It's done with the defaults in https://github.com/OCA/project/blob/f6200ce3a365483b12e215abc1b165325e059da9/project_stock/models/stock_move.py#L79.

> image
> 
> Anyway, I left some points to review in this flow.
> 
> My settings: **ST2:**
> 
> ```
>    Use Stock Moves > True
>    Done Stock Moves > False
> ```
> 
> **ST3:**
> 
> ```
>    Use Stock Moves > True
>    Done Stock Moves > True
> ```
> 
> **Case 1: Try to cancel or delete a line** When the task is in stage ST2, I added the products. In this stage, I click `Confirm Materials`; if later I don’t want to move this product, I cannot cancel or delete the line. What can I do in this case?
> 
> Could you display the Cancel button in this case?
> image image
> 

It is important to note that this PR aims to add functionality to existing behaviors, therefore, the Cancel button will only be displayed under the same conditions as the “Cancel materials” button, i.e., when the material has been transferred (stock moves in done status) and the task is in stage ST3.

http://oca-project-16-0-pr1592-88b993c72d3e.runboat.odoo-community.org/web#id=44&cids=1&menu_id=112&action=203&active_id=5&model=project.task&view_type=form

> **Case 2: Select lots or serials to consume** When the product is tracked by lot or serial, I cannot indicate which lots or serials to use. Could you add a button to see detailed operations and allow selecting them?
> image
> 
> Similar to how Odoo does it in stock.picking: image
> 

You're right, but currently batches are not taken into account in this module. I will add it to the ROADMAP so that it can be considered in the future.

> **Case 3: Delete a task with reserved products** When I delete a task, the reservation is preserved. I think this is a mistake, the reservation should be unreserved in this case.

Again, this is not currently being taken into account, but the best option can probably be discussed in another PR.

carlos-lopez-tecnativa (changes requested):
> the Cancel button will only be displayed under the same conditions as the “Cancel materials” button, 

Based on your comment, this doesn’t seem to be happening. I can see the Cancel button at the task level, but not at the line level when the materials haven’t been transferred yet. I only see the button once the stock moves are done.
image

> > **Case 3: Delete a task with reserved products** When I delete a task, the reservation is preserved. I think this is a mistake, the reservation should be unreserved in this case.
> 
> Again, this is not currently being taken into account, but the best option can probably be discussed in another PR.

Hmm, I consider this an issue, but I understand your point. Could you add it to the roadmap in the meantime, until this issue is resolved?

victoralmau:
> > the Cancel button will only be displayed under the same conditions as the “Cancel materials” button,
> 
> Based on your comment, this doesn’t seem to be happening. I can see the Cancel button at the task level, but not at the line level when the materials haven’t been transferred yet. I only see the button once the stock moves are done. image
> 
You need to transfer the materials in order to cancel it later.

carlos-lopez-tecnativa (approved):
LGTM. Please don’t forget to add to the ROADMAP that when a stock.move is reserved, this reservation is not released if the task is deleted.

victoralmau:
Ping @pedrobaeza

project_stock/models/account_analytic_line.py
```diff
@@ -9,6 +9,9 @@ class AccountAnalyticLine(models.Model):
     stock_task_id = fields.Many2one(
         comodel_name="project.task", string="Project Task", ondelete="cascade"
     )
+    stock_move_id = fields.Many2one(
+        comodel_name="stock.move", string="Stock Move", ondelete="cascade", copy=False
```
pedrobaeza, line 13: Adding this field now will make the previous lines orphaned. Aren't you able to do it without this field?

victoralmau: It is possible to do without this field, but it is not recommended. 
In fact, this field should have been added when the module was created. 
Here are some reasons why it is necessary:

- It is possible to consume the same product several times in a task (stock moves). In that case, there would be several linked analytical lines and we would not know exactly which one corresponds to each one, especially if the same quantity is consumed.
- It is possible to modify the name of the task or product name after creating the analytical lines, so it would not be appropriate to differentiate the corresponding analytical entry by filtering by name https://github.com/Tecnativa/project/blob/7c0656a5540c90efb22057b1e40bccd6f1e609e7/project_stock/models/stock_move.py#L105

Until now, we had not realized that this `stock_move_id` field in the analytical lines is necessary because, until now, all linked stock moves were “canceled” using the button in the task, and therefore all analytical lines were deleted. However, if we want to “cancel” only one stock moves, that is when this link proves to be necessary.

In any case, compatibility with “old analytical entries” has been taken into account at https://github.com/Tecnativa/project/blob/7c0656a5540c90efb22057b1e40bccd6f1e609e7/project_stock/models/stock_move.py#L65

pedrobaeza (approved):
OK, thanks for the explanations. Let's go with it.

/ocabot merge major

PR #1594, opened 2025-11-01
[19.0][MIG] project_task_stage_mgmt: Migration to 19.0

NICO-SOLUTIONS:
- Standard migration to 19


CristianoMafraJunior (approved):

NICO-SOLUTIONS:
@leemannd 
the dummy test is to avoid...

WARNING odoo odoo.tests.result: 0 failed, 0 error(s) of 0 tests when loading database 'odoo'

Maybe this have been changed. i can test, if the warning is still arised without it.

leemannd:
Hello @NICO-SOLUTIONS , I prefer to have a warning than a fake test.

NICO-SOLUTIONS:
> Hello @NICO-SOLUTIONS , I prefer to have a warning than a fake test.

@leemannd 
Well... i got this approach from another OCA maintainer. To get rid of it. 
So i think clear guidelines Could help

pedrobaeza:
/ocabot migration project_task_stage_mgmt

@NICO-SOLUTIONS can you point where that advise was given. Anyway, that's not a guideline.

NICO-SOLUTIONS:
@pedrobaeza 
sure.... https://github.com/OCA/website/pull/1065#issuecomment-2487674953

pedrobaeza:
Thank you for the reference.

Yeah, indeed when the first module of the repository doesn't have tests, the only solution to merge it is to do that, but that doesn't mean all the modules should use the same trick. Now that this module is not the first one in the repository, just rebase and remove the fake test.

NICO-SOLUTIONS:
> Thank you for the reference.
> 
> Yeah, indeed when the first module of the repository doesn't have tests, the only solution to merge it is to do that, but that doesn't mean all the modules should use the same trick. Now that this module is not the first one in the repository, just rebase and remove the fake test.

@pedrobaeza 
Ok got it...

leemannd (approved):

NICO-SOLUTIONS:
@pedrobaeza  done

NICO-SOLUTIONS:
@leemannd thx!

PR #1599, opened 2025-11-12
[16.0][ADD] project_update_portal: add new module

marcelsavegnago:
This module extends project updates functionality by allowing portal access for project and update followers.

## Main Features

* **Portal Access for Followers**: Allows project followers and update followers to access project updates through the portal
* **Automatic Access Control**: Automatically grants access based on follower relationships
* **Update List View**: Portal users can view all updates for projects they follow
* **Update Detail View**: Portal users can view detailed information about specific updates
* **Breadcrumb Navigation**: Integrated breadcrumb navigation for easy access to updates

## Benefits

* Keep stakeholders informed about project progress without requiring full system access
* Share project updates with external partners and clients through the portal
* Maintain transparency by allowing followers to track project status
* Easy access to project updates through a dedicated portal interface

## Dependencies

This module requires:
* `project`: Odoo's project management module
* `portal`: Odoo's portal module for customer access

cc @parzewski @kaynnan


CristianoMafraJunior (approved):
LGTM

WesleyOliveira98 (approved):
LGTM

project_update_portal/controllers/portal.py
```diff
@@ -0,0 +1,99 @@
...
+        }
+        return self._get_page_view_values(
+            update, access_token, values, "my_project_updates_history", False, **kwargs
+        )
+
+    @http.route(
+        ["/my/projects/<int:project_id>/update/<int:update_id>"],
+        type="http",
+        auth="public",
+        website=True,
```
alexey-pelykh, line 29: This route uses `auth="public"` while the list route at line 57 uses `auth="user"`. This inconsistency means unauthenticated users can probe this endpoint. The redirect behavior (`/my` vs `/my/projects/<id>`) leaks whether a given project_id exists. Change to `auth="user"` for consistency.

project_update_portal/controllers/portal.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+    def portal_my_project_update(
+        self, project_id=None, update_id=None, access_token=None, **kw
+    ):
+        """Route to view a specific project update"""
+        try:
+            # Check project access first
+            self._document_check_access("project.project", project_id, access_token)
```
alexey-pelykh: The same `access_token` is used to validate both the project and the update. Portal access tokens are record-specific -- a token for `project.update` won't pass `_document_check_access` for `project.project` and vice versa.

Consider either removing the project access check here (the update->project_id integrity check is sufficient after validating the update), or checking project access without the token.

project_update_portal/models/project_update.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2025 Marcel Savegnago - Escodoo <https://escodoo.com.br>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, models
+
+
+class ProjectUpdate(models.Model):
+    _name = "project.update"
+    _inherit = ["project.update", "portal.mixin"]
+
```
alexey-pelykh, line 10: `message_partner_ids` in `@api.depends` but never used in the method body. Remove to avoid needless recomputation on follower changes:
```suggestion
    @api.depends("project_id")
```

project_update_portal/views/portal_project_update_templates.xml
```diff
@@ -0,0 +1,150 @@
...
+                                            </small>
+                                        </p>
+                                        <div class="progress" style="height: 5px;">
+                                            <div
+                                                t-attf-class="progress-bar bg-{{ 'success' if update.status == 'on_track' else 'warning' if update.status == 'at_risk' else 'danger' if update.status == 'off_track' else 'info' }}"
+                                                t-attf-style="width: {{ update.progress }}%"
+                                                role="progressbar"
+                                                aria-valuenow="update.progress"
+                                                aria-valuemin="0"
+                                                aria-valuemax="100"
```
alexey-pelykh, line 133: `aria-valuenow` is a literal string `"update.progress"` instead of the actual value. Should be `t-attf-aria-valuenow="{{ update.progress }}"` to match the detail view template.

alexey-pelykh (changes requested):
Thanks for the contribution -- exposing project updates in the portal is a genuinely useful feature. The module structure, tests, and documentation are solid overall. However, there are a few security and correctness issues to address.

**Security: `auth="public"` on detail route is too permissive.** The detail route uses `auth="public"` while the list route uses `auth="user"`. Unauthenticated visitors can probe the detail endpoint, and redirect behavior leaks whether a project ID exists. Both routes should be consistent.

**Security: single `access_token` validates two different records.** The same token is used to check access to `project.project` and `project.update`. Portal tokens are record-specific -- a token for one won't pass `_document_check_access` for the other. For portal followers this is bypassed via ir.rule, but for token-based flows both checks will fail.

**Correctness: unnecessary `@api.depends`.** `_compute_access_url` depends on `message_partner_ids` but only uses `project_id` and `id`. Adding/removing followers would trigger needless recomputation.

**Accessibility: `aria-valuenow` in list template.** The progress bar has literal string `"update.progress"` instead of the actual value via `t-attf-aria-valuenow`.

Happy to re-review once updated.

CristianoMafraJunior:
@alexey-pelykh The corrections from your review have been made. Could you please check again?

alexey-pelykh (approved):
All four items from the previous review have been addressed:

1. Detail route now uses `auth="user"`, consistent with the list route.
2. Project access check no longer passes the `access_token` -- token is only used for the update record, project access relies on ir.rule. Correct separation.
3. `@api.depends` reduced to `("project_id")` only.
4. `aria-valuenow` uses `t-attf-aria-valuenow` with the actual value in both list and detail templates.

Looks good, thank you for the quick turnaround.

PR #1600, opened 2025-11-14
[16.0][IMP] project_task_code_portal: add code to sharing views and sharing URL.

JordiBForgeFlow:
The portal url is useful when you want to share the link with other project contributors. Otherwise an internal user will not know, when accessing the task from the backend, what would be the url to use when sharing the task with an external contributor

image

@halbtonjazz

### DIFF
--- a/project_task_code_portal/models/project_task.py
+++ b/project_task_code_portal/models/project_task.py
@@ -1,12 +1,34 @@
 # Copyright (C) 2025 Cetmix OÜ
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import models
+from odoo import fields, models
 
 
 class ProjectTask(models.Model):
     _inherit = "project.task"
 
+    portal_url = fields.Char(compute="_compute_portal_url")
+    portal_url_visible = fields.Boolean(compute="_compute_portal_url")
+
     @property
     def SELF_READABLE_FIELDS(self):
-        return super().SELF_READABLE_FIELDS | {"code"}
+        return super().SELF_READABLE_FIELDS | {
+            "code",
+            "portal_url",
+            "portal_url_visible",
+        }
+
+    @property
+    def SELF_WRITABLE_FIELDS(self):
+        return super().SELF_WRITABLE_FIELDS | {"code"}
+
+    def _compute_portal_url(self):
+        for rec in self:
+            rec.portal_url = ""
+            rec.portal_url_visible = False
+            if rec.project_id.privacy_visibility == "portal":
+                rec.portal_url_visible = True
+                base_url = rec.get_base_url()
+                rec.portal_url = (
+                    f"{base_url}/my/projects/{rec.project_id.id}/task/{rec.code}"
+                )
--- a/project_task_code_portal/views/project_sharing_views.xml
+++ b/project_task_code_portal/views/project_sharing_views.xml
@@ -1,0 +1,67 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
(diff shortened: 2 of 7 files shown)


JordiBForgeFlow:
@ivs-cetmix what do you think about this one?

ivs-cetmix (changes requested):
@JordiBForgeFlow  thank you, this is a nice update! 
Some UX improvement suggestion, also please update the documentation accordingly.

project_task_code_portal/views/project_sharing_views.xml
```diff
@@ -0,0 +1,61 @@
...
+        <field name="name">project.sharing.project.task.view.form</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.project_sharing_project_task_view_form" />
+        <field name="arch" type="xml">
+            <field name="name" position="before">
+                <field name="code" class="oe_inline" />
+                <span class="oe_inline"> - </span>
+            </field>
+            <field name="tag_ids" position="after">
+                <field name="portal_url" widget="url" />
```
ivs-cetmix: We can make it even more convenient by allowing to copy the url instantly with a simple click. Here is an [example](https://github.com/cetmix/cetmix-tower/blob/973cd136f6b465956d958e640bf157f88c868908/cetmix_tower_server/views/cx_tower_key_view.xml#L16)

ivs-cetmix (changes requested):

JordiBForgeFlow:
@ivs-cetmix very nice UI addition!

image

JordiBForgeFlow:
@ivs-cetmix Updated documentation. The URL is only visible in the task when the project visibility is 'Portal'.

ivs-cetmix (approved):
@JordiBForgeFlow looks good now, thank you!

project_task_code_portal/models/project_task.py
```diff
@@ -1,12 +1,36 @@
...
+            "code",
+            "portal_url",
+            "portal_url_visible",
+        }
+
+    @property
+    def SELF_WRITABLE_FIELDS(self):
+        return super().SELF_WRITABLE_FIELDS | {"code"}
+
+    def _compute_portal_url(self):
```
ivs-cetmix, line 25: This function should be covered by tests to maintain the Codecov thresholds.

ivs-cetmix (changes requested):
Sorry, missed one point 😄

project_task_code_portal/models/project_task.py
JordiBForgeFlow: @ivs-cetmix Done!

project_task_code_portal/models/project_task.py
```diff
@@ -1,12 +1,36 @@
...
+        for rec in self:
+            rec.portal_url = ""
+            rec.portal_url_visible = False
+            if rec.project_id.privacy_visibility == "portal":
+                rec.portal_url_visible = True
+                base_url = rec.get_base_url()
+                rec.portal_url = "%s/my/projects/%s/task/%s" % (
+                    base_url,
+                    rec.project_id.id,
+                    rec.code,
```
rvalyi: would have been more furure proof to use an f-string here but not something blocking.

rvalyi (approved):

project_task_code_portal/models/project_task.py
JordiBForgeFlow: @rvalyi fixed!

ivs-cetmix (approved):
@JordiBForgeFlow thank you for your contribution!
LGTM

JordiBForgeFlow:
@ivs-cetmix can you merge?

PR #1606, opened 2025-11-18
[16.0][FIX] project_task_description_template: no duplicate description on template onchange

ntsirintanis:


ntsirintanis:
@gfcapalbo

gfcapalbo:
@ntsirintanis  check out: https://github.com/OCA/project/pull/1585/commits/32a77d63761bf9faef4b50cd337ef9b97fc494af

project_task_description_template/models/project_task.py
```diff
@@ -13,6 +13,14 @@ class ProjectTask(models.Model):
...
     @api.onchange("description_template_id")
     def _onchange_description_template_id(self):
-        if self.description_template_id:
-            description = self.description if self.description else ""
-            self.description = description + self.description_template_id.description
+        for task in self:
+            if not task.description_template_id:
+                continue
+            template_text = task.description_template_id.description or ""
+            if not template_text:
```
gfcapalbo: this code is extra.  it is strange we allow empty description, but if there is one, it would change nothing.  just extra code. not needed.

project_task_description_template/models/project_task.py
```diff
@@ -13,6 +13,14 @@ class ProjectTask(models.Model):
...
+                continue
+            current = task.description or ""
+            # Avoid duplicating the same template content
+            if current == template_text or current.endswith(template_text):
```
gfcapalbo: Checkout my regex solution on v14.  it covers the template present in any part of the text.
https://github.com/OCA/project/pull/1585/commits/32a77d63761bf9faef4b50cd337ef9b97fc494af

gfcapalbo (changes requested):
@ntsirintanis two small changes. but in general seems the way to go.
As revealed in testing , it will not cover the case of multiple addition once the user has changed the template. but that's okay, this is an improvement on current.

project_task_description_template/models/project_task.py
ntsirintanis: @gfcapalbo or instead of current.endswith(template_text), do something like current in template_text. The results should be the same exactly as the one

gfcapalbo: Works fine too. yes. Simpler as well.

alexey-pelykh (approved):
Thanks for the fix and for iterating on the feedback.

The updated code addresses both points from the previous review:
- Removed the redundant empty-template guard
- Switched from `endswith` to `template_text in current` for full substring detection

The `str()` casts are the right call here since Odoo 16 HTML fields return `markupsafe.Markup` objects and the `in` operator may not behave as expected across `Markup`/`str` boundaries.

CI is green (the codecov failures are pre-existing -- this module has no tests directory).

One minor note for awareness: if a user's own description text happens to contain the exact template string, the template won't be re-appended. This was already acknowledged as acceptable in the earlier discussion.

Looks good to merge.

PR #1609, opened 2025-11-19
[18.0][IMP] project_timesheet_time_control: Add search functionality for date_time_end and improve computation for end time based on day unit

CRogos:
When you use hr_timesheet_calendar, the calendar control searches for date_time_end. Therefore there are a lot of errors in the log.

```
 2025-11-19 19:21:33,393 4 ERROR db_name odoo.osv.expression: Non-stored field account.analytic.line.date_time_end cannot be searched. 
NoneType: None
```

### DIFF
--- a/project_timesheet_time_control/models/account_analytic_line.py
+++ b/project_timesheet_time_control/models/account_analytic_line.py
@@ -9,6 +9,7 @@ from dateutil.relativedelta import relativedelta
 
 from odoo import api, fields, models
 from odoo.exceptions import UserError
+from odoo.tools.sql import SQL
 
 
 class AccountAnalyticLine(models.Model):
@@ -22,6 +23,7 @@ class AccountAnalyticLine(models.Model):
         string="End Time",
         compute="_compute_date_time_end",
         inverse="_inverse_date_time_end",
+        search="_search_date_time_end",
     )
     show_time_control = fields.Selection(
         selection=[("resume", "Resume"), ("stop", "Stop")],
@@ -32,6 +34,7 @@ class AccountAnalyticLine(models.Model):
     @api.depends("date_time", "unit_amount", "product_uom_id")
     def _compute_date_time_end(self):
         hour_uom = self.env.ref("uom.product_uom_hour")
+        day_uom = self.env.ref("uom.product_uom_day")
         for record in self:
             if (
                 record.product_uom_id == hour_uom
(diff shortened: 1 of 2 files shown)


CRogos:
@lbarry-apsl could you also have a look at this. I think you have the same issue. Thank you.

MohamedOsman7 (approved):
Code review LGTM

CRogos:
@victoralmau could you please review?

victoralmau (approved):

project_timesheet_time_control/models/account_analytic_line.py
```diff
@@ -166,3 +178,23 @@ def button_end_work(self):
                 )
             line.unit_amount = line._duration(line.date_time, end)
         return True
+
+    @api.model
+    def _search_date_time_end(self, operator, value):
+        # reference value is 1 day == 8 hours
+        hour_uom = self.env.ref("uom.product_uom_hour")
+
```
pedrobaeza: Please don't add empty lines inside methods (being also consistent with existing code).

pedrobaeza (commented):
Is `SQL` totally needed in this case?

CRogos:
@pedrobaeza do you have another suggestion regarding the SQL?

The date_time_end depends on date_time + (duration * unit_factor). I did not find a solution to query this without SQL.
Another solution would be to make date_time_end stored, but than we need to make sure that start/end/duration is always correct.

project_timesheet_time_control/tests/test_project_timesheet_time_control.py
```diff
@@ -305,6 +305,30 @@ def test_non_timesheet_analytic_line(self):
...
+                {
+                    "date": date(2023, 1, 10),
+                    "date_time": datetime(2023, 1, 10, 8, 0, 0),
+                    "unit_amount": 2.0,
+                    "project_id": self.project.id,
+                    "name": "Test line",
+                }
+            )
+        )
+
```
pedrobaeza: Here you still have empty lines inside methods.

pedrobaeza (commented):
OK, let's go with the SQL. The alternative is to make a search with the desired elements, and then pass a domain `("id", "in", ...)`, but it can be less performant.

pedrobaeza (approved):
Thanks for attending my nitpickings :)

/ocabot merge patch

PR #1617, opened 2025-12-05
[19.0][MIG] project_template: Migration to 19.0

vangnp3702:


hieulucky111 (approved):
LGTM!

huan-tq (approved):
LGTM

Quan-nhm (approved):
Functional Test and Code review LGTM!

leemannd:
Hello @Vang-NguyenPhu could you quickly take the time to reduce a bit the administrative commits? https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

stferraro (approved):

Vang-NguyenPhu:
> Hello @Vang-NguyenPhu could you quickly take the time to reduce a bit the administrative commits? https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate


All good now.

project_template/models/project.py
```diff
@@ -0,0 +1,51 @@
...
+from odoo import api, fields, models
+
+# See _map_tasks_default_valeus
+TASK_DEFAULT_COPY_CONTEXT_KEY = f"{__name__}.task_default_copy_context_key"
+
+
+class Project(models.Model):
+    _inherit = "project.project"
+
+    is_template = fields.Boolean(copy=False)
```
alexey-pelykh, line 11: This field is already defined natively in Odoo 19.0 `project.project` model (`is_template = fields.Boolean(copy=False, export_string_translation=False)`). Re-declaring it here is redundant.

project_template/models/project.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # CREATE A PROJECT FROM A TEMPLATE AND OPEN THE NEWLY CREATED PROJECT
+    def create_project_from_template(self):
```
alexey-pelykh, line 14: Odoo 19.0 provides `action_create_from_template()` natively, which handles date shifting, field blacklisting, role-to-user mapping, and uses proper `copy_from_template` context. This simpler method bypasses all of that and produces inconsistent UX.

project_template/models/project.py
```diff
@@ -0,0 +1,51 @@
...
+        if " (TEMPLATE)" in self.name:
+            new_name = self.name.replace(" (TEMPLATE)", " (COPY)")
+        else:
+            new_name = self.name + " (COPY)"
+        new_project = self.with_context(**{TASK_DEFAULT_COPY_CONTEXT_KEY: True}).copy(
+            default={"name": new_name, "active": True, "alias_name": False}
+        )
+
+        # OPEN THE NEWLY CREATED PROJECT FORM
+        return {
```
alexey-pelykh, line 24: `view_type` is deprecated/removed in recent Odoo versions. This key is ignored in 19.0.

project_template/views/project.xml
```diff
@@ -0,0 +1,51 @@
...
+                />
+                <filter
+                    string="Non-Templates"
+                    name="projects"
+                    domain="[('is_template', '=', False)]"
+                />
+                <separator />
+            </filter>
+        </field>
+    </record>
```
alexey-pelykh, line 44: This overrides the native `open_view_project_all` and **removes** the domain `[("is_template", "=", False)]` that Odoo 19.0 sets natively. The native action filters templates via domain; this override replaces that with a search default, which is weaker (users can deselect it). This is a functional regression.

project_template/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License LGPLv3.0 or later (https://www.gnu.org/licenses/lgpl-3.0.en.html).
```
alexey-pelykh, line 1: License header says LGPLv3.0 but the manifest declares AGPL-3. Should be consistent.

alexey-pelykh (changes requested):
Thanks for the migration work.

The main issue here is that Odoo 19.0 natively includes comprehensive project template support -- `is_template` field, `action_create_from_template()`, `action_toggle_project_template_mode()`, template ribbon, dedicated template views (form/list/kanban), and template filters in search. The native `open_view_project_all` action already uses `domain=[("is_template", "=", False)]` to exclude templates from the project list.

This module as-is conflicts with the native implementation in several ways:

1. **Redundant `is_template` field**: Odoo 19.0 already defines `is_template = fields.Boolean(copy=False)` on `project.project`. Re-declaring it in this module is harmless but pointless.

2. **Conflicting action window override**: The module overrides `project.open_view_project_all` to set `search_default_projects: True`, but the native action already uses a hard `domain=[("is_template", "=", False)]`. This override actually *removes* that domain filter and replaces it with a weaker search default that users can deselect, which would then show templates mixed in with projects. This is a regression from the native behavior.

3. **Simpler `create_project_from_template` vs native `action_create_from_template`**: The native method handles date shifting, field blacklisting, role-to-user mapping, and proper copy context. This module's method bypasses all of that.

4. **`on_change_is_template` name mangling**: The native 19.0 doesn't append "(TEMPLATE)" to the project name -- it uses a ribbon widget instead.

Given that Odoo 19.0 has native template support, this module either needs to be:
- **Deprecated/removed** with a note that functionality is now native, or
- **Significantly reworked** to only add value *on top of* the native implementation

The task `date_end` preservation via `copy_data`/`update_date_end` overrides might still add value, but the rest conflicts with or duplicates native functionality.

Minor: license header mismatch in `__init__.py` and `__manifest__.py` (copyright header says LGPLv3.0, manifest `license` field says AGPL-3).

stferraro:
Is it necessary if this functionality already exists in Odoo 19?

Gerloav (approved):

alexey-pelykh (commented):
One correction to finding #2 (the action-window override): I said the module "removes the native `domain=[('is_template','=',False)]` filter" — that's not right. The override only sets `<field name="context">`, and Odoo's `<record>` update semantics preserve fields that aren't re-specified, so the native hard domain still applies and templates stay excluded. Not a regression — better stated as "the `search_default_projects` context is redundant." The change request itself stands on findings #1/#3/#4.

Co-Reviewed-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

moylop260:
[project_template](https://github.com/OCA/project/tree/18.0/project_template) was already implemented by Odoo core:
 - https://github.com/odoo/odoo/pull/209034

Check the following [project_template doc](https://www.odoo.com/documentation/19.0/applications/services/project/project_management/project_templates.html)

Feel free to re-open if you think it is not enough implemented

PR #1626, opened 2025-12-12
[19.0] [ADD] project_task_product, project_task_product_stock

MarinaAForgeFlow:


project_task_product/views/project_task_views.xml
```diff
@@ -0,0 +1,17 @@
...
+    <record id="view_task_form2_inherit" model="ir.ui.view">
+        <field name="name">view.task.form2.inherit</field>
+        <field name="model">project.task</field>
+        <field name="inherit_id" ref="project.view_task_form2" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='user_ids']" position="after">
+                <field name="product_id" />
+                <field
+                    name="lot_id"
+                    invisible="not product_id or product_id.tracking == 'none'"
```
AaronHForgeFlow:
```suggestion
                    invisible="not product_id or product_id.tracking == 'none'"
                    groups="stock.group_production_lot"
```

AaronHForgeFlow (commented):
LGTM Functional + code, just minor a comment

project_task_product/views/project_task_views.xml
MarinaAForgeFlow: Thanks!

AaronHForgeFlow (approved):
Thanks

CRogos:
Interesting... could we split this into two modules?
stock_product and project_task_product
We are using project, but we are not happy with stock, because we have no shipping/production business.

MarinaAForgeFlow:
> Interesting... could we split this into two modules? stock_product and project_task_product We are using project, but we are not happy with stock, because we have no shipping/production business.

@CRogos  Just to make sure I get it. You mean moving the "stock" dependency into a separate module. So we have project_task_product that only adds the product to the task, and then something like project_task_stock_product (or something along those lines) that adds the stock dependency and the lot number, right?

CRogos:
@MarinaAForgeFlow yes exactly.

CRogos (commented):
Code LGTM, but I am not sure if this module fits well in the Odoo system. 

When you install sale_project, each task get a sale_line_id, and a sale_line_id has a product_id itself. 
- The task.product_id is not prefilled with the sale_line_id.product, when a task is created from a sale order.
- There could be a mismatch between the task.sale_line_id.product_id and the task.product_id.

MarinaAForgeFlow:
> Code LGTM, but I am not sure if this module fits well in the Odoo system.
> 
> When you install sale_project, each task get a sale_line_id, and a sale_line_id has a product_id itself.
> 
> * The task.product_id is not prefilled with the sale_line_id.product, when a task is created from a sale order.
> * There could be a mismatch between the task.sale_line_id.product_id and the task.product_id.

I see your concern. However, if I am not wrong, this could be solved with a glue module sale_project_task_product, where we ensure that the product on the project task matches the sale line product. 

Our idea for this module is simply to allow users to define to which product the task relates to, for visibility purposes.

project_task_product/models/product.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2025 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
```
alexey-pelykh, line 8: The `_compute_task_count` here runs `search_count` per record (N+1). Consider `_read_group` with `groupby=['product_id']` for a single query. Same applies to `ProductTemplate._compute_task_count` below.

project_task_product_stock/models/project_task.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    lot_id = fields.Many2one(
+        "stock.lot",
+        string="Lot/Serial Number",
+        domain="[('product_id', '=', product_id)]",
+        tracking=True,
```
alexey-pelykh, line 13: Consider adding `check_company=True` here for consistency with `product_id`.

project_task_product_stock/models/project_task.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2025 ForgeFlow S.L. (https://www.forgeflow.com)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    lot_id = fields.Many2one(
```
alexey-pelykh, line 10: When `product_id` changes, `lot_id` should be reset to avoid referencing a lot from the previous product. An `@api.onchange('product_id')` that clears `lot_id` (or a compute/inverse pattern) would handle this.

project_task_product/models/project_task.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ProjectTask(models.Model):
+    _inherit = "project.task"
+
+    product_id = fields.Many2one(
+        "product.product", string="Product", check_company=True
```
alexey-pelykh: Minor: consider adding `index=True` since this field is used in search domains across both modules.

alexey-pelykh (commented):
Thanks for the PR and for the clean split into two modules -- that addresses @CRogos' feedback well.

The code is functionally correct and the test coverage is solid. I have a couple of observations, mostly around performance and a potential data consistency gap:

**1. `_compute_task_count` -- N+1 queries**

All three `_compute_task_count` methods (`product.product`, `product.template`, `stock.lot`) execute `search_count` inside a `for` loop. When viewing a product list or any form with multiple records in the recordset, this causes one SQL query per record.

Consider using `_read_group` to batch into a single query, e.g. for `product.product`:

```python
def _compute_task_count(self):
    task_data = self.env["project.task"]._read_group(
        [("product_id", "in", self.ids)],
        groupby=["product_id"],
        aggregates=["__count"],
    )
    mapped = {product.id: count for product, count in task_data}
    for product in self:
        product.task_count = mapped.get(product.id, 0)
```

Same pattern would apply to the template and lot computes. Not blocking, but worth improving since stat buttons can trigger these on list views.

**2. `lot_id` not cleared on product change**

If the user changes `product_id` on a task, the `lot_id` still points to the old product's lot. The domain in the view prevents selecting an invalid lot, but the old value remains. An `onchange` (or `@api.depends` / `compute` with `store=True` and `readonly=False` pattern) to reset `lot_id` when `product_id` changes would prevent stale data.

**3. Minor: `index=True` on `product_id`**

Since `product_id` is used in `search_count` domains, `domain` filters in actions, and the stock module's lot domain -- adding `index=True` on the `product_id` field would help query performance as the dataset grows.

project_task_product_stock/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+{
+    "name": "Project Task Product Stock",
+    "summary": "Allows to specify in a project task, the lot or serial number of"
+    " the task product.",
+    "version": "19.0.1.0.0",
+    "category": "Project",
+    "website": "https://github.com/OCA/project",
+    "author": "ForgeFlow, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "depends": ["project_task_product", "stock"],
```
MiquelRForgeFlow:
```suggestion
    "depends": ["project_task_product", "project_stock"],
```

MiquelRForgeFlow (approved):

DavidJForgeFlow (approved):

PR #1631, opened 2025-12-19
[16.0][FIX] project_stock: analytic_line_from_task values for returned moves

Tisho99:
When a stock move linked to a task is returned, the generated analytic line currently keeps the same amount as the original move’s analytic line.

This PR updates the behavior so that the amount is created with the opposite sign, correctly reflecting the return.

### DIFF
--- a/project_stock/models/stock_move.py
+++ b/project_stock/models/stock_move.py
@@ -1,5 +1,6 @@
 # Copyright 2022-2025 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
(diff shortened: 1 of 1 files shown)


Jaimermaccione (approved):
@Tisho99 LGTM. Reviewed on local environment.

Tests performed:
1. Material consumption
2. Partial and full returns
3. Mixed products with returns

All scenarios behaved as expected: correct signs, totals, pricing, and full analytic/stock traceability.

victoralmau (approved):
It seems to be coherent

pedrobaeza:
Please cherry-pick it to upper versions.

PR #1633, opened 2025-12-23
[16.0][IMP] project_stock: add unit test with a returned move

Tisho99:
Addition of a unit test for https://github.com/OCA/project/pull/1631

After this, i'll forward port both commits

PR #1638, opened 2026-01-02
[18.0][FIX] project_task_stock: analytic_line_from_task values for returned moves

victoralmau:
FWP from 16.0: https://github.com/OCA/project/pull/1631 + https://github.com/OCA/project/pull/1633

When a stock move linked to a task is returned, the generated analytic line currently keeps the same amount as the original move’s analytic line.

This PR updates the behavior so that the amount is created with the opposite sign, correctly reflecting the return.

@Tecnativa


david-banon-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1642, opened 2026-01-05
[17.0][ADD] project_kanban_form_direct_access

IriaAlonso:
Currently, to access the project form view, users must click on "three dots>Settings". 
With this improvement, users only need to click on the project name.


IriaAlonso:
@anddago78 @yajo, can you review it please?

Saran440 (commented):
I think this module is useful.
But should we add clickable on title? now it shows duplicate name

Image

project_kanban_form_direct_access/static/description/icon.png
yajo, line 1: issue: add a specific icon or use standard OCA icon, but not your branded icon.

yajo (changes requested):
Code review. Just one little issue.

luisDIXMIT (commented):
Functionally it works, but I agree with Saran, is there a way to avoid duplicating the title?

IriaAlonso:
The project name is duplicated because `project_sequence` hide the original name and adds the name again https://github.com/OCA/project/blob/17.0/project_sequence/views/project_project.xml#L52. 
Perhaps a fix should be made to avoid this?

Saran440:
Okay, I see. Maybe we use button action instead? what do you think?

Saran440 (approved):
Functional Test 👍 
It's duplicate from module `project_sequence`

alexey-pelykh (approved):
Thanks! This is the 17.0 counterpart of the 18.0 version I reviewed recently. View-only module, clean and simple.

LGTM.

yajo (changes requested):
issue: Project names still duplicate:

Image

PR #1341, opened 2024-09-26
[16.0][ADD] Addon: project_required_field_by_stage

Tiago370:
This PR add the module `project_required_field_by_stage` that allow to define required fields depending on the stage of the task.

NOTE: I was unable to find a satisfactory way to cover lines 30-33 and 35-37 of the models/project_task.py file with tests, I accept suggestions.

PR #1650, opened 2026-01-09
[16.0] [ADD] project_required_field_by_stage: New module

gfcapalbo:
reopening https://github.com/OCA/project/pull/1341 with original commits.
Added changes requested in that original MR-.


ddejong-therp:
Code LGTM.
I've tested it, and run the tests as well.
All looks good.

ddejong-therp (approved):

UrzaBlock:
Tested its functionality on runboat, and it does what it's supposed to do: When you have added fields in the Required Field section of a stage, a task that gets put in that stage will have those fields highlighted as required.

project_required_field_by_stage/models/project_task.py
```diff
@@ -0,0 +1,77 @@
...
+        key = super()._get_view_cache_key(view_id, view_type, **options)
+        return key + tuple(
+            self.env["project.task.type"]
+            .search([("required_field_ids", "!=", False)])
+            .mapped("required_field_ids.name")
+        )
+
+    @api.constrains("stage_id")
+    def _check_stage_id_(self):
+        for this in self:
```
NL66278, line 72: Maybe I miss something, but I do not see why the inner loops are necessary. A project can only be in one stage, so why would this be wrong?
```
    @api.constrains("stage_id")
    def _check_stage_id_(self):
        for this in self:
            required_fields = this.stage_id.required_field_ids
            if not required_fields:
                continue
            for field in required_fields:
                if hasattr(this, field.name) and not getattr(this, field.name):
                    raise UserError(
                        _(  
                            "Field '%(field)s' is mandatory in stage '%(stage)s'."
                        )   
                        % ( 
                            {   
                                "field": field.display_name.split(" (")[0],
                                "stage": this.stage_id.display_name,
                            }   
                        )   
                    ) 
```

ddejong-therp: Ah yes I agree, I see that `stage_id` is a `Many2one` field

gfcapalbo: Every record has one stage.  we only need to check mandatory fields for that one stage, not all stages every time. thank you @NL66278 , i  will simplify this , by removing the loop.

project_required_field_by_stage/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2024 KMEE
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Project Required Field By Stage",
+    "summary": """
+        KMEE""",
```
luisDIXMIT: Can you add a more detailed summary, please?

luisDIXMIT (commented):
When I remove required fields from a stage, it still requires those fields.

luisDIXMIT:
PR tittle must be "[16.0] [ADD] project_required_field_by_stage: New module"

project_required_field_by_stage/models/project_task.py
gfcapalbo: I re-reviewed this. @NL66278 

stages = this.stage_id.filtered(lambda x: x.required_field_ids)           makes no sense. and some other filtering makes no sense either.
  
All we want in this costrains is to verify that on stage change all mandatory fields specified for that stage  are there.

Added a fix commit, please re-check.

project_required_field_by_stage/__manifest__.py
gfcapalbo: Done. thank you for reviewing. if module is satisfactory please leave an approval.

project_required_field_by_stage/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,3 @@
+* KMEE (https://kmee.com.br/):
+
+  * Tiago Amaral
```
NL66278, line 3: Enough enhancements to add yourself to contributors list

project_required_field_by_stage/models/project_task.py
```diff
@@ -0,0 +1,73 @@
...
+                            attrs["required"] = [
+                                "|",
+                                ("stage_id", "in", stages_with_field.ids),
+                            ] + attrs["required"]
+                        else:
+                            attrs["required"] = [
+                                ("stage_id", "in", stages_with_field.ids)
+                            ]
+                    else:
+                        attrs["required"] = [("stage_id", "in", stages_with_field.ids)]
```
NL66278: All this can be done much more concise (and without repeating three times the same condition added):
` ``
from odoo.osv.expression import OR  # Obviously this at the top of the python file
...
required_domain = attrs.get("required", [])
attrs["required"] = OR([required_domain, [("stage_id", "in", stages_with_field.ids)]])
```
No `if` or `else` needed!

File: project_required_field_by_stage/__manifest__.py [hunk 5]
```diff
@@ -0,0 +1,19 @@
...
+
+{
+    "name": "Project Required Field By Stage",
+    "summary": """
+        This module adds checks to allow certain stages to be set only if
+        some fields are populated. After install every stage can have
+        mandatory fields associated.""",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "KMEE,Odoo Community Association (OCA)",
```
NL66278, line 12: There is as yet no maintainer defined. Maybe you can add yourself as maintainer? This will give the possibility to merge stuff, even if you are not a PSC member for project and you will be notified of PR's. Of course this brings responsibility as well.

project_required_field_by_stage/models/project_task.py
```diff
@@ -0,0 +1,73 @@
+# Copyright 2024 KMEE
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import ast
+import json as simplejson
```
alexey-pelykh: `import json as simplejson` is a misleading alias (`simplejson` was a different library). Just use `import json` and call `json.dumps` below.

project_required_field_by_stage/models/project_task.py
```diff
@@ -0,0 +1,73 @@
...
+            [("required_field_ids", "!=", False)]
+        )
+        if view.type == "form" and stages:
+            for field in stages.mapped("required_field_ids"):
+                stages_with_field = stages.filtered(
+                    lambda stage, field=field: field in stage.required_field_ids
+                )
+                for node in arch.xpath("//field[@name='%s']" % field.name):
+                    attrs = ast.literal_eval(node.attrib.get("attrs", "{}"))
+                    if attrs:
```
alexey-pelykh: As @NL66278 suggested -- this whole `if/elif/else` block can be replaced with:
```python
from odoo.osv.expression import OR
...
required_domain = attrs.get("required", [])
attrs["required"] = OR([required_domain, [("stage_id", "in", stages_with_field.ids)]])
```
No branching needed.

project_required_field_by_stage/models/project_task.py
```diff
@@ -0,0 +1,73 @@
...
+        key = super()._get_view_cache_key(view_id, view_type, **options)
+        return key + tuple(
+            self.env["project.task.type"]
+            .search([("required_field_ids", "!=", False)])
+            .mapped("required_field_ids.name")
+        )
+
+    @api.constrains("stage_id")
+    def _check_stage_id_(self):
+        for this in self:
```
alexey-pelykh, line 72: Trailing underscore on `_check_stage_id_` looks like a typo. Consider `_check_required_fields_by_stage` which also better describes the intent.

project_required_field_by_stage/models/project_task.py
```diff
@@ -0,0 +1,73 @@
...
+            if this.stage_id.required_field_ids:
+                fields = (
```
alexey-pelykh: This search is redundant -- `this.stage_id.required_field_ids` is already the `ir.model.fields` recordset you want. The `.sudo().search(...)` just re-fetches the same records. Replace the whole block with:
```python
for field in this.stage_id.required_field_ids:
```

project_required_field_by_stage/models/project_task.py
```diff
@@ -0,0 +1,73 @@
...
+                    this.env["ir.model.fields"]
+                    .sudo()
+                    .search([("id", "in", this.stage_id.required_field_ids.ids)])
+                )
+                for field in fields:
+                    if hasattr(this, field.name) and not getattr(this, field.name):
```
alexey-pelykh: `not getattr(this, field.name)` is falsy for `0` and `0.0`, so a numeric field set to zero would incorrectly fail validation. For scalar fields, Odoo uses `False` to mean "not set" -- check `getattr(this, field.name) is False` for non-relational types, or inspect `field.ttype` to pick the right emptiness check.

project_required_field_by_stage/models/project_task.py
```diff
@@ -0,0 +1,73 @@
...
+    @api.model
+    def _get_view_cache_key(self, view_id=None, view_type="form", **options):
+        """The override of _get_view changing the required fields labels according
+        to the stage makes the view cache dependent on the stages with required fields."""
+        key = super()._get_view_cache_key(view_id, view_type, **options)
+        return key + tuple(
+            self.env["project.task.type"]
+            .search([("required_field_ids", "!=", False)])
+            .mapped("required_field_ids.name")
+        )
```
alexey-pelykh, line 68: Cache key only includes field names, not which stages they belong to. Removing a required field from one stage won't invalidate the cache if another stage still requires the same field name. Include stage IDs or `write_date` in the key to ensure proper invalidation on config changes.

alexey-pelykh (changes requested):
Thanks for reviving this from #1341 and for the work on tests and the view cache key -- this is a useful module. A few things need attention before merge:

**1. `_get_view` attrs logic -- apply NL66278's suggestion**

As @NL66278 pointed out, the three-branch `if/elif/else` for building the required domain can be collapsed using `odoo.osv.expression.OR`:

```python
from odoo.osv.expression import OR
...
required_domain = attrs.get("required", [])
attrs["required"] = OR([required_domain, [("stage_id", "in", stages_with_field.ids)]])
```

This removes all the branching and is the idiomatic Odoo way to combine domains.

**2. `_check_stage_id_` -- redundant search**

`this.stage_id.required_field_ids` is already a recordset of `ir.model.fields`. The `.sudo().search(["id", "in", ...ids])` re-queries the same records for no reason. Simplify to:

```python
for field in this.stage_id.required_field_ids:
```

The `.sudo()` is also unnecessary here since you already have access to the stage through the task.

**3. Falsy check fails on numeric zero**

`not getattr(this, field.name)` treats `0` and `0.0` as empty. If someone marks an integer or float field as required and sets it to `0`, the constraint will incorrectly reject it. Consider checking the field type: for relational fields check `not value`, for others check `value is False` (which is how Odoo represents "not set" for scalar fields).

**4. View cache invalidation bug**

@luisDIXMIT reported that removing required fields from a stage still enforces them. The `_get_view_cache_key` only includes field names but not which stages they belong to. If stage A requires `user_ids` and you remove it, but stage B also requires `user_ids`, the cache key is unchanged and the stale view is served. Include stage IDs in the cache key.

**5. Minor items**

- `import json as simplejson` -- misleading alias, just use `import json` and `json.dumps`.
- Method name `_check_stage_id_` has a trailing underscore -- likely a typo. Rename to `_check_required_fields_by_stage` for clarity.
- Missing `"installable": True` in manifest (OCA convention).

Overall the concept is solid and the test coverage is decent. Looking forward to the next iteration.

project_required_field_by_stage/models/project_task.py
gfcapalbo: yes attrs , previously in  node.attrib.get defaults to {} so it is safe to remove all branches.

project_required_field_by_stage/models/project_task.py
gfcapalbo: Managed this in refactored is_empty function for clarity, where i check the ttype and do the appropriate check.

gfcapalbo:
@alexey-pelykh @NL66278   all review comments and simplifications implemented specifically in commit 4f16056f8694c481ac92a1335a773c0a9937b35e .  

Added a function that correctly evaluates what "not set" means for the different types of fields.  Added tests to cover those code branches  too and keep codeCov coverage percentage from dipping.
I made a choice:  Adding booleans is not a noop operation,  a boolean is always "set" ( true or false),    so i semantically interpret  booleanfield = false as "not set"  and true as "set"   i think  is the right interpretation .

if i  had made   {boolean_field} = true or false   = set    that would had made adding booleans to restrictions on stages effectively useless.

alexey-pelykh (approved):
All review comments from the previous round have been addressed in commit 4f16056f:

1. **`import json as simplejson`** -- replaced with clean `import json`.
2. **`_get_view` attrs logic** -- collapsed to `OR([required_domain, ...])` as suggested by @NL66278.
3. **Redundant search removed** -- constraint now iterates `this.stage_id.required_field_ids` directly.
4. **Method renamed** -- `_check_stage_id_` is now `_check_required_fields_by_stage`.
5. **Type-aware emptiness check** -- new `_is_empty()` function correctly differentiates by `field.ttype`. The boolean interpretation (`False` = "not set") is reasonable for this use case.
6. **Cache key fix** -- `_get_view_cache_key` now includes `(stage.id, field.name)` tuples, so config changes properly invalidate the cached view.
7. **`"installable": True`** added to manifest.

Tests cover the constraint, view injection for multiple field types, and the `OR` domain merging. Looks good.

project_required_field_by_stage/tests/test_project_required_field_by_stage.py
```diff
@@ -0,0 +1,146 @@
...
+                }
+            )
+        self.project_task_1.write(
+            {
+                "user_ids": [(4, self.res_users_1.id)],
+            }
+        )
+        self.project_task_1.write(
+            {
+                "stage_id": self.project_task_type_2.id,
```
NL66278, line 92: This write can be combined with the previous write, which actually would make the test stronger, as writing the required field for a stage together with the stage should succeed.

project_required_field_by_stage/tests/test_project_required_field_by_stage.py
```diff
@@ -0,0 +1,146 @@
...
+            {
+                "name": "Project 1",
+            }
+        )
+        self.project_task_type_1 = self.project_task_type_model.create(
+            {
+                "name": "Project Stage 1",
+                "project_ids": [(4, self.project_project_1.id)],
+            }
+        )
```
NL66278, line 29: Just a suggestion for more concise code:
```
from odoo.fields import Command
...
required_field_names = [
    "planned_hours",  # integer
    "project_id",  # m20
    "user_ids",  # m2m
    "active",  # bool
]
....
"required_fields": [
    Command.link(self.env.ref("project.field_project_task__%s" % field_name).id)
    for field_name in required_field_names
]
```

NL66278:
@gfcapalbo Please squash your own commits into one.

leemannd:
@NL66278 I missed your comment before starting the merge of the module : /

PR #1651, opened 2026-01-11
[18.0][MIG] project_task_stage_change_restriction: migration from 16.0

Mohajiro:
Standard migration from 16.0 to 18.0.


project_task_stage_change_restriction/models/project_task.py
```diff
@@ -43,7 +43,7 @@ def _check_stage_restriction(self, vals):
             return True
 
         new_stage = self.env["project.task.type"].browse(stage_id)
-        if not new_stage:
+        if not new_stage:  # pragma: no cover
```
ivs-cetmix: Why?

ivs-cetmix (changes requested):

project_task_stage_change_restriction/models/project_task.py
Mohajiro: The # pragma: no cover is added only to satisfy codecov, without changing functional logic or test semantics.

ivs-cetmix: You must add tests if needed, not silence the test coverage tool.

ivs-cetmix (changes requested):

project_task_stage_change_restriction/models/project_task.py
Mohajiro: This line is practically untestable because `browse()` always returns a recordset (even for non-existent IDs). The check `if not new_stage:` will never execute in real scenarios since invalid IDs are handled earlier by the ORM. 
Marked # pragma: no cover only to avoid false codecov failure, not to hide testable logic.

ivs-cetmix: Then you need to check if the stage record exists. Otherwise this check makes no sense.

ivs-cetmix (changes requested):

project_task_stage_change_restriction/models/project_task.py
Mohajiro: This case is not reachable in real usage: invalid stage IDs are handled by the ORM before this check is relevant. The line is kept as a defensive guard and excluded from coverage to avoid a false codecov failure.

ivs-cetmix: Then we need to remove this code completely.

ivs-cetmix (changes requested):

ivs-cetmix (approved):
Code review LGTM

Aldeigja (approved):
Functional LGTM

francesco-ooops (approved):
Functional is ok!

ivs-cetmix:
Hey @Mohajiro thank you for your contribution!
/ocabot merge nobump

PR #1101, opened 2023-04-06
[14.0][ADD] project_sequence

anddago78:
Add a sequence field to projects, filled automatically and add a code sequence filter in tree view project.

MT-1506 @moduon @rafaelbn @Shide @yajo please review :)

PR #1658, opened 2026-01-20
[IMP][16.0] project_sequence: set analytic's `code` as per project's `sequence_code`

arnaudlayec:
Before, the project's `display_name` was synched in the project's analytic account (aac) name. This way, the "code" field of aac was empty, and the name had both the project's code and the name in the default format "[code] - [name]" Proposal : synch the 2 fields ("sequence_code" and "name") of the project respectively in the 2 fields ("code" and "name") of the aac

This allow:
- standard display of aac (see aac's `display_name` method)
- ability to target aac with according project's code. Can be useful when importing account move and needing to set an `analytic_distribution` using the aac's code

This PR also update tests.
I needed to fix some:
- `default_plan_id` must be computed once to exist
- the 2 following tests cannot coexist. I aligned the code & tests like in v18.0 : `test_project_without_sequence` and `test_project_with_empty_sequence`. Doing that required to change project's `create` method

Note: if OK, I can port to upper version.


project_sequence/tests/test_project_sequence.py
```diff
@@ -118,35 +111,16 @@ def test_sequence_unique(self):
...
-        )
-        self.assertEqual(proj1.sequence_code, "23-00011")
-        self.assertEqual(proj1.display_name, "23-00011 - whatever")
-        # Sequence is applied when creating project with sequence in False
-        proj2 = self.env["project.project"].create(
-            {"name": "whatever", "sequence_code": False}
-        )
-        self.assertEqual(proj2.sequence_code, "23-00012")
-        self.assertEqual(proj2.display_name, "23-00012 - whatever")
-
```
yajo, line 149: question: Why do you need to remove this test? Is it not passing anymore?

yajo (commented):
Please could you share screenshots or screencast to see how it was before and how it is now? I think that will make it easier to review.

Code looks OK. Just one doubt.

project_sequence/tests/test_project_sequence.py
arnaudlayec: Actually the other test `test_project_without_sequence` was not passing, because in a test environment, there is not previous project created. I fixed this test like it is fixed in v18.0.
This required this slight change in `create` method:
```python
if not vals.get("sequence_code", False): # removed
if "sequence_code" not in vals: # added
```

Which make this test `test_project_with_empty_sequence` fail.
==> So I align the test and code for this with v18.0's.

yajo: I understand. The correct way to do this would be to fwport https://github.com/OCA/project/pull/1101/changes/6ac7e102a99f65ef3bf209807723194ecf408b71 in a separate commit. Could you please do that?

arnaudlayec:
> Please could you share screenshots or screencast to see how it was before and how it is now? I think that will make it easier to review.
> 
> Code looks OK. Just one doubt.

Thank you for your review!
Yes indeed, a quick Screencast here (00:32), with `hr_timesheet` installed so project's analytic account is created right at project creation.

https://github.com/user-attachments/assets/8966c62b-13f6-469c-bc77-a1dffa155419

Se how Analytic Account's field "code" is filled in, as per project's sequence code:
image

project_sequence/tests/test_project_sequence.py
arnaudlayec: @yajo is it ok like this? I splitted in 3 commits: 1 for my proposal, 2 for the tests update.

yajo: Thanks! Much easier to follow and fwport now 😊

project_sequence/tests/test_project_sequence.py
```diff
@@ -20,18 +20,7 @@ def setUpClass(cls):
...
-            .search(
-                [
-                    "|",
-                    ("company_id", "=", False),
-                    ("company_id", "=", cls.env.company.id),
-                ],
-                limit=1,
-            )
-        )
+        default_plan_id = cls.env.company.analytic_plan_id
```
yajo, line 23: question: regarding this change, I see a behavioral change. Before, an AAP without company would be considered as a default plan. Now, only the one from the company will be considered.

Could you please explain why this change?

Thanks!

luisDIXMIT (approved):
Code review, LGTM!

arnaudlayec:
> Please could you share screenshots or screencast to see how it was before and how it is now? I think that will make it easier to review.
> 
> Code looks OK. Just one doubt.

Hello @yajo , would you like to review this PR?

alexey-pelykh (commented):
Thanks for this improvement, @arnaudlayec. Syncing the project's `name` and `sequence_code` separately into the analytic account's `name` and `code` fields is the right approach -- it lets the AAC's own `display_name` handle formatting and makes code-based lookups possible (e.g. for analytic distribution in imports).

Code is clean and CI is green. One small gap I noticed:

In `test_sync_analytic_account`, you now assert `proj.analytic_account_id.name == proj.name`, but since the method now also writes `code`, it would be good to assert `proj.analytic_account_id.code == proj.sequence_code` as well -- for consistency with the other tests that already do this.

Also a note on the `create` condition change (`not vals.get("sequence_code", False)` -> `"sequence_code" not in vals`): this means passing `sequence_code=""` will no longer auto-assign a sequence. I see this aligns with v18.0, just flagging it as a behavioral change for other reviewers.

Otherwise LGTM.

project_sequence/tests/test_project_sequence.py
```diff
@@ -195,12 +158,12 @@ def test_sync_analytic_account_name(self):
             }
         )
         proj.analytic_account_id = analytic_account
-        proj._sync_analytic_account_name()
-        self.assertEqual(proj.analytic_account_id.name, proj.display_name)
+        proj._sync_analytic_account()
```
alexey-pelykh, line 161: The method now syncs both `name` and `code`, but this test only asserts `name`. Consider adding:
```python
self.assertEqual(proj.analytic_account_id.code, proj.sequence_code)
```
The other test methods (`test_analytic_account_after_creation_named`, `test_sequence_copied_to_name_if_emptied`) already verify both fields.

alexey-pelykh, line 161: suggestion (non-blocking): Since `_sync_analytic_account()` now writes both `name` and `code`, consider adding:
```python
self.assertEqual(proj.analytic_account_id.code, proj.sequence_code)
```
The other test methods (`test_analytic_account_after_creation_named`, `test_sequence_copied_to_name_if_emptied`) already verify both fields -- this would keep coverage consistent.

alexey-pelykh (approved):
Good improvement. Syncing `sequence_code` into the analytic account's `code` field separately from `name` is the right call -- it leverages the AAC's own `display_name` formatting and enables code-based lookups (useful for import scenarios with `analytic_distribution`).

CI is green, tests are updated and cover the new behavior well. The `create` condition change (`not vals.get(...)` to `"sequence_code" not in vals`) is a deliberate v18.0 alignment and has been discussed in the review thread.

One minor suggestion below (non-blocking).

project_sequence/tests/test_project_sequence.py
arnaudlayec: Thanks for review, indeed the analytic's code should be tested here too, it's now added

project_sequence/tests/test_project_sequence.py
arnaudlayec: Hello Yajo, sorry for responding so late

There is not intended behavior change here, this is only to fix the tests with the native data.
From what I see in v16.0, the default project plan is created per company only, and through a compute method : https://github.com/odoo/odoo/blob/16.0/addons/project/models/company.py .

Calling it like this `cls.env.company.analytic_plan_id` allows the compute method to be triggered and return something.
The previous way with a `search` seems not to trigger the compute and thus returning a void default analytic plan.

I agree this change only consider a default analytic plan from a company, but this is only to fix the tests.

PR #1660, opened 2026-01-22
[19.0][ADD] project_profitability_threshold_alert: Notify project followers and project manager about costs exceeding based on a set threshold

nihelgabsi-acsone:
Notify followers and project manager about costs exceeding based on a set threshold


project_profitability_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,104 @@
...
+
+from datetime import datetime, timedelta
+
+from odoo import api, fields, models
+
+
+class ProjectProject(models.Model):
+    _inherit = "project.project"
+
+    costs_threshold = fields.Float()
```
rousseldenis: You should add a help to explain this.

rousseldenis: Moreover, the default value coming from settings is not set at project creation.

A good idea should be to change to a computed field (store=True/readonly=False) to get that default value.

project_profitability_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,104 @@
...
+    is_cost_exceeded = fields.Boolean(
+        compute="_compute_is_cost_exceeded", default=False, store=True
+    )
+    force_notification_send = fields.Boolean(
```
rousseldenis: For those three fields here, add 'cost_threshold' in their names to avoid naming conflict with possible other fields we are not aware here

project_profitability_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,104 @@
...
+        string="Force mail notification send", default=False
+    )
+    create_activity = fields.Boolean(default=False)
+    is_notfication_sent = fields.Boolean(default=False)
+
+    def _cron_find_costs_threshold_exceeded(self):
+        projects = self.env["project.project"].search([("active", "=", True)])
```
rousseldenis:
```suggestion
        projects = self.env["project.project"].search([])
```

Inactive projects are not returned by `search()` by default.

project_profitability_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,104 @@
...
+        for project in projects:
+            project._compute_is_cost_exceeded()
```
rousseldenis: Is this really necessary ? The compute is triggered by its depends.

project_profitability_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,104 @@
...
+            if project.is_cost_exceeded and not project.is_notfication_sent:
```
rousseldenis: Those fields should be included in search criteria (add a dedicated method for the domain like `_get_costs_threshold_notification_domain()`)

project_profitability_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,111 @@
...
+        self.message_notify(
+            partner_ids=users_to_notify.filtered(
+                lambda u: u.receive_project_threshold_notification
+            ).partner_id.ids,
+            body=f"Cost threshold exceeded for project {self.name}!",
+            subject="Project Cost Threshold Exceeded",
+            email_layout_xmlid=None,
+        )
+
+    @api.depends("costs_threshold", "company_id.project_costs_threshold")
```
rousseldenis: So, no need of `company_id.project_cost_threshold`

project_profitability_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,111 @@
...
+            costs_threshold = (
+                project.costs_threshold or project.company_id.project_costs_threshold
+            )
+            costs_exceeded = (total_costs_expected * -1) > (
+                total_revenues_expected * costs_threshold
+            )
+            project.is_cost_exceeded = costs_exceeded
+
+    def _post_message_to_partners(self, partners):
+        tpl = "project_profitability_threshold_alert.project_costs_exceeded_template"
```
rousseldenis: Prefer explicit names. Clearer for external readers

```suggestion
        template = "project_profitability_threshold_alert.project_costs_exceeded_template"
```

project_profitability_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,111 @@
...
+        if manager:
+            today = datetime.today()
+            self.env["mail.activity"].create(
+                {
+                    "res_model_id": self.env["ir.model"]._get_id("project.project"),
+                    "res_id": self.id,
+                    "user_id": manager.id,
+                    "activity_type_id": self.env.ref("mail.mail_activity_data_todo").id,
+                    "summary": "Project Cost Threshold Exceeded",
+                    "date_deadline": today + timedelta(days=4),
```
rousseldenis: Why hardcoded 4 days ?

It should be configurable

project_profitability_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,111 @@
...
+    def _get_internal_users(self):
+        manager = self.user_id
+        internal_user_followers = self.message_partner_ids.user_ids.filtered(
+            lambda x: not x.share
+        )
+        if manager and manager not in internal_user_followers:
+            internal_user_followers += manager
+        return internal_user_followers
+
+    @api.onchange("costs_threshold")
```
rousseldenis: Set `is_notification_sent` as computed / store=True depending on `costs_threshold` to avoid onchanges.

project_profitability_threshold_alert/models/res_config_settings.py
```diff
@@ -0,0 +1,19 @@
...
+    _inherit = "res.config.settings"
+
+    project_costs_threshold = fields.Float(
+        related="company_id.project_costs_threshold", readonly=False
+    )
+
+    @api.onchange("project_costs_threshold")
+    def _reset_is_notfication_sent(self):
+        related_projects = self.env["project.project"].search(
+            [("costs_threshold", "=", False)]
```
rousseldenis: Add `is_notification_sent = True`

rousseldenis (changes requested):

project_profitability_threshold_alert/models/project_project.py
nihelgabsi-acsone: I need to execute _get_profitability_items() to get updated data related to project costs and revenues

project_minimum_margin_threshold_alert/data/mail_template.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2026 ACSONE SA/NV
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="project_minimum_margin_exceeded_template" model="mail.template">
+        <field name="name">Project: Minimum Margin threashold exceeded</field>
```
alexey-pelykh, line 6: Typo: `threashold` → `threshold` (in the template name and subject line)

alexey-pelykh (commented):
Nice module, the margin threshold alert with cron + activity + email is well thought out. Small typo: "Threashold" should be "Threshold" in the mail template subject and body.

project_minimum_margin_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,157 @@
...
+        "that triggers an alert",
+        default=lambda self: self._default_margin_threshold,
+        tracking=True,
+    )
+    is_margin_threshold_exceeded = fields.Boolean(
+        help="This is set if margin is exceeded from threshold",
+        tracking=True,
+    )
+    is_margin_threshold_exceeded_notfication_sent = fields.Boolean(
+        compute="_compute_is_margin_threshold_exceeded_notfication_sent",
```
alexey-pelykh, line 22: Typo: `notfication` should be `notification`. Since this is a stored + indexed field, this typo will become the database column name and propagate everywhere (views, filters, translations). Better to fix it now before any deployment.

```suggestion
    is_margin_threshold_exceeded_notification_sent = fields.Boolean(
```

project_minimum_margin_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,157 @@
...
+        ]
+
+    def _cron_margin_threshold_exceeded(self) -> None:
+        """
+        This method will:
+            - Update the field 'is_margin_threshold_exceeded'
+            - Notify users
+        """
+        projects = self.env["project.project"].search([])  # pylint: disable=no-search-all
+        projects._update_is_margin_threshold_exceeded()
```
alexey-pelykh, line 83: This line sets `is_margin_threshold_exceeded_notfication_sent = True` on ALL searched projects, not just the ones that were actually notified in the loop above. Projects that were filtered out (did not match `_get_margin_threshold_to_notify_domain()`) will also be incorrectly marked as "sent".

Consider moving this inside the loop to only mark projects that were actually notified:
```python
for project in projects.filtered_domain(...):
    # ... notification logic ...
    project.is_margin_threshold_exceeded_notfication_sent = True
```

project_minimum_margin_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,157 @@
...
+                project._create_margin_threshold_activity_for_manager(project.user_id)
+        projects.is_margin_threshold_exceeded_notfication_sent = True
+
+    def _send_margin_threshold_notifications(self, users_to_notify):
+        self.message_notify(
+            partner_ids=users_to_notify.filtered(
+                lambda u: u.receive_project_margin_threshold_notification
+            ).partner_id.ids,
+            body=f"Cost threshold exceeded for project {self.name}!",
+            subject="Project Cost Threshold Exceeded",
```
alexey-pelykh, line 103: This calls `_get_profitability_values()` but the tests mock `_get_profitability_items()`. In standard Odoo 19.0, `project.project` has `_get_profitability_items()` which returns a dict (not a tuple). The method `_get_profitability_values()` does not exist by default.

The test passes because the mock is applied to a different method than the one actually called at runtime. This will likely cause an `AttributeError` in production.

Please verify which method provides the margin/profitability data you need and align both the code and tests.

project_minimum_margin_threshold_alert/models/res_config_settings.py
```diff
@@ -0,0 +1,23 @@
...
+        help="Check this to define a default margin",
+    )
+    project_margin_threshold_send_email = fields.Boolean(
+        config_parameter="project_margin_threshold_alert.project_margin_threshold_send_email",
+        help="Check this to enable by default the sending of emails"
+        "to internal users when project margin is exceeded.",
+    )
+    project_margin_threshold_create_activity = fields.Boolean(
+        config_parameter="project_margin_threshold_alert.project_margin_threshold_create_activity",
+        help="Check this to enable by default the creation of activity"
```
alexey-pelykh, line 21: Incomplete help text: `"...when project margin threshold (minimum) is ."` -- the sentence ends with "is ." which appears truncated. Also missing space: `"activityfor"` should be `"activity for"`.

project_minimum_margin_threshold_alert/data/mail_template.xml
```diff
@@ -0,0 +1,36 @@
...
+        <field name="model_id" ref="project.model_project_project" />
```
alexey-pelykh, line 7: Typo: `"threashold"` should be `"threshold"` in both the template name and the subject line (line 8).

project_minimum_margin_threshold_alert/data/cron.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2026 ACSONE SA/NV
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
```
alexey-pelykh, line 4: Consider wrapping this in `<odoo noupdate="1">` so that module updates do not overwrite any user customizations to the cron schedule (interval, active status, etc.).

project_minimum_margin_threshold_alert/tests/test_project_minimum_margin_alert.py
```diff
@@ -0,0 +1,223 @@
...
+        # Set threshold at 80%
+        # Get 0.0 values of invoiced/billed
+        # Threshold is not exceeded
+        with mock.patch.object(
+            ProjectProject, "_get_profitability_items"
+        ) as margin_mock:
+            margin_mock.return_value = self.project_margin_items_15
+            self.project.margin_threshold = 0.2
+            self.project._update_is_margin_threshold_exceeded()
+            self.assertTrue(self.project.is_margin_threshold_exceeded)
```
alexey-pelykh, line 133: The method is named `test_margin_not_exceeded_15` but the assertion is `self.assertTrue(self.project.is_margin_threshold_exceeded)` -- the name says "not exceeded" but the test verifies it IS exceeded. Consider renaming to `test_margin_exceeded_15` for clarity.

project_minimum_margin_threshold_alert/models/project_project.py
```diff
@@ -0,0 +1,157 @@
+# Copyright 2026 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+from odoo.addons.base.models.res_users import ResUsers
```
alexey-pelykh, line 5: This direct import of `ResUsers` from `odoo.addons.base.models.res_users` is unusual in Odoo modules and couples to an internal module path. It is only used as a type annotation for `_get_internal_users_for_margin_threshold`. Consider removing this import and using standard Odoo recordset patterns instead (the method can just return `self.env["res.users"]` typed recordsets without the explicit import).

alexey-pelykh (commented):
## Review: `project_minimum_margin_threshold_alert` (Draft)

This PR is marked as **draft/WIP**, so this is an early-stage review to help guide development. The module concept -- alerting project stakeholders when margin thresholds are breached -- is valuable and well-structured. Below are findings to address before this moves out of draft.

---

### PR Title vs Module Name Mismatch

The PR title references `project_profitability_threshold_alert` but the module is named `project_minimum_margin_threshold_alert`. The title should be updated to match.

---

### Critical / High Priority

**1. Typo in field name: `notfication` instead of `notification`**

The field `is_margin_threshold_exceeded_notfication_sent` has a typo that propagates through the codebase (model, views, filters, i18n). Since this is a stored+indexed field, the typo will be baked into the database column name. This should be fixed now before any deployment.

**2. `_cron_margin_threshold_exceeded` searches ALL projects then marks ALL as notified**

At line 46 of `project_project.py`, the cron does `projects = self.env["project.project"].search([])` then at the end sets `projects.is_margin_threshold_exceeded_notfication_sent = True` on ALL projects -- even those that were NOT notified (filtered out by the domain). This incorrectly marks projects that don't meet the notification criteria as "sent", preventing future notifications if their status changes.

**3. `_get_profitability_values()` does not exist in standard Odoo `project.project`**

The `_update_is_margin_threshold_exceeded` method calls `project._get_profitability_values()` which returns `(margin_values, _show)` and then reads `margin_values.get("expected_percentage", "0")`. However, the standard Odoo project module provides `_get_profitability_items()` (which the tests mock), not `_get_profitability_values()`. This will raise an `AttributeError` at runtime. The tests pass because they mock `_get_profitability_items` but the actual code calls a different method. The test mocking and the runtime code are out of sync.

**4. Missing `ir.model.access` (security rules)**

The `__manifest__.py` does not list any `security/ir.model.access.csv` file. While the module inherits existing models (`project.project`, `res.users`, `res.config.settings`), OCA guidelines typically require this file to be present. Verify whether any new model access rules or record rules are needed.

---

### Moderate Priority

**5. Typo: "Threashold" in mail template**

`mail_template.xml` has "threashold" (should be "threshold") in both the template name and subject line. This was noted in a previous review but appears to still be present.

**6. Incomplete help texts in `res_config_settings.py`**

- `project_margin_threshold_create_activity` help: `"...when project margin threshold (minimum) is ."` -- sentence is incomplete ("is" what?).
- `project_margin_threshold_send_email` help: `"...sending of emailsto internal users..."` -- missing space between "emails" and "to".

**7. `_default_margin_threshold` uses `@property` with `float()` on potentially falsy value**

If the config parameter is not set, `get_param()` returns `False`. Calling `float(False)` returns `0.0` which works, but this is fragile. Consider using `get_param(..., default=0.0)` for clarity.

**8. Missing `__all__` or explicit exports consideration**

The import `from odoo.addons.base.models.res_users import ResUsers` in `project_project.py` is used only as a type hint in `_get_internal_users_for_margin_threshold`. This direct import of a specific model class is unusual in Odoo modules and couples to internal structure. Consider using `self.env["res.users"]` return type or standard Odoo patterns instead.

**9. Cron defined without `noupdate="1"`**

`data/cron.xml` does not use `noupdate="1"`. This means module updates will reset any user customization of the cron schedule. Consider wrapping it in `<odoo noupdate="1">`.

**10. `.pot` file has extra entries not in `.po`**

The `.pot` file contains `"margin Notifications Management"` (lowercase 'm') and `"margin notifications"` (lowercase) which differ from the `.po` file entries (`"Margin Notifications Management"` and `"Margin Notifications"`). This suggests the view XML was updated but translations were not regenerated.

---

### Minor / Suggestions

**11. Test naming is misleading**

`test_margin_not_exceeded_15` actually asserts `assertTrue(self.project.is_margin_threshold_exceeded)`, contradicting the method name which says "not exceeded".

**12. Consider adding `installable: True` to manifest**

While it defaults to `True`, OCA convention is to be explicit.

**13. `fr.po` is missing newline at end of file**

---

alexey-pelykh (commented):
Correction to finding #3 ("_get_profitability_values() does not exist … AttributeError"): that's wrong — it exists on project.project in Odoo 19.0 (addons/project/models/project_project.py:1132), returns (values, show) and calls _get_profitability_items(), so your code is correct and the test's mock takes effect through that call chain. Please disregard that finding — apologies for flagging correct code as Critical. (Findings #1, #2, #4–13 stand.)

Co-Reviewed-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

project_minimum_margin_threshold_alert/tests/test_project_minimum_margin_alert.py
nihelgabsi-acsone: @rousseldenis  can you check this pls

PR #1670, opened 2026-02-17
Add GitHub Actions workflow for Python application

siyabongasilindokuhle22-ux:
This workflow installs Python dependencies, runs linting with flake8, and executes tests using pytest for the specified branch.


siyabongasilindokuhle22-ux (commented):
I need a working app sooner

PR #1671, opened 2026-02-19
[17.0][MIG] project_task_recurring_activity: Migration to 17.0

CristianoMafraJunior:
@Escodoo MIGOCA-50

Migrated `project_task_recurring_activity` to Odoo 17 compatibility, removing legacy recurrence API dependencies (`next_recurrence_date, _get_recurring_fields, planned_hours, etc`.).
Updated `project_task.xml` for V17: recurring activities are no longer inserted in the old recurrence page (V16), but injected into the current task form layout.

**Refactored `models/project_task.py`:**
- removed old logic tied to next_recurrence_date;
- updated call_create_recurring_tasks() to use V17 recurrence flow via _create_next_occurrence();
- preserved propagation of custom_activity_ids to newly created recurring tasks.

**Refactored `models/project_task_recurrence.py`:**
- replaced `_get_recurring_fields override` with `_get_recurring_fields_to_copy`;
- added compatibility helpers for manual and cron creation (create_recurring_tasks and _cron_create_recurring_tasks);
- cron now uses business date (date_deadline) instead of create_date, avoiding freezegun-related instability.

**Refactored `models/recurring_activity.py`:**
- removed references to recurrence_id.next_recurrence_date;
- activity date_deadline is now computed from task date context (with safe fallback).

**Updated `tests/test_project_recurrence.py`:**
- replaced planned_hours with allocated_hours;
- removed assertions relying on legacy fields not present in V17;
- aligned task/activity count expectations with Odoo 17 recurrence behavior;
- removed brittle assertions based on create_date (DB/server real time).


alexey-pelykh (commented):
Thanks for working on this! A few things I noticed: the `next_recurrence_date` field has `string="next_date"` which looks like an internal reference rather than a user-visible label — should be something like "Next Recurrence Date". Also, `project_task_recurrence.py` and `recurring_activity.py` are missing license headers (other files have them). codecov/patch is failing which suggests some paths aren't covered by tests.

project_task_recurring_activity/models/recurring_activity.py
```diff
@@ -0,0 +1,109 @@
...
+    description = fields.Html(
+        sanitize_style=True,
+        compute="_compute_on_activity_type_id",
+        store=True,
+        readonly=False,
+    )
+    days_after_task_creation_date = fields.Integer()
+    next_recurrence_date = fields.Date(
+        string="next_date", compute="_compute_next_recurrence_date", store=True
+    )
```
alexey-pelykh, line 48: `string="next_date"` looks like an internal name. Consider a proper label like "Next Recurrence Date".

project_task_recurring_activity/models/project_task_recurrence.py
```diff
@@ -0,0 +1,30 @@
+from odoo import api, fields, models
+
+
+class ProjectTaskRecurrence(models.Model):
+    _inherit = "project.task.recurrence"
+
+    old_date_recurring_task = fields.Date(default=fields.Date.today())
```
alexey-pelykh: **Bug: Mutable default evaluated at class load time.**

`fields.Date(default=fields.Date.today())` calls `today()` once when the module is loaded, so every new record gets the server start date.

Fix: use a callable:
```python
old_date_recurring_task = fields.Date(default=fields.Date.today)
```
or
```python
old_date_recurring_task = fields.Date(default=lambda self: fields.Date.today())
```

project_task_recurring_activity/security/ir.model.access.csv
```diff
@@ -0,0 +1,2 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
+project_recurring_activity,recurring.activity,model_recurring_activity,,1,1,1,1
```
alexey-pelykh: **Security: Overly permissive access.**

The `group_id:id` column is empty, granting full CRUD access to `recurring.activity` for ALL users (including portal and public). This should be restricted to at minimum `project.group_project_user`:
```csv
project_recurring_activity,recurring.activity,model_recurring_activity,project.group_project_user,1,1,1,1
```

project_task_recurring_activity/models/recurring_activity.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class RecurringActivity(models.Model):
+    _name = "recurring.activity"
+    _description = "Recurring activity"
+
+    project_task_id = fields.Many2one("project.task")
+    activity_type_id = fields.Many2one(
+        "mail.activity.type", string="Activity Type", ondelete="restrict"
+    )
```
alexey-pelykh, line 19: **Missing AGPL-3 license header.** OCA requires `# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).` at the top of every Python file. This file and `project_task_recurrence.py` are both missing it.

project_task_recurring_activity/models/recurring_activity.py
```diff
@@ -0,0 +1,109 @@
...
+                    _(
+                        f"Assigned user {record.user_id.name} has no access"
+                        " to the document and is not able to handle this activity."
+                    )
+                )
+
+    @api.depends("activity_type_id")
+    def _compute_on_activity_type_id(self):
+        for activity in self:
+            if (not activity.description) or activity.description == "<p><br></p>":
```
alexey-pelykh, line 95: **Translation extraction broken: f-string inside `_()`.**

Odoo's `_()` requires a static string literal for the `.pot` extraction tools to work. Using an f-string means this message will never be translatable.

Fix:
```python
raise UserError(
    _(
        "Assigned user %s has no access"
        " to the document and is not able to handle this activity."
    ) % record.user_id.name
)
```

project_task_recurring_activity/models/project_task_recurrence.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @api.model
+    def _get_recurring_fields_to_copy(self):
+        return ["custom_activity_ids"] + super()._get_recurring_fields_to_copy()
+
+    def create_recurring_tasks(self):
+        """Compatibility helper used by the manual button."""
+        for recurrence in self.filtered("task_ids"):
+            last_task = recurrence.task_ids.sorted("id")[-1]
+            recurrence._create_next_occurrence(last_task)
```
alexey-pelykh, line 18: **Performance: Unbounded `search([])`.**

`self.search([]).filtered("task_ids")` loads every recurrence record into memory, then filters in Python. As the database grows, this becomes a memory and performance concern.

Use a domain instead:
```python
for recurrence in self.search([('task_ids', '!=', False)]):
```

alexey-pelykh (changes requested):
Thank you for the migration work on `project_task_recurring_activity` to 17.0. The overall approach of adapting to V17's `_create_next_occurrence` flow looks reasonable. However, there are several issues that need to be addressed before this can be merged.

---

### Critical / Blocking

1. **Missing license headers** (`project_task_recurrence.py`, `recurring_activity.py`): OCA modules require AGPL-3 license headers on every Python source file. Two model files are missing them.

2. **Security: Access rules too permissive** (`ir.model.access.csv`): The `recurring.activity` model grants read/write/create/unlink to ALL users (empty `group_id:id` column). This should be restricted to at minimum `project.group_project_user` or `project.group_project_recurring_tasks`.

3. **Mutable default on field** (`project_task_recurrence.py:7`): `fields.Date(default=fields.Date.today())` evaluates `today()` at class definition time (module load), so every record will get the date the server started, not the date of creation. Use `default=fields.Date.today` (without parentheses) or `default=lambda self: fields.Date.today()`.

4. **`_cron_create_recurring_tasks` scans all recurrences** (`project_task_recurrence.py:17`): `self.search([]).filtered("task_ids")` loads every recurrence record into memory. This is an unbounded query that will degrade as the database grows. Use a domain filter, e.g., `self.search([('task_ids', '!=', False)])`.

5. **f-string inside `_()` translation function** (`recurring_activity.py:84-87`): Using an f-string inside `_()` breaks Odoo's translation extraction. The `_()` function requires a static string at extraction time. Use `%` formatting or `.format()` with `_()` containing only a static template string.

---

### Moderate

6. **`next_recurrence_date` field string** (`recurring_activity.py:44-46`): `string="next_date"` looks like an internal identifier, not a user-facing label. Should be something like `"Next Recurrence Date"`.

7. **Translation files are stale** (`i18n/es.po`, `i18n/it.po`): The `.po` files reference `Project-Id-Version: Odoo Server 14.0` and contain entries for fields like `__last_update` which was removed in V16+. The `.pot` file references `Odoo Server 16.0`. These should be regenerated for 17.0, and stale field references removed.

8. **`_get_next_date` logic is confusing** (`recurring_activity.py:53-58`): The conditional `days + 1 if days == 0 else days` means that setting 0 days actually results in 1 day offset, which is non-obvious. Consider documenting this behavior or using a minimum constraint on the field instead.

9. **`custom_activity_ids` as stored computed Many2many** (`project_task.py:12-14`): This field is `compute="_compute_activity_ids", store=True, copy=True` but it simply mirrors `recurring_activity_ids.ids`. A stored computed field that is also copyable is unusual -- when a task is copied, does the compute fire and overwrite the copy? This interaction should be verified.

10. **`create` override uses `self._forming_activity_data(item, item.custom_activity_ids)`** (`project_task.py:60-65`): But `custom_activity_ids` on the newly created task is a computed field that mirrors `recurring_activity_ids`. If the task was just created as a recurrence copy, this may re-create activities that were already copied. Verify there is no duplication of `recurring_activity_ids` on recurrence creation.

11. **Inconsistent XML indentation** (`data/recurring_activity.xml`): The cron record uses mixed indentation (8 spaces for `<record>` children, then 12 spaces for `<field name="nextcall">`). Follow consistent indentation.

---

### Minor / Suggestions

12. **`recurring_activity.xml` form view**: `readonly="0"` attributes on `user_id`, `summary`, and `description` fields in the form view may be unnecessary if the fields are not readonly by default at the model level (they are `store=True, readonly=False` already).

13. **Test assertion messages**: Some assertion messages are generic (`"Must be equal to 0"`, `"Must be equal to 1"`). More descriptive messages would help debugging test failures.

14. **Unused import in tests** (`test_project_recurrence.py:5`): `from odoo import _, fields` -- the `_` import is used inside `assertRaisesRegex`, which is fine, but note that this makes the test dependent on the exact translated string. Consider matching on a substring instead.

15. **`delta_time` method** (`recurring_activity.py:98-99`): This is a trivial `(new - old).days` wrapper that is only used in tests. Consider whether it needs to be on the model at all, or can be a test helper.

16. **Usage documentation** (`readme/USAGE.md`): Still references "In tab Recurrence" but per the PR description, the V17 migration moved the fields out of the old recurrence page. The documentation should reflect the actual V17 location.

---

project_task_recurring_activity/views/project_task.xml
```diff
@@ -0,0 +1,47 @@
...
+                expr="//div[@id='date_deadline_and_recurring_task']/.."
+                position="inside"
+            >
+                <label
+                    for="recurring_activity_ids"
+                    string="Activities"
+                    groups="project.group_project_recurring_tasks"
+                    invisible="not recurring_task"
+                />
+                <field
```
luisDIXMIT: If you add a label, remove the default field label with nolabel="1" or it will be duplicated. On the other hand, the tree looks too small.

Image

alexey-pelykh (commented):
Thanks for the update — significant improvements on the critical items from my previous review:

### Fixed ✅

- **Security** (#2): ACL now properly restricted to `project.group_project_recurring_tasks`
- **Mutable default** (#3): `fields.Date.today` (callable, not evaluated at load)
- **Unbounded query** (#4): Domain filter `('task_ids', '!=', False)` instead of `search([]).filtered()`
- **f-string in `_()`** (#5): Switched to `%` formatting — translation extraction will work
- **License header** (#1): Added to `recurring_activity.py`
- **`next_recurrence_date` label** (#6): No longer shows `"next_date"`
- **`_get_next_date` documentation** (#8): Field help text explains the 0→1 day behavior
- **Usage docs** (#16): USAGE.md now reflects V17 layout

### Still outstanding

- **License header on `project_task_recurrence.py`** (#1): This file is still missing the AGPL-3 header (see inline comment). This is the last blocking item.
- **Translation files** (#7): `.pot` still references `Odoo Server 16.0` and `.po` files contain stale `__last_update` entries. Should be regenerated for 17.0.
- **XML indentation** (#11): `data/recurring_activity.xml` has inconsistent indentation around the `<field name="nextcall">` element.

Once the license header is added, the remaining items are minor and shouldn't block merge.

project_task_recurring_activity/models/project_task_recurrence.py
```diff
@@ -0,0 +1,30 @@
+from odoo import api, fields, models
```
alexey-pelykh, line 2: Still missing the AGPL-3 license header. This file needs:
```python
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
at the top, before the import. (`recurring_activity.py` now has it — this is the last file missing it.)

project_task_recurring_activity/views/project_task.xml
CristianoMafraJunior: done

project_task_recurring_activity/models/project_task_recurrence.py
CristianoMafraJunior: done

luisDIXMIT (commented):
I think it is too small; it should be placed on a full page, like in 16. This way, I can't see anything in the columns.

Image

CristianoMafraJunior:
> I think it is too small; it should be placed on a full page, like in 16. This way, I can't see anything in the columns.

Done, create page

luisDIXMIT (approved):

CristianoMafraJunior:
@alexey-pelykh review?

CristianoMafraJunior:
@rvalyi review?

alexey-pelykh (approved):
Thanks for pushing this through, @CristianoMafraJunior — all the blocking items from my change request are resolved:

- License header is now present on `project_task_recurrence.py` (the last file that was missing it).
- ACL restricted to `project.group_project_recurring_tasks` (no longer world-writable).
- `old_date_recurring_task` uses the callable `fields.Date.today` (no load-time evaluation).
- The cron scans with a domain — `search([("task_ids", "!=", False)])` instead of `search([]).filtered(...)`.
- `_()` now wraps a static string with `%` formatting applied outside it, so extraction works.
- `next_recurrence_date` dropped the `string="next_date"` label, and `_get_next_date` spells out the 0→1 day rule.

Only cosmetic leftover I see is the `.pot` still reading `Odoo Server 16.0` plus some `__last_update` entries in the `.po` files — non-blocking, as noted before. With luisDIXMIT's runboat test and green CI, LGTM.

PR #1676, opened 2026-02-22
[18.0][ADD] project_task_manual_progress

cvinh:
This module allows to manually enter a 'based on intuition' progress in tasks.
The global project manual progress is calculated from tasks manual progress.
This progress value is not related to time spent on tasks as it is in hr_timesheet.


cvinh:
tests fail because of sale_project_task_selection

luisDIXMIT (commented):
Can you add some tests? :)

project_task_manual_progress/models/project_project.py
```diff
@@ -0,0 +1,24 @@
...
+
+class ProjectProject(models.Model):
+    _inherit = "project.project"
+
+    manual_progress = fields.Float(
+        compute="_compute_manual_progress",
+        help="Shows the progress of the project based on the task manual progress",
+    )
+
+    @api.depends()
```
alexey-pelykh: `@api.depends()` with no args means the field is always recomputed but the ORM cache won't know when to invalidate. Use `@api.depends('task_ids.manual_progress')` so the value stays in sync when tasks are updated.

project_task_manual_progress/models/project_task.py
```diff
@@ -0,0 +1,22 @@
...
+    manual_progress = fields.Float(
+        "Manual Progress (%)",
+        help="Manually enter the estimated progress of the task",
+    )
+
+    @api.constrains("manual_progress")
+    def check_progress(self):
+        for task in self:
+            if task.manual_progress > 100:
+                raise UserError(
```
alexey-pelykh: `@api.constrains` should raise `ValidationError`, not `UserError`. Also this only rejects > 100 but allows negative values -- worth adding `< 0` to the check.

alexey-pelykh (changes requested):
Thanks for the contribution! A couple of things on top of the tests request from @luisDIXMIT.

alexey-pelykh (approved):
Looks good, all three points addressed. Tests would be a nice addition to cover the constraint and the computed progress — not blocking but would help with long-term maintenance.

project_task_manual_progress/__manifest__.py
```diff
@@ -0,0 +1,17 @@
...
+    "category": "Generic Modules/Projects & Services",
+    "website": "https://github.com/OCA/project",
+    "depends": [
+        "project",
+    ],
+    "data": [
+        "views/project_project_views.xml",
+        "views/project_task_views.xml",
+    ],
+}
```
alexey-pelykh, line 20: Missing `"installable": True` -- OCA convention requires this to be explicitly set.

Also consider adding `"development_status": "Beta"` to match the README badge.

project_task_manual_progress/models/project_project.py
```diff
@@ -0,0 +1,24 @@
...
+
+from odoo import api, fields, models
+
+
+class ProjectProject(models.Model):
+    _inherit = "project.project"
+
+    manual_progress = fields.Float(
+        compute="_compute_manual_progress",
+        help="Shows the progress of the project based on the task manual progress",
```
alexey-pelykh, line 14: This computed field is not stored (`store=True` missing). On projects with many tasks it will recompute on every read. Consider `store=True` for performance and to enable searching/grouping.

Also, with `store=True` you may need to add `readonly=True` explicitly (it is implicit for non-stored computed fields).

project_task_manual_progress/models/project_project.py
```diff
@@ -0,0 +1,24 @@
...
+    )
+
+    @api.depends("task_ids.manual_progress")
```
alexey-pelykh, line 17: Depending on `task_ids.manual_progress` will trigger recomputation for ANY task change in the project. In Odoo 18, `task_ids` may include subtasks as well.

Consider whether you want to:
1. Filter out subtasks or tasks in specific stages
2. Only average tasks where `manual_progress > 0` (so unrated tasks don't drag the average to 0)
3. Document the current behavior as intentional

project_task_manual_progress/views/project_project_views.xml
```diff
@@ -0,0 +1,27 @@
+<odoo>
+    <record id="edit_project" model="ir.ui.view">
+        <field name="name">project.project.form.progress</field>
+        <field name="model">project.project</field>
+        <field name="inherit_id" ref="project.edit_project" />
+        <field name="type">form</field>
```
alexey-pelykh: `<field name="type">form</field>` is deprecated in Odoo 18 -- the view type is inferred from the inherited view. This line can be removed.

Same applies to the list view record below (line 20).

project_task_manual_progress/models/project_task.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2026 INVITU (<https://www.invitu.com>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, api, fields, models
```
alexey-pelykh: The `_` import is used, which is good. However, OCA linting typically expects `from odoo import api, fields, models` and `from odoo.tools.translate import _` as separate imports, or at minimum the current style is acceptable if `oca-lint` passes. Just flagging in case CI complains.

project_task_manual_progress/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1 @@
+- Cyril VINH-TUNG \<cyril@invitu.com>\>
```
alexey-pelykh: Typo: extra `>` at the end of the line. Should be:
```
```

alexey-pelykh (commented):
## Review: `project_task_manual_progress` (18.0)

This PR is currently marked as **Draft**, so this is an early-stage review to provide constructive feedback before the PR is ready for formal review.

The module idea is sound -- providing a manual "intuition-based" progress field for tasks is a useful complement to the timesheet-based progress. Here are findings to address before moving out of draft.

---

### Critical: No tests

OCA requires test coverage for new modules. At minimum, tests should cover:
- Creating a task, setting `manual_progress`, verifying the constraint (0-100 range)
- Verifying the project-level `_compute_manual_progress` correctly averages task progress
- Edge case: project with no tasks (division by zero is handled, but should be tested)

### Critical: Missing `installable` key in manifest

While `installable` defaults to `True`, OCA convention requires it to be explicitly set in `__manifest__.py`.

### Important: Project progress computation includes all tasks

`project.task_ids` includes **all** tasks (including subtasks in Odoo 18). Tasks with `manual_progress = 0` (the default) will drag down the project average. Consider:
- Filtering to only tasks that have been explicitly rated (e.g., `manual_progress > 0`), or
- Documenting that this is intentional behavior, or
- Using a weighted approach

Also, `task_ids` may include tasks in folded/cancelled stages -- is that desired?

### Important: Computed field not stored

`manual_progress` on `project.project` is not stored. For projects with many tasks this will recompute on every access. Consider adding `store=True` and appropriate dependencies, which also enables searching/grouping by progress.

### Moderate: Deprecated `<field name="type">` in view XML

In Odoo 18, the `<field name="type">form</field>` / `<field name="type">list</field>` attributes on `ir.ui.view` records are deprecated. The view type is inferred from the `inherit_id`. Remove these lines from both `project_project_views.xml` and `project_task_views.xml`.

### Minor: Typo in CONTRIBUTORS.md

The contributor line has an extra `>` character:
```
```
Should be:
```
```
The same typo appears in `README.rst`.

### Minor: No `.pot` file

A translatable string exists in `project_task.py` (the validation error message). Run `oca-gen-addon-readme` and export translations to generate the `.pot` file.

### Minor: Missing `development_status` in manifest

The README badges show "Beta" but the manifest does not include `"development_status": "Beta"`. Adding it ensures consistency.

### Suggestion: Consider `group_operator` for list view aggregation

The task list view uses `avg="Average of Progress"` on the `manual_progress` field. In Odoo 18, you may want to set `group_operator="avg"` on the field definition itself so that grouping in list view correctly shows averages.

---

**Summary:** The concept is good but the module needs tests, a few model design decisions to be clarified (stored vs. computed, task filtering), and some OCA packaging fixes before it can be considered for merge.

project_task_manual_progress/models/project_project.py
cvinh: this one is still in progress...

cvinh:
tests added

PR #1677, opened 2026-02-24
[16.0][FIX] project_timeline: Prevent at uninstall the error "No default view of type 'timeline'"

arnaudlayec:
**Before**
When uninstalling the module, specific **timeline** views are well removed.
But not the attribute `view_mode` of `ir_act_window` model.
This causes the `UserError` from base `ir_ui_view` model to raise when the user browse to the **Project** menu, blocking all browsing into Project apps. The error is : _"No default view of type 'timeline' could be found !"_

**Fix proposal**
This module adds an `uninstall_hook` which trims the string **timeline** from the "view_mode" attributes of `ir.actions.act_window` model.

If approved, I will forward-port to 18.0.


luisDIXMIT (approved):
Code review, LGTM!

alexey-pelykh (approved):
Good fix. The uninstall hook properly cleans up the `timeline` view mode from affected actions, preventing the "No default view of type 'timeline'" error after uninstalling. Clean implementation. LGTM.

project_timeline/hooks.py
```diff
@@ -0,0 +1,20 @@
...
+
+def _clean_view_mode_timeline(env):
+    models = ["project.project", "project.task", "report.project.task.user"]
+    domain = [
+        ("view_mode", "ilike", "timeline"),
+        ("res_model", "in", models),
+    ]
+    for action in env["ir.actions.act_window"].search(domain):
+        view_mode = action.view_mode.split(",")
+        view_mode.remove("timeline")
```
alexey-pelykh: Minor robustness suggestion: `list.remove()` raises `ValueError` if the element is not found. Since the domain uses `ilike` (case-insensitive), a safer approach would be a list comprehension:
```python
view_mode = [m for m in action.view_mode.split(",") if m.strip() != "timeline"]
```
This also handles any accidental whitespace in the `view_mode` string.

project_timeline/hooks.py
```diff
@@ -0,0 +1,20 @@
...
+        action.view_mode = ",".join(view_mode)
```
alexey-pelykh: Edge case: if `view_mode` was `"timeline"` only (single view mode), after removal `action.view_mode` would be set to `""` (empty string). This might cause issues since Odoo expects at least one view mode. Consider guarding:
```python
if view_mode:
    action.view_mode = ",".join(view_mode)
```
or setting a sensible fallback like `"tree,form"`.

alexey-pelykh (commented):
Follow-up review with a couple of robustness observations on the uninstall hook.

Overall the approach is sound -- cleaning `view_mode` during uninstall is the right fix for the "No default view of type 'timeline'" error. Two edge cases worth considering:

1. **`ValueError` if `remove()` finds no match:** The domain uses `ilike`, which is case-insensitive. If an action's `view_mode` somehow contains `Timeline` (capitalized) or a substring match like `timeline2`, the `ilike` filter would match it, but `view_mode.remove("timeline")` would raise `ValueError` because the exact string `"timeline"` is not in the split list. A safer pattern would be a list comprehension filter:
   ```python
   view_mode = [m for m in action.view_mode.split(",") if m.strip() != "timeline"]
   ```

2. **Empty `view_mode` after removal:** If an action has `view_mode = "timeline"` as its only view mode, after removal the result would be an empty string. Odoo expects at least one view mode in `ir.actions.act_window`. Consider either skipping such actions or setting a fallback (e.g., `"tree,form"`).

Neither of these is likely to occur in practice with the standard `project_timeline` module, so they are minor robustness improvements rather than blockers. The existing approval stands.

project_timeline/hooks.py
arnaudlayec: True, done. Also added `.lower()`

project_timeline/hooks.py
arnaudlayec: True, but I preferred removing the action cleanly.
This requires removing related `ir.ui.menu`.
I also added tests for uninstall.

alexey-pelykh (approved):
Fixes look good. The list comprehension with `.lower()` handles edge cases cleanly, and removing the action + menu when `view_mode` becomes empty is a better approach than a fallback -- an action that only served the timeline view has no purpose post-uninstall. Tests cover both cases (multi-mode trim and single-mode cleanup) and the demo data independence improvement is a welcome bonus.

pedrobaeza:
I have rescued the idea in #1793, but doing directly over the expected actions.

PR #1678, opened 2026-02-25
[16.0][FIX] project_hr: Use the appropriate user employee

victoralmau:
Use the appropriate user employee. Avoid always having to select (in addition to another company) the user's default company.

Use case example:
- Create a user in company A who has access to companies A + B
- Create an employee linked to the user in company A and link category A
- Create an employee linked to the user in company B and link category A
- Log in with the user created above and select only company B
- You will have access to category A

@Tecnativa TT61177


pedrobaeza (approved):
Thanks for spotting the bug. Let's fix it.

/ocabot merge patch

PR #1681, opened 2026-02-26
[18.0][FIX] project_hr: Use the appropriate user employee

victoralmau:
FWP from 17.0: https://github.com/OCA/project/pull/1680

Use the appropriate user employee. Avoid always having to select (in addition to another company) the user's default company.

Use case example:
- Create a user in company A who has access to companies A + B
- Create an employee linked to the user in company A and link category A
- Create an employee linked to the user in company B and link category A
- Log in with the user created above and select only company B
- You will have access to category A

@Tecnativa TT61177

### DIFF
--- a/sale_project_task_selection/tests/test_sale_project.py
+++ b/sale_project_task_selection/tests/test_sale_project.py
@@ -455,7 +455,8 @@ class TestSaleProject(TransactionCase):
         ):
             self.order.action_confirm()
 
-    def test_product_create_task_manually_set_already_linked_order(self):
+    # Disabled for now as it's giving incompatibility error
+    def __test_product_create_task_manually_set_already_linked_order(self):
         """The manually set Task is already linked to another order"""
         # Set up another order with a task linked to it
         another_order = self.order.copy()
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):
Let's finish the chain.

/ocabot merge patch

sale_project_task_selection/tests/test_sale_project.py
```diff
@@ -455,33 +455,33 @@ def test_product_create_task_manually_set_already_linked_sale_line(self):
         ):
             self.order.action_confirm()
 
-    def test_product_create_task_manually_set_already_linked_order(self):
```
pedrobaeza, line 458: Don't comment full method, but just rename it to `def __test...` and put a comment saying that it has been disabled.

victoralmau: I have renamed the method and added the appropriate comment in the commit.

sale_project_task_selection/tests/test_sale_project.py
```diff
@@ -455,7 +455,7 @@ def test_product_create_task_manually_set_already_linked_sale_line(self):
...
+    def _test_product_create_task_manually_set_already_linked_order(self):
```
pedrobaeza:
```suggestion
    # Disabled for now as it's giving incompatibility error
    def __test_product_create_task_manually_set_already_linked_order(self):
```

pedrobaeza: With 2 `_`

PR #1682, opened 2026-02-27
[18.0][MIG] project_kanban_form_direct_access

Saran440:
This PR fixed logic to support project v18

Migration from v17
- [ ] https://github.com/OCA/project/pull/1642

Note:
- It still display wrong when you install `project_sequence` before `project_kanban_form_direct_access`. Name will double. Fixed it in https://github.com/OCA/project/pull/1683
selection_023


project_kanban_form_direct_access/views/project_project.xml
```diff
@@ -4,15 +4,17 @@
...
+                    class="text-truncate d-block fs-4 fw-bold text-primary"
+                    t-att-title="record.display_name.value"
+                >
+                    <field name="display_name" />
                 </a>
             </xpath>
-            <xpath expr="//span[hasclass('o_text_overflow')]" position="attributes">
-                <attribute name="groups">!project.group_project_manager</attribute>
+            <xpath expr="//span[hasclass('text-truncate')]" position="attributes">
+                <attribute name="invisible">1</attribute>
```
luisDIXMIT, line 17: Why are you making it invisible to everyone when before it was only invisible to the group_project_manager?

project_kanban_form_direct_access/views/project_project.xml
```diff
@@ -4,15 +4,17 @@
         <field name="model">project.project</field>
         <field name="inherit_id" ref="project.view_project_kanban" />
         <field name="arch" type="xml">
-            <xpath expr="//span[hasclass('o_text_overflow')]" position="after">
-                <a type="edit" groups="project.group_project_manager">
```
luisDIXMIT, line 8: You are also removing the group restriction here.

project_kanban_form_direct_access/views/project_project.xml
Saran440: The purpose of this module is to access the project from the Kanban view. To keep things simple, I used the old invisible field and added a new field to link directly to the project. What do you think?

luisDIXMIT: Yes, it looks right. Thanks for explaining it to me.

luisDIXMIT (approved):
Tested on runboat and LGTM!

project_kanban_form_direct_access/views/project_project.xml
Saran440: @IriaAlonso Could you clarify the purpose of adding groups? If I change it to invisible, would that work for you?

IriaAlonso: This restriction is to maintain the same restriction on who can access the form view. But I think all users can access it, I have to change that. So it would be good to keep it invisible.

alexey-pelykh (approved):
Clean migration. View-only module, nothing to trip over.

LGTM.

project_kanban_form_direct_access/views/project_project.xml
IriaAlonso: > This restriction is to maintain the same restriction on who can access the form view. But I think all users can access it, I have to change that. So it would be good to keep it invisible.

It is already fixed in the original PR

project_kanban_form_direct_access/views/project_project.xml
```diff
@@ -0,0 +1,21 @@
+<odoo>
+    <record id="project_buttom_kanban_view" model="ir.ui.view">
```
alexey-pelykh, line 2: Nit: typo in the record id -- `project_buttom_kanban_view` should be `project_button_kanban_view` ("buttom" -> "button"). This becomes a persistent `ir.model.data` external identifier, so worth fixing while it's fresh.

alexey-pelykh (commented):
Clean migration to 18.0. The view approach is correct: `type="open"` replaces the v17 `type="edit"`, xpath targets updated from `o_text_overflow` to `text-truncate` matching the Odoo 18.0 base kanban view, and Bootstrap utility classes (`fs-4 fw-bold text-primary`) are consistent with the base view's styling.

The group restriction removal was discussed and confirmed by the original author -- all users can access the project form view, so `invisible="1"` is the right approach.

One minor nit:

- The view record id `project_buttom_kanban_view` contains a typo ("buttom" instead of "button"). This is cosmetic and won't affect functionality, but it will persist as an `ir.model.data` external identifier in the database. Worth fixing before merge if convenient, but non-blocking.

BhaveshHeliconia (approved):
Functional review LGTM!

leemannd (approved):

PR #1688, opened 2026-03-08
[FIX] project_task_pull_request: allow to put a pr_uri, without it is mandatory.

legalsylvain:
Trivial

Rational: For the time being, a PR can be mandatory, if defined via pr_required_states settings. however, in some cases, user want to write this information, but not in all cases. So the field should not be hidden, even if it not required


luisDIXMIT (approved):
LGTM!

CristianoMafraJunior (approved):

leemannd (approved):

PR #1690, opened 2026-03-11
[18.0][MIG] project_wbs: Migration to 18.0

lbarry-apsl:
Module migrated to version 18.0

cc https://github.com/APSL
@miquelalzanillas @javierobcn @mpascuall @BernatObrador @ppyczko @peluko00 please review


hbrunn (changes requested):
this needs much more work, amongst others compatibility with `project_parent`, lack of which is part of why tests fail left and right here

cubells (approved):
:+1: 

LGTM

Code review and test

peluko00 (approved):
LGTM!

mpascuall (approved):
LGTM!

AaronHForgeFlow (commented):
When I click on WBS list to see the children:

Image

An I hit create then I expect to be able to be the name of the new project for quick creation, but it seems to be readonly, can you check?

Image

I understand that make this a dependent of project_parent is the natural evolution of the module.

lbarry-apsl:
Done @AaronHForgeFlow !

Agreed, that makes sense as the next step.
This PR is the 18.0 migration of project_wbs; the inline creation adjustment is part of that migration.
I suggest handling the project_parent dependency/refactor in a separate follow-up PR.

lbarry-apsl:
Hi @hbrunn and @AaronHForgeFlow , can you review please?

AaronHForgeFlow (commented):
LGTM Functional.

Also I am no so sure we should make it depend on project_parent. This module relies on the analytic_account hierarchy not the project hierarchy. Moving all the features from the analytic account to the project is a dramatic change that affects all the features in the module and all the module that relies on this one, and requires complex migration scripts, for so little benefit.

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,238 @@
...
+                    break
+                current = current.parent_id
+
+    @api.model
+    def _default_parent(self):
+        return self.env.context.get("parent_id", None)
+
+    @api.model
+    def _default_partner(self):
+        return self.env.context.get("partner_id", None)
```
hbrunn, line 125: please remove those and change context keys to `default_*` where appropriate

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,238 @@
...
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class AccountAnalyticAccount(models.Model):
+    _inherit = "account.analytic.account"
+    _order = "complete_wbs_code"
+
+    def get_child_accounts(self):
```
hbrunn, line 16: are you sure this is faster than a modern `child_of` search?

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,238 @@
...
+        return result
+
+    def write(self, vals):
+        res = super().write(vals)
+        if vals.get("parent_id"):
+            for account in self.browse(self.get_child_accounts().keys()):
+                account._complete_wbs_code_calc()
+                account._complete_wbs_name_calc()
+        if "active" in vals:
+            for account in self:
```
hbrunn, line 50:
```suggestion
            for account in self.with_context(active_test=False):
```

otherwise you won't find nonactive projects

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,238 @@
...
+        ],
+        "Class",
+        default="project",
+        help="The classification allows you to create a proper project "
+        "Work Breakdown Structure",
+    )
+    parent_id = fields.Many2one(
+        default=_default_parent, string="Parent Analytic Account"
+    )
+    partner_id = fields.Many2one(default=_default_partner)
```
hbrunn, line 181: remove as of above

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def copy(self, default=None):
+        if self.mapped("project_ids"):
+            raise ValidationError(
+                _("Duplicate the project instead of the " "Analytic Account")
+            )
+        default = {}
+        default["code"] = self.env["ir.sequence"].next_by_code(
+            "account.analytic.account.code"
+        )
```
hbrunn, line 191: why do we need this? the default should be provided by account_analytic_sequence already.

in any case, you overwrite existing defaults here which is a no-no

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,238 @@
...
+    wbs_indent = fields.Char(compute=_wbs_indent_calc, string="Level")
+
+    complete_wbs_code = fields.Char(
+        compute=_complete_wbs_code_calc,
+        string="Full WBS Code",
+        help="The full WBS code describes the full path of this component "
+        "within the project WBS hierarchy",
+        store=True,
+    )
+    code = fields.Char(copy=False)
```
hbrunn, line 140: remove, account_analytic_sequence already does this

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,238 @@
...
+    )
+    parent_id = fields.Many2one(
+        default=_default_parent, string="Parent Analytic Account"
+    )
+    partner_id = fields.Many2one(default=_default_partner)
+
+    def copy(self, default=None):
+        if self.mapped("project_ids"):
+            raise ValidationError(
+                _("Duplicate the project instead of the " "Analytic Account")
```
hbrunn, line 186: join the string. but I don't think I agree with making this a hard failure anyways

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,238 @@
...
+                else:
+                    data.insert(0, "")  # pragma: no cover
+
+                acc = acc.parent_id
+            data = " / ".join(data)
+            res.append((account.id, data))
+        return res
+
+    @api.depends("name")
+    def name_get(self):
```
hbrunn, line 212: name_get isn't a thing any more, see https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

`code_get` should be integrated here as it's not used anywhere else

project_wbs/models/account_analytic_account.py
```diff
@@ -0,0 +1,238 @@
...
+            res.append((account.id, data))
+        return res
+
+    _sql_constraints = [
+        (
+            "analytic_unique_wbs_code",
+            "UNIQUE (complete_wbs_code)",
+            "The full wbs code must be unique!",
+        ),
+    ]
```
hbrunn, line 238: I stopped reviewing here, before asking for another review I invite you to adhere to all migration notes for all versions mentioned in https://github.com/OCA/maintainer-tools/wiki#migration, including reordering fields and functions, assigning compute functions by name and not by reference, and having a critical look on all field definitions if you don't redefine things that are already done elsewhere

hbrunn (changes requested):
> Also I am no so sure we should make it depend on project_parent. This module relies on the analytic_account hierarchy not the project hierarchy. Moving all the features from the analytic account to the project is a dramatic change that affects all the features in the module and all the module that relies on this one, and requires complex migration scripts, for so little benefit.

nobody has asked for this. But not breaking it is mandatory.

Out of scope for the migration, but I'd like to see this module gate its functionality with a type field/feature flag on accounts and projects, so that you can have analytic accounts and projects that don't have anything to do with WBS

PR #1701, opened 2026-04-01
Draft: [16.0][FIX] Project_task_code does not work with portal kanban sharing

KKamaa:
This is a fix for portal_task_code  where you can invite a portaluser to be a 'sharing user' so that they can access an interactive project Kanban on portal however when a portal user creates a task via the project sharing view an AccessError is raised because the code field introduced by OCA/project_task_code is not included in SELF_WRITABLE_FIELDS.

To reproduce: 

- Enable project sharing on a project and invite a portal user to 'Share Editable'.
- then login as a portal user, open the shared project and create a new task.
- Task creation fails with: odoo.exceptions.AccessError: You cannot write on code fields in task.

### DIFF
--- a/project_task_code/models/project_task.py
+++ b/project_task_code/models/project_task.py
@@ -3,6 +3,10 @@
 
 from odoo import _, api, fields, models
 
+PROJECT_TASK_PORTAL_FIELDS = {
+    "code",
+}
+
 
 class ProjectTask(models.Model):
     _inherit = "project.task"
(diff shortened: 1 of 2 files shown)


luisDIXMIT (approved):
Code review and tested on runboat, LGTM! Maybe the two commits can be squashed into one.

ntsirintanis (approved):
This PR is exactly the way to go. The code field is introduced in project_task_code, so the corresponding guards should also be properly defined there. Without this fix, installing project_task_code alone causes issues; For example, portal users cannot create tasks from the sharing kanban view when project_task_code is installed.

Installing project_task_code_portal currently hides the problem, since it adds  the code field in protected fields, though that behavior belongs in the base module rather than the portal extension.

So this fine by me, and here is my approval.

It would also make sense to clean up the duplication in project_task_code_portal.

leemannd:
Open question do we have the same issue in newer odoo versions?

leemannd (approved):

PR #1708, opened 2026-04-23
[11.0][FIX] project_timeline: fix error when date_start is not set

oihane:


pedrobaeza (approved):
Let's fix this simple error.

PR #1714, opened 2026-04-27
[ADD] website_project_task_code

etobella:
This module makes a link between project_task_code and website_project to show the task code in the submission of tasks. By default, that is changed, but can be unforced by website:

image


etobella:
The error comes from an incompatibility of the tests with website_project. I will try to find a solution

luisDIXMIT (approved):

PR #1716, opened 2026-04-27
[18.0][ADD] project_task_crm_lead

cvinh:
This module links crm leads from project tasks.
This is useful when we want to contact partners as part of a task without including them in the task.
This is not the same behaviour as "crm_lead_to_task" or "crm_project_task" modules.


cvinh:
tests fixed

PR #1717, opened 2026-04-28
[19.0][MIG] project_task_stage_auto_state

Shide:
Straight forward migration

MT-14700 @moduon @chienandalu @EmilioPascual @yajo please review if you want 😄


EmilioPascual (approved):
LGTM. Functional and code review

luisDIXMIT (approved):
Code review and tested on runboat, LGTM

project_task_stage_auto_state/views/project_task_type_view.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="project_task_type_form_view_inherit" model="ir.ui.view">
+        <field name="name">project.task.type.form.inherit.auto.state</field>
+        <field name="model">project.task.type</field>
+        <field name="inherit_id" ref="project.task_type_edit" />
+        <field name="arch" type="xml">
+            <xpath expr="//sheet/group" position="inside">
+                <group>
+                    <group string="Auto-Done">
```
chienandalu, line 10: The resulting layout is a little bit weird:

Image

chienandalu (commented):
Non blocking:

Did you see the now record [rotting feature](https://www.odoo.com/documentation/19.0/applications/essentials/stages.html#rotting-lead-detection) in v19?

Do you think it'd make sense that auto_done/auto_cancel would revolve around it?

- technical ref: https://github.com/odoo/odoo/blob/19.0/addons/mail/models/mail_tracking_duration_mixin.py

project_task_stage_auto_state/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+    "maintainers": ["Shide", "rafaelbn"],
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "project",
+    ],
+    "external_dependencies": {
+        "python": ["freezegun"],
+    },
```
chienandalu, line 21: It'd included in the core. You don't need to declare it in the manifest

Shide:
> Non blocking:
> 
> Did you see the now record [rotting feature](https://www.odoo.com/documentation/19.0/applications/essentials/stages.html#rotting-lead-detection) in v19?
> 
> Do you think it'd make sense that auto_done/auto_cancel would revolve around it?
> 
> * technical ref: https://github.com/odoo/odoo/blob/19.0/addons/mail/models/mail_tracking_duration_mixin.py

I will take a deep look at it! But I think this module should be as it is. Maybe another module using the rot feature could be interesting.

project_task_stage_auto_state/views/project_task_type_view.xml
Shide: It's a separate group with a lot of possibilities... :/
You don't need to visit this stages a lot, so I'll keep it as it is

chienandalu: No, that's ok, what I meant is that they are weirdly aligned with the other group

rafaelbn (approved):
Thank you!

/ocabot merge patch

BhaveshHeliconia (approved):
Code and functional review LGTM!

PR #1725, opened 2026-05-11
[19.0][MIG] project_task_default_user: Migration to 19.0

NICO-SOLUTIONS:
- Migrate relational field commands to Command API
- Adapt test data for Odoo 19 API changes (groups, M2M handling)
- Fix failing tests after migration from 18.0
- Keep existing default user assignment logic unchanged


NICO-SOLUTIONS:
@CRogos 

Would you mind taking a look?

project_task_default_user/models/project_task_type.py
```diff
@@ -0,0 +1,37 @@
+# Copyright 2025 NICO SOLUTIONS - ENGINEERING & IT
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+from odoo import fields, models
+
+
+class ProjectTaskType(models.Model):
+    _inherit = "project.task.type"
+
+    default_user_ids = fields.Many2many(
```
CRogos, line 10: In case of replace, it is also more than a default. Maybe rename it to assign_user_ids?

CRogos (approved):
Migration LGTM code + tested.

Not sure if **default** could get into confusion with the build-in user defined defaults.

Image

I am also not sure if there is a more generic solution with base.automation. I agree that the UI here is easier to use, but maybe use technically base.automation in the background? (just loud thinking)

NICO-SOLUTIONS:
@CRogos 
Thanks for the input — I see where you're coming from regarding base.automation.

In this case I don't think it's the best fit. The assignment logic is quite tightly coupled to project/stage behavior and the replace/merge strategy. With automations we would likely still end up writing custom Python logic, which in my view would make things harder to maintain, debug and reason about. Keeping this in the model layer feels more explicit and deterministic for this use case.

Regarding the default_* naming: I checked the related core usage, and the pattern is indeed used for context defaults (e.g. default_user_ids in context). That said, I don't see a realistic risk of breaking behavior here.

The core snippet only affects the initial default evaluation of user_ids during task creation via context-based defaults.

This implementation doesn't interfere with that mechanism, since we neither inject default_user_ids into the context nor override default_get. Also, the fields in this module live on project.project and project.task.type, while the referenced core logic is scoped to project.task.

Overall, while the concern is valid from a theoretical standpoint, I don't see a practical path where this would lead to conflicts with the core behavior.

WesleyOliveira98 (approved):
LGTM

luisDIXMIT (commented):
Commit history is not preserved correctly, this is in 18.0 branch, from where you are migrating 

Image

NICO-SOLUTIONS:
> Commit history is not preserved correctly, this is in 18.0 branch, from where you are migrating
> 
> Image

@luisDIXMIT 
The translations were added to v18 later via Weblate after the PR was created.

I have now included the missing translation commits and pushed the updated changes.

luisDIXMIT:
Now there are 2 commits, but if I try to migrate from 18.0, 5 commits appear. Something is wrong.
image

NICO-SOLUTIONS:
@luisDIXMIT 
Squashes have been applied
 
image

luisDIXMIT:
> @luisDIXMIT Squashes have been applied
> 
> image

I didn’t know the commit history could be modified and squash commits together.

luisDIXMIT (approved):

NICO-SOLUTIONS:
> > @luisDIXMIT Squashes have been applied
> > image
> 
> I didn’t know the commit history could be modified and squash commits together.

@luisDIXMIT 
https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

luisDIXMIT:
> > > @luisDIXMIT Squashes have been applied
> > > image
> > 
> > 
> > I didn’t know the commit history could be modified and squash commits together.
> 
> @luisDIXMIT https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests#mergesquash-the-commits-generated-by-bots-or-weblate

Ohh nice, thanks for the explanation!

leemannd (approved):

NICO-SOLUTIONS:
@leemannd 
thx

PR #1691, opened 2026-03-12
[19.0][ADD] project_task_customer_ref

JasminSForgeFlow:
Adds an Order Customer Reference field to Task so that customer PO/requisition numbers are visible on tasks, capturable via website webforms, and kept in sync with the linked sale order.

@ForgeFlow

PR #1730, opened 2026-05-14
[19.0][FIX] project_sequence: make creation test independent of sale_project

DavidJForgeFlow:
test_sequence_after_creation used Form(project.project) on a fresh record. Form initialization runs every model onchange to compute defaults, including sale_project's _onchange_sale_line_id, which reads reinvoiced_sale_order_id — a field restricted by sales_team.group_sale_salesman. When sale_project is installed in the same database, the project-only manager test user has no sale rights and the onchange raises AccessError, even though the behaviour being tested has nothing to do with sales. Check #1691 for an example.

Replace the Form-based pre-save check with a non-persisted new() record, which exercises the same invariant ("no sequence_code before creation") without triggering unrelated onchanges. The post-save assertions still use create() and remain unchanged.

### DIFF
--- a/project_sequence/tests/test_project_sequence.py
+++ b/project_sequence/tests/test_project_sequence.py
@@ -41,10 +41,14 @@ class TestProjectSequence(TransactionCase):
     @users("manager")
     def test_sequence_after_creation(self):
         """Sequence is applied only after project creation."""
-        prj_f = Form(self.env["project.project"])
-        self.assertFalse(prj_f.name)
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow (approved):
👍

MiquelRForgeFlow (approved):

AaronHForgeFlow:
Merging to unlock other Migration PRs, apologies for the rush

/ocabot merge nobump

PR #1736, opened 2026-05-28
[19.0][ADD] project_sequence_type: per-type project sequence

dnplkndll:
New glue module that bridges `project_sequence` and `project_type` so each
project type can have its own sequence — answering the recurring "I need a
different sequence per project type" request.

It adds a `sequence_id` (ir.sequence) field on `project.type`. Projects of a
type take their `sequence_code` from that sequence; types with no sequence —
and projects with no type — keep using the global `project.sequence`. It is
`auto_install`, so it activates only when both base modules are present.

## Non-mechanical adaptations worth flagging
- `project.project.create()` pre-fills `sequence_code` from the type's
  sequence *before* `super()`, mirroring `project_sequence`'s own
  "only when the key is absent" guard. This keeps `project_sequence`'s
  default-name logic intact and never overrides an explicit `sequence_code`.


Yoana-RB (commented):
Hello, thank you very much for the contribution. Overall, everything LGTM (Functional review).

I have one question regarding manual project creation. Currently, it is not possible to select the project type before creating the project, so it is created with the default sequence. If a type is assigned afterward, the sequence is not updated accordingly.

Would it be possible to add a Type field to the creation view so the project can be created with the correct configuration from the start?

Thanks, and best regards!

Image

PR #1738, opened 2026-05-28
[19.0][FIX] project_task_description_template: HTML fields wrap text in <p>

dnplkndll:
`TestDescriptionTemplate` fails on 19.0: the two assertions expect the template
text wrapped in `<span>`, but Odoo 19.0's `Html` field now wraps bare text in
`<p>`. So `_onchange_description_template_id` produces `<p>- Sample Description</p>`.

The module behaviour is correct — only the test's hard-coded expectations were
stale. This updates them to `<p>`, turning the module's tests green again on the
19.0 branch.

### DIFF
--- a/project_task_description_template/tests/test_project_task.py
+++ b/project_task_description_template/tests/test_project_task.py
@@ -21,7 +21,7 @@ class TestDescriptionTemplate(BaseCommon):
         record._onchange_description_template_id()
         self.assertEqual(
             record.description,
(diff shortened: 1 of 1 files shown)


dnplkndll:
@NICO-SOLUTIONS @alan196 @Shide 

just looking to get the CI running. will investigate more if there is a better option.

dnplkndll:
https://github.com/OCA/project/pull/1739

PR #1741, opened 2026-05-29
[18.0][ADD] project_no_portal: new module to block portal access on projects

divad1196:
## Context

A project with `privacy_visibility = 'portal'` ("Invited portal users") lets any portal follower of the project or its tasks read it, and managers can share project/task data via the "Share Project" wizard and the "Share Task" public link.

## What it does

A per-company switch `block_project_portal_access` (`res.company`, on by default, under Settings > Project).

For a blocking company:
- A constraint forbids `privacy_visibility = 'portal'` (default becomes `employees`).
  A `post_init_hook` migrates existing portal projects. This alone removes portal read access.
- A mixin on `project.project`/`project.task` also denies portal users in `_search`/`_check_access`.
  This adds an additional security without deleting existing `ir.model.access`
- The "Share Project" / "Share Task" actions are unbound.
  Calling them, for example from the code, raises an error.
  An `uninstall_hook` restores them.

Scoped per company (multi-company friendly) and fully reversible.


project_no_portal/models/project_portal_block_mixin.py
```diff
@@ -0,0 +1,51 @@
...
+        companies = self.env["res.company"].sudo().search([])
+        allowed_ids = companies.filtered(
+            lambda c: not c.block_project_portal_access
+        ).ids
+        keep = [("company_id", "in", allowed_ids)]
+        # company_id is False -> falls back to the current company.
+        if not self.env.company.block_project_portal_access:
+            keep = ["|", ("company_id", "=", False)] + keep
+        domain = expression.AND([domain, keep])
+        return domain
```
mmequignon:
```suggestion
        keep = [("company_id.block_project_portal_access", "=", False)]
        # company_id is False -> falls back to the current company.
        if not self.env.company.block_project_portal_access:
            keep = ["|", ("company_id", "=", False)] + keep
        domain = expression.AND([domain, keep])
        return domain
```

project_no_portal/models/project_portal_block_mixin.py
```diff
@@ -0,0 +1,51 @@
...
+    _description = "Deny portal users access for companies that block it"
+
+    def _portal_block_restricted_user(self):
+        # Only portal users (outside superuser mode) are subject to the block.
+        return not self.env.su and self.env.user._is_portal()
+
+    def _portal_block_company(self):
+        # A record with no company falls back to the current company.
+        self.ensure_one()
+        return self.company_id or self.env.company
```
mmequignon, line 19: Those could be decorated with `@property`.

vvrossem (approved):
LGTM

project_no_portal/models/project_project.py
```diff
@@ -0,0 +1,77 @@
...
+        # Core defaults to "portal"; avoid clashing with the constraint when the
+        # current company blocks portal access.
+        if self.env.company.block_project_portal_access:
+            return "employees"
+        return "portal"
+
+    @api.constrains("privacy_visibility", "company_id")
+    def _check_no_portal_visibility(self):
+        for project in self:
+            company = project.company_id or self.env.company
```
mmequignon:
```suggestion
            company = project._portal_block_company()
```

project_no_portal/models/project_project.py
```diff
@@ -0,0 +1,77 @@
...
+            if (
+                company.block_project_portal_access
+                and project.privacy_visibility == "portal"
+            ):
+                raise ValidationError(
+                    _(
+                        "Portal visibility is disabled for company %(company)s: "
+                        "portal users must not access its projects or tasks.",
```
mmequignon: I would use `can` instead of `must`.
```
portal users cannot access their projects or task
```
What do you think?

project_no_portal/models/project_project.py
```diff
@@ -0,0 +1,77 @@
...
+                        company=company.display_name,
+                    )
+                )
+
+    def action_open_share_project_wizard(self):
+        self.ensure_one()
+        company = self.company_id or self.env.company
```
mmequignon:
```suggestion
        company = project._portal_block_company()
```

project_no_portal/models/project_project.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _set_share_project_action(self, enabled):
+        """Toggle the "Share Project" action"""
+        action = self.env.ref(
+            "project.project_share_wizard_action", raise_if_not_found=False
+        )
+        if not action:
+            return
+        model = self.env.ref("project.model_project_project")
+        action.binding_model_id = model.id if enabled else False
```
mmequignon, line 77: This is basically doing the same as `project.task._set_share_task_action`.
Could this be factorized and extracted in the mixin ?
```
self.env["ir.model"].search([("model", "=", self._name)])
```
etc…

SilvioC2C (approved):
Pre-approving: I agree w/ @mmequignon comments 

Also: code and tests both might use some more comments to make everything easier to read and maintain

project_no_portal/models/project_project.py
divad1196: no preference  
- `must` = cause/why
- `can` = consequence

I will make the change

project_no_portal/models/project_project.py
divad1196: The mixin blocks access to the portal users. It does not handle the `grant` side which can differ from models.
Adding it to the mixin would be convenient but breaks [Interface Segregation](https://en.wikipedia.org/wiki/SOLID) (SOLID 4th principle)

IMO, another mixin just for this function would be overkill.
So maybe a free utility function ?

project_no_portal/models/project_portal_block_mixin.py
divad1196: I am usually against `@property` (and `@api.models`) in the context of Odoo:
- we never know how someone might want to override the function
- it makes it clear to the user that it's not a field.

`()` are a small inconvenience in comparison

wouitmil (approved):

leemannd:
Hello @divad1196 could you have a look at the pre-commit and tests?

Let me know when your work is ready

divad1196:
@leemannd pre-commit and tests passing.
Sorry for the delay, I wasn't available.

project_no_portal/models/project_portal_block_mixin.py
leemannd: Hello @mmequignon , @divad1196 what's the final take on this one?

divad1196: @leemannd it didn't seem like a requirement, more like an advice
> Those **could** be

@mmequignon is that correct?

If others also think that this should be changed, I will do the change and I would appreciate if someone could give me the rationale behind the decision.

mmequignon: Yeah, I'm fine with both

leemannd (approved):

BhaveshHeliconia (approved):
LGTM!

project_no_portal/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+{
+    "name": "Project No Portal",
+    "summary": "Block portal access on project.project and project.task",
+    "version": "18.0.1.0.0",
+    "category": "Project",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "maintainers": ["camptocamp"],
```
simahawk: we must have a person here, not a company

divad1196: I have put myself, 
hope this is fine

leemannd: @simahawk are you aware of this issue? https://github.com/OCA/project/pull/1741#issuecomment-4750153831

simahawk: yes, nothing we can do if not retry once more later

simahawk:
@divad1196 pls squash

PR #1742, opened 2026-06-05
[18.0][FIX] project_group: compatibility with sale_project

Sloweneuh:
The official module **sale_project** injects the *privacy_visibility* field in the *button_box* of the form, causing the *group_ids* field injected by **project_group** to be injected there too.

This PR changes the Xpath of the view to make sure the field is inserted in the right spot.


hildickethan (approved):
Image
need to change this for precommit to pass

PR #1744, opened 2026-06-08
[18.0][FIX] project_hr: use compute_sudo on cross-model computes

Abranes:
When a basic project user opens a task assigned to them, and that task belongs to a project where only invited internal users have access, opening the task raises `AccessError`.

The root cause is that the module's computes read from project.project, but the affected user does not have read access to that record. Marking those computes as `compute_sudo=True` restores the pre-installation behaviour.


hildickethan (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #1748, opened 2026-06-15
[19.0][ADD] project_task_stage_state_sync

IsabelAForgeFlow:
Added project_task_stage_state_sync module


project_task_stage_state_sync/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
```
CristianoMafraJunior, line 1: According to the OCA, you don't put a License [header] in `__init__.py` files.

project_task_stage_state_sync/tests/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
```
CristianoMafraJunior, line 1: According to the OCA, you don't put a License [header] in `__init__.py` files.

project_task_stage_state_sync/__init__.py
```diff
@@ -0,0 +1,3 @@
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
```
CristianoMafraJunior, line 1: According to the OCA, you don't put a License [header] in __init__.py files.

CristianoMafraJunior (changes requested):

MarinaAForgeFlow (approved):

project_task_stage_state_sync/models/project_task.py
```diff
@@ -0,0 +1,32 @@
...
+                task.stage_id = stage
+
+    def _find_stage_for_state(self, state):
+        self.ensure_one()
+        if not state:
+            return self.env["project.task.type"]
+        stage = self.project_id.type_ids.filtered(
+            lambda s: s.task_state == state and s.sync_state_to_stage
+        )[:1]
+        return stage or self.env["project.task.type"].search(
```
MarinaAForgeFlow: If there is no stage in the project that is configured with the wanted state, then it returns any stage of any other project that it is sync to the required state. Why would we want this? If we don't find a stage that is linked to that state in the project, shouldn't we simply return self.env["project.task.type"] ?

IsabelAForgeFlow: You are right, I have changed this!

PR #1758, opened 2026-06-22
[18.0][ADD] project_task_create_date: new module

chienandalu:
image

cc @moduon MT-15112

please review @fcvalgar @Gelojr


sergio-teruel (approved):

CristianoMafraJunior:
Wouldn't it be better to just add optional="hide" to the main view for project_task_create_date instead of creating a separate module just for that?

chienandalu:
> Wouldn't it be better to just add optional="hide" to the main view for project_task_create_date instead of creating a separate module just for that?

@CristianoMafraJunior What do yo mean? Yes this module is as simple as it is... but the need is just to have that column...

CristianoMafraJunior:
> > Wouldn't it be better to just add optional="hide" to the main view for project_task_create_date instead of creating a separate module just for that?
> 
> @CristianoMafraJunior What do yo mean? Yes this module is as simple as it is... but the need is just to have that column...

Sorry for the confusion, I thought I could put it in another module since it's just adding a single column, but it's ok

CristianoMafraJunior (approved):

leemannd (approved):

BhaveshHeliconia (approved):
LGTM!

rafaelbn (approved):
👍🏼 thank you!

/ocabot merge patch

PR #1761, opened 2026-07-01
[18.0][ADD] project_helpdesk_convert_restrict: prevent non-portal task to be converted into tickets

divad1196:
## Context
Tasks of a project can technically leak even with a non-portal privacy setting.  
This can happen by converting the task into a ticket

## What it does
Feature must be enabled with a company-related setting.
When the feature is enabled:

The conversion task-to-ticket is only allowed for tasks where the project has privacy set to `Invited portal users ...`


divad1196:
@leemannd @simahawk Hi, I was wondering if there was a place for enterprise-dependent modules in the OCA ?

divad1196:
moving to a private repository due to enterprise dependency

PR #1785, opened 2026-08-05
[19.0][ADD] project_hr_expense_analytic

simahawk:
This module links expense records to a project and, optionally, a task,
and derives the expense's analytic distribution from that project.


imlopes (approved):
LGTM

project_hr_expense_analytic/i18n/fr.po
gurneyalex, line 1: please rename to `fr.po` there is nothing specifically swiss

gurneyalex: Also maybe we can have a pot file?

project_hr_expense_analytic/models/project.py
```diff
@@ -0,0 +1,10 @@
+# Copyright 2026 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ProjectProject(models.Model):
+    _inherit = "project.project"
+
+    hr_expense_ids = fields.One2many("hr.expense", "project_id", string="HR Expenses")
```
gurneyalex, line 10: is this used anywhere?

gurneyalex (approved):
preapproving, the i18n needs fixing IMO

project_hr_expense_analytic/models/project.py
simahawk: I don't think so. We could get rid of it.

project_hr_expense_analytic/i18n/fr.po
simahawk: pot will be generated, po name fixed

BhaveshHeliconia (approved):

leemannd (approved):

PR #1786, opened 2026-08-05
[19.0][IMP] project_task_code: per-project task name display

JasminSForgeFlow:
Adds a Task Name Display setting on the project, choosing how its tasks are identified:

 - Number and Title (default) — the current [code] Title behaviour, unchanged.
 - Number only — the task number alone. The title is hidden in the form and kanban and is no longer required, so tasks can be identified by their number without carrying a redundant title.

 This replaces the usual downstream workaround of overwriting name with code and flattening display_name, which destroys the title at the data level — most visibly for tasks created from an incoming email, where Odoo puts the mail subject in the title.

 Two related fixes come with it:

 - display_name falls back to the task number when a task has no title, instead of computing an empty display name.
 - Duplicating an untitled task keeps it untitled, instead of producing a "False (copy)" title.

 name is redefined as not required at model level (a number-only project has no title to require); the requirement moves to the form view, where it applies only to projects displaying the title.

### DIFF
--- a/project_task_code/README.rst
+++ b/project_task_code/README.rst
@@ -42,6 +42,17 @@ This module adds a sequential code for tasks.
 Configuration
 =============
 
+To choose how the tasks of a project are identified, you must:
+
+1. Go to Project > Configuration > Projects and open a project.
+2. In the Settings tab, set "Task Name Display" to:
+
+   - "Number and Title" (default) to show the task number followed by
+     its title.
+   - "Number only" to show the task number alone. Titles are still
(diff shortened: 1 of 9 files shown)


AaronHForgeFlow (approved):
Code + Functional all changes LGTM

MarinaAForgeFlow (approved):

BhaveshHeliconia (approved):
LGTM!

leemannd (approved):

PR #1788, opened 2026-08-10
[18.0][FIX] project_timesheet_time_control: keep translatable terms free of markup whitespace

brambuijs:
## Problem

Four view terms in this module span several XML lines. Odoo's `translate_xml_node()` groups consecutive translatable siblings into one term and only strips the *outer* whitespace, so the embedded newline and indentation end up **inside the msgid**:

```
".\n                            If you continue, it will be stopped with"
"<br/>\n                    <span>to</span>"
"<span class=\"o_label\">\n                            <i class=\"fa fa-lg fa-play-circle text-success\"/>\n                            Start work\n                        </span>"
"<span class=\"o_label\">\n                            <i class=\"fa fa-lg fa-stop-circle text-warning\"/>\n                            Stop work\n                        </span>"
```

Any reindentation of the surrounding markup changes the msgid and silently orphans every existing translation.

That is not hypothetical, it already happened. `nl.po`, `fi.po` and `he_IL.po` still carry the previous translation as an obsolete entry with a different indentation width, while the current msgid sits there empty:

```po
#. module: project_timesheet_time_control
#: model_terms:ir.ui.view,arch_db:project_timesheet_time_control.hr_timesheet_switch_form
msgid ""
".\n"
"                            If you continue, it will be stopped with"
msgstr ""

#~ msgid ""
#~ ".\n"
#~ "                        If you continue, it will be stopped with"
#~ msgstr ""
#~ ".\n"
#~ "                        als u doorgaat zal dit gestopt worden met"
```

Note the 28 spaces in the live entry against 24 in the obsolete one. Across the 44 language files, 176 instances of these four terms are untranslated. In practice a Dutch user reads a sentence that switches to English halfway through:

> U heeft een lopende timer genaamd X en gestart op Y. **If you continue, it will be stopped with** Z uur/uren.

## Fix

Keep each translatable run on a single line, and take the `<br/>` out of the run in the kanban template so the term is just the `<span>`. Resulting terms:

```
". If you continue, it will be stopped with"
"<span>to</span>"
"<span class=\"o_label\"><i class=\"fa fa-lg fa-play-circle text-success\"/> Start work</span>"
"<span class=\"o_label\"><i class=\"fa fa-lg fa-stop-circle text-warning\"/> Stop work</span>"
```

No term contains a newline any more, so reindentation can no longer break them.

## Verification

- Ran `odoo.tools.translate.xml_translate` over the arch blocks before and after. The same set of terms is produced, no terms added or removed, and none of them contains a newline.
- Checked formatting with this repository's prettier configuration (`prettier@3.3.3` + `@prettier/plugin-xml@3.4.1`): `All matched files use Prettier code style!`
- Rendering is unchanged, apart from one small improvement: the period after the timer start time no longer has a stray space in front of it, because that space came from the newline in the markup.

## Notes

- No `.po` files are touched, since translations are managed through Weblate. The four msgids do change, so they will need translating again there. 176 of those instances were already untranslated, and for the three languages that still hold the old text as an obsolete entry it is a copy-paste away.
- `<br t-translation="off" />` is used in the kanban template purely to keep the `<br/>` out of the translatable run, since `<br>` is in `TRANSLATED_ELEMENTS`. Happy to use another approach if you prefer, but moving it or giving the `<span>` a block class both change the layout.

### DIFF
--- a/project_timesheet_time_control/i18n/ar.po
+++ b/project_timesheet_time_control/i18n/ar.po
@@ -22,7 +22,6 @@ msgstr ""
 #. module: project_timesheet_time_control
 #. odoo-python
 #: code:addons/project_timesheet_time_control/wizards/hr_timesheet_switch.py:0
-#, python-format
 msgid ""
 "%d running timers found. Cannot know which one to stop. Please stop them "
 "manually."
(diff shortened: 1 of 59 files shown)


pedrobaeza (changes requested):
- Check the new AI policy https://github.com/pypa/pip/blob/main/AI_POLICY.md
- This doesn't include the .pot and .po files changes.
- The module version is changed by the bot, not here.

brambuijs:
Thanks for the review.

**Module version**: reverted, I had missed that `[BOT] post-merge updates` handles it.

**AI policy**: read and understood. No LLM co-author tags on the commit, and I take full
ownership of the change.

**.pot and .po**: added. Flagging that CONTRIBUTING says PRs should never modify `.po`
and leave it to Weblate. I read your comment as the exception for PRs that change source
terms, as in your own 1f0e307 in this module. Say the word and I will keep only the
`.pot`.

* `.pot` regenerated, all 55 `.po` run through `msgmerge --previous`, so changed entries
  keep their translation with `#, fuzzy` and the `#|` previous msgid.
* Three translations recovered that had been obsolete since the indentation last shifted
  (nl, fi, he_IL).
* One hand edit I am least comfortable with: `msgmerge` does not match
  `"<br/>\n <span>to</span>"` to `"<span>to</span>"`, so that entry came out empty in
  every language. I carried the existing translation over without the `<br/>` for the ten
  languages that had one, marked fuzzy. Happy to drop this part.

Checks: all catalogs pass `msgfmt --check-format`; the `model_terms` msgids in the `.pot`
are exactly the 19 terms `xml_translate` extracts from the views after the change, none
containing a newline; comparing per language and per term, no translation is lost.

`<br t-translation="off" />` is only there to keep the `<br/>` out of the translatable
run, since `br` is in `TRANSLATED_ELEMENTS`. Putting the two on one line gets reverted by
prettier and a block class changes the layout, but I am open to another approach.

PR #1790, opened 2026-08-13
[MIG] project_task_project_required: Migration to 18.0

SirPyTech:
Migration from `16.0` (https://github.com/OCA/project/tree/d5909e76883f1228768986cef745a7e300da1204/project_task_project_required).


tonygalmiche (approved):
Tested on a local Odoo 18.0 instance: with "Require Projects on Tasks" enabled (default), creating a task without a project correctly raises a validation error; disabling the setting allows it again. 
Works as expected.

project_task_project_required/views/project_task.xml
```diff
@@ -42,11 +31,12 @@
...
-                <field name="is_project_required" invisible="1" />
-                <field
-                    name="project_id"
-                    attrs="{'required':[('is_project_required','=',True)], 'invisible':[('is_project_required','=',False)]}"
+            <field name="project_id" position="attributes">
+                <attribute name="required">is_project_required</attribute>
+                <attribute
+                    name="invisible"
+                    separator=" and "
+                    add="not is_project_required"
```
HekkiMelody, line 39: **comment**: In the base view we have

```xml
invisible="context.get('default_project_id', False)"
```

so the field is invisible if the context would automatically populate it. Whereas here, if the project is not required, then it's simply not possible to set the field, which I don't think is correct.

I'd remove this `<attribute>` since it affects a pre-existing field (instead of being added to an entirely new field like in v16). What do you think?

HekkiMelody (commented):
A functional comment, otherwise LGTM

project_task_project_required/views/project_task.xml
HekkiMelody: Nevermind, I think I misread it

HekkiMelody (approved):
Code review, LGTM

HekkiMelody:
I've done some functional testing, and with the changes in v18, maybe this module can't be ported as-is anymore.

For example, the module `project_todo` from base Odoo lets the user create tasks without project that act as personal todos. With this module installed and the setting checked to require projects on tasks, `project_todo` is no longer usable as intended.

I think we should review the actual use-case for this module and maybe create something more specific for our needs.

SirPyTech:
Working on a different implementation is more effort than we can spend for this requirement, closing

PR #1791, opened 2026-08-13
[ADD] project_portal_hide_timesheet

SirPyTech:
Migrating the module that has been proposed for `16.0` in https://github.com/OCA/project/pull/1521.

> This module hides details about timesheets and time spent on the tasks in the portal pages.


HekkiMelody (approved):
Code review, LGTM

PicchiSeba (approved):
Code review: LGTM

BhaveshHeliconia (approved):

HekkiMelody:
@OCA/project-service-maintainers this PR has 3 positive reviews, could you please take a look? Thanks

project_portal_hide_timesheet/views/project_task_templates.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <template id="portal_tasks_list" inherit_id="project.portal_tasks_list">
```
leemannd: What's the scope of this? Required to hide the timesheets?

leemannd (changes requested):
This module is doing more than expected.

- if you login as "portal" user. You don't see the tasks anymore in `http://oca-project-18-0-pr1791-5ff7dd934bde.runboat.odoo-community.org/my/projects/12`
- you hide SO/invoice billing informations
- you create a breaking change regarding the OCA (deactivating a view)

**Current proposition**
This is open to discussion. The module hides more than timesheets and may break other OCA modules that are depending on the archived views.

1. Changes that goes beyond the simple scope of hiding timesheets must be described into the readme (bug / feature /...)
2. The possibility for this module to break other ones installed (by archiving the views must be clearly stated)
3. If you want to hide a little bit more than just timesheets, maybe we want to change the module name
4. As per se the module is in my opinion the module maturity should be `alpha`
5. I want another maintainer to agree on this PR

Do you agree with these points?

project_portal_hide_timesheet/views/project_task_templates.xml
SirPyTech: No, this shows the following label in task views: 

Image

I agree it does not belong in this module, removed.

SirPyTech (commented):
Thanks for having a look!

> - if you login as "portal" user. You don't see the tasks anymore in `http://oca-project-18-0-pr1791-5ff7dd934bde.runboat.odoo-community.org/my/projects/12`

This does not depend on this module, that happens in `18.0` too, see http://oca-project-18-0-3991ffca5c21.runboat.odoo-community.org/my/projects/12.
Please try to install only the proposed module and let me know if it causes any issue.

> - you hide SO/invoice billing informations

Could you please clarify which information is hidden and where do you expect to see it? From what I can see the deactivated views only hide timesheet-related information.

> - you create a breaking change regarding the OCA (deactivating a view)

How is deactivating a view a breaking change?

> 1. Changes that goes beyond the simple scope of hiding timesheets must be described into the readme (bug / feature /...)

I have removed the extra changes

> 2. The possibility for this module to break other ones installed (by archiving the views must be clearly stated)

The README is for the end user and they should not care about how the module's scope has been achieved, exposing in the README the implementation details seems redundant.

> 3. If you want to hide a little bit more than just timesheets, maybe we want to change the module name

I don't

> 4. As per se the module is in my opinion the module maturity should be `alpha`

The definitions are:

> - **Alpha**: Unstable, for development or testing purpose
> - **Beta**: Pre-production quality but with potential instability

**Beta** seems more appropriate to me, why should this be Alpha?

> 5. I want another maintainer to agree on this PR

Why? We already have multiple reviews.

leemannd:
@SirPyTech thanks for the time taken. I will redo a test with only this module installed.

> The README is for the end user and they should not care about how the module's scope has been achieved, exposing in the README the implementation details seems redundant.

I disagree. IMO the README is not only for end users.
The way of archiving views with a hook may result in having other modules not working. The readme is the place to acknowledge this.

> Why? We already have multiple reviews.
Because we are on a community and I don't see much other example of archiving a view with an install hook. I want to ensure to be on the same page as other members of this PSC.

SirPyTech:
> @SirPyTech thanks for the time taken. I will redo a test with only this module installed.
> 
> > The README is for the end user and they should not care about how the module's scope has been achieved, exposing in the README the implementation details seems redundant.
> 
> I disagree. IMO the README is not only for end users.
> The way of archiving views with a hook may result in having other modules not working. The readme is the place to acknowledge this.

I'm happy to see a more human interaction :relaxed:
Hi, yes the README is for developers too: it's unusual to disclose implementation details in the README, but I agree it's just as unusual to deactivate views, so I have added it to the README.

> > Why? We already have multiple reviews.
> 
> Because we are on a community and I don't see much other example of archiving a view with an install hook. I want to ensure to be on the same page as other members of this PSC.

:+1: let's see if someone else gets involved then.
Please have a look at the discussion that happened in the original PR https://github.com/OCA/project/pull/1521#issuecomment-5279205482 about the archived views, I motivated this choice there almost a month ago but I've had no reply yet.

PR #1804, opened 2026-09-10
[18.0][ADD] project_prevent_analytic_rename

Shide:
Prevent renaming analytic account when chaning project's name

https://www.loom.com/share/43209d35a5a94b07aace50e35b500a34

MT-9379 @moduon @rafaelbn @fcvalgar @Gelojr @EmilioPascual @chienandalu please review if you want 😄


project_prevent_analytic_rename/models/project_project.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class ProjectProject(models.Model):
+    _inherit = "project.project"
+
+    def write(self, vals):
+        return super(
+            ProjectProject,
+            self.with_context(prevent_analytic_rename=True),
+        ).write(vals)
```
chienandalu: non blocking: maybe we inject the context just in case that "name" is in `vals`

chienandalu (approved):
Code review 👍

project_prevent_analytic_rename/models/project_project.py
Shide: done

EmilioPascual (approved):
LGTM

rafaelbn (commented):
Hi Edu,

I personally reviewed and tested this in Runboat :+1:

- Standard Odoo renames the analytic account when its only linked project is renamed.
- With this module installed, the analytic account keeps its name.
- The analytic account can still be renamed directly and keeps its new name after renaming the project again.

Everything works as expected.

Please make these changes before merging:

- [ ] Rename `project_prevent_analytic_rename` to `project_analytic_account_name_no_sync`.

The current name suggests that users cannot rename the analytic account, but they can. The module only stops the automatic name change caused by renaming the project.

The usual Odoo, OCA and Moduon naming convention goes from general to specific:

`<functional area>_<object>_<behavior>`

In this case: `project` → `analytic_account` → `name` → `no_sync`.

Examples: `project_task_stage_lock`, `project_task_analytic_propagation` and `account_analytic_sequence`.

- [ ] Update the module name and summary:

```python
"name": "Project Analytic Account Name: No Sync",
"summary": "Keep analytic account names unchanged when projects are renamed",
```

- [ ] Update the PR title to `[18.0][ADD] project_analytic_account_name_no_sync`.

- [ ] Replace `readme/CONTEXT.md` with:

```markdown
By default, Odoo renames an analytic account when its only linked project is
renamed. It does not do this when the same account is shared by several
projects.

This can be inconvenient when project and accounting teams use different naming
conventions.
```

- [ ] Replace `readme/DESCRIPTION.md` with:

```markdown
This module keeps the analytic account name unchanged when its project is
renamed.

It applies to all projects, requires no configuration, and users can still
rename the analytic account directly.
```

- [ ] Replace `readme/USAGE.md` with:

```markdown
No configuration is required after installation.

1. Enable Analytic Accounting if it is not already enabled.
2. Open or create a project linked to an analytic account.
3. Rename the project and save it.
4. The analytic account keeps its existing name.
5. To change it, rename the analytic account directly.
```

- [ ] Make sure the OCA icon is displayed correctly in the Apps view. It was not shown during my Runboat test.

Once this is done, it is OK for me.

Thanks!

